Issue #1, opened 2018-04-19
Generation of source code bundle fails


maierthomas:
A generated source code bundle contains only 0-byte files.
All containing files are empty.

mcjaeger:
This issue was moved to sw360/sw360portal#820

PR #2, opened 2018-04-19
fix(attachments): Allows to set attachments for source bundle generation

maierthomas:
- add complete attachments to source bundle generation (zip file)

closes eclipse/sw360#1

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/ProjectPermissions.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/ProjectPermissions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2014-2017. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2014-2018. Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
  *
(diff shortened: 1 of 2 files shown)


libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/DocumentPermissions.java
```diff
@@ -125,12 +123,7 @@ protected boolean isClearingAdminOfOwnGroup() {
                 .collect(Collectors.toList());
     }
 
-    public boolean isAllowedToDownload(AttachmentContent attachment){
-        return isAllowedToDownload(attachment.getId());
-    }
-
-    public boolean isAllowedToDownload(String attachmentContentId){
-        return nullToEmptySet(getAttachmentContentIds()).contains(attachmentContentId) &&
```
maxhbr: I think that is for the "one is only allowed to download an attachment if it is in the list of attachments of something AND if one is allowed to read the something"

Without that check one is able to download arbitrary attachments in the context of something to which one has read access.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/ProjectPermissions.java
```diff
@@ -68,7 +68,7 @@ public static boolean isUserInBU(Project document, String bu) {
               && !isNullOrEmpty(document.getBusinessUnit()) && document.getBusinessUnit().startsWith(buFromOrganisation);
     }
 
-    public static boolean userIsEquivalentToModeratorinProject(Project input, String user) {
+    public static boolean userIsEquivalentToModerationProject(Project input, String user) {
```
maxhbr: I think this change makes the sitution worse, the better alternative might be `userIsEquivalentToModeratorInProject`

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/DocumentPermissions.java
maxhbr: > > - Why do we need to verify the attachment context and permissions for components and releases. [...]
> 
> to be future proof and to have a generic and uniform implementation.

If we provide some endpoint to download atatchments (e.g. of an release) without context, then one could use that endpoint to again download arbitrary attachments.

Issue #130, opened 2018-04-26
Export Spreadsheet is missing "Project Mainline State"


maierthomas:
<a href="https://github.com/daral"><img src="https://avatars3.githubusercontent.com/u/18406195?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [daral](https://github.com/daral)**
_Wednesday Jul 12, 2017 at 17:58 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/496_

----

When exporting the Clearing Status of a project, Project Mainline State doesn't exist in the spreadsheet.

maierthomas:
<a href="https://github.com/alexbrdn"><img src="https://avatars0.githubusercontent.com/u/16780167?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [alexbrdn](https://github.com/alexbrdn)**
_Thursday Jul 13, 2017 at 07:09 GMT_

----

related to #425

maierthomas:
<a href="https://github.com/mcjaeger"><img src="https://avatars1.githubusercontent.com/u/3418977?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mcjaeger](https://github.com/mcjaeger)**
_Tuesday Oct 10, 2017 at 15:04 GMT_

----

actually it is like all items from the project-release-relation is missing, could also contain relation type, for example.

PR #253, opened 2018-04-26
feat(release): Add project mainline state to export spreadsheet (clearing status)

maierthomas:
- adds new column "project mainline state" to the release spreadsheet export file

closes eclipse/sw360#130


mcjaeger:
tested for the internal release, maybe retest is required because of different code base.

PR #255, opened 2018-04-27
fix(test): add IOException to the expected Exceptions in BlackHole test

### LINKED ISSUE #254: Blackhole tests fail randomly / on different machine setups
They can be found in `AttachmentContentDownloaderTest.java` and `RemoteAttachmentDownloaderTest.java`.

**The new reason why the tests fail:** On some environments the `BLACK_HOLE_ADDRESS = "192.0.2.0"` answers with some response code (e.g. `40?`) => this results in an `IOException` which is currently not expected.

maxhbr:
closes #254

### DIFF
--- a/libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/couchdb/AttachmentContentDownloaderTest.java
+++ b/libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/couchdb/AttachmentContentDownloaderTest.java
@@ -19,6 +19,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.runners.MockitoJUnitRunner;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.SocketException;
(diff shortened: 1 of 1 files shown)


blaumeiser-at-bosch:
Fix solves the build break in the relevant build here at Bosch, I recommend to merge

maxhbr:
IPv6 has a preserved address which acts as a black hole (see [RFC 6666](https://tools.ietf.org/html/rfc6666) )

Questions: does this break on agents which do not support IPv6?

Another option for a IPv4 black hole might be from `240.0.0.0/4`, which is "reserved for future use"  (e.g. see https://superuser.com/questions/698244/ip-address-that-is-the-equivalent-of-dev-null )

##
@maierthomas @alexbrdn this PR currently consists out of two micro commits, which could be merged, once the IPv6 version has proven to work.

blaumeiser-at-bosch:
Latest Changes did still not break the build :-). Proceed with merge from my side

maierthomas:
I will test these changes on our CI env.

libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/couchdb/AttachmentContentDownloaderTest.java
```diff
@@ -88,7 +89,9 @@ public void testABlackHoleUrl() throws Exception {
                 } catch (ExecutionException e) {
                     Throwable futureException = e.getCause();
                     assertThat(futureException, is(notNullValue()));
-                    assertTrue(futureException instanceof SocketException || futureException instanceof SocketTimeoutException);
+                    assertTrue(futureException instanceof SocketException ||
```
maierthomas: Just IOException?
SocketException and SocketTimeoutException both extends IOException

maxhbr: good point!

maierthomas:
successfully tested on siemens CI env.

PR #256, opened 2018-04-29
fix(test): CPE ID not required for release

sweetca:
Regarding conversation with Michael Jaeger, we want to proceed with release adding without CPE ID for components from NPM


mcjaeger:
yes, right, it was found that the CPE field seems should be filled out, but is in fact not available at many times. So this is about unenabling it.

maierthomas:
Alexey thank you for contributing.

Did you sign the papers for the Eclipse Contributor Agreement? ([link](https://www.eclipse.org/legal/ECA.php))

Issue #156, opened 2018-04-26
My Task Assignments view is messed up when document names are too long


maierthomas:
<a href="https://github.com/tngraf"><img src="https://avatars1.githubusercontent.com/u/1225388?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [tngraf](https://github.com/tngraf)**
_Thursday Oct 05, 2017 at 07:22 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/610_

----

The 'My Task Assignments' view is messed up when document names are too long, for example when there is a document called "Redistributable libraries for Intel Visual Fortran Composer XS 2103 for Windows Update 2 (2015.2.179)", see attached screenshot.
There are similar problems with other views. Maybe there should be a width limitation for certain columns. When the text is the column is (far) too long, it should just get cut off.
![2017-10-05 09_13_17-home - sw360](https://user-images.githubusercontent.com/1225388/31215148-b307d8f8-a9ae-11e7-93f7-f68bc354ea08.png)

maierthomas:
<a href="https://github.com/mcjaeger"><img src="https://avatars1.githubusercontent.com/u/3418977?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mcjaeger](https://github.com/mcjaeger)**
_Thursday Oct 05, 2017 at 12:55 GMT_

----

it is actually not a too long list but too much data for too narrow columns. Should be truncated.

PR #271, opened 2018-05-11
fix(homepage): Truncate long document names in homepage datatables

maierthomas:
- truncates the datatable content on homepage

closes eclipse/sw360#156


adrian-evo:
Tested and works as expected, text truncated successfully for both names and descriptions.

Issue #277, opened 2018-05-28
Add new User in couchDB is not working for new Logins


maierthomas:
Since pull request #258 it is not possible to create new Users in couchDB.
The user service validates the user objects, especially the field `"id"`.
`Id` is now set as UUID from couchDB and should not be validated.

```
2018-05-28 14:28:52 INFO  DatabaseConnector:136 - Document not found for ID: xxx
 2018-05-28 14:28:52 INFO  UserHandler:53 - User does not exist in DB
 2018-05-28 14:28:52,491 ERROR UserCacheHolder:74 - Unable to fetch user...
 java.util.concurrent.ExecutionException: org.apache.thrift.TApplicationException: getByEmail failed: unknown result
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:502)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:461)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:78)
	at com.google.common.util.concurrent.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:142)
	at com.google.common.cache.LocalCache$Segment.getAndRecordStats(LocalCache.java:2461)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2425)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2298)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2211)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4154)
2018-05-28 14:28:52,493 INFO  UserUtils:79 - Creating new user.
 2018-05-28 14:28:52,494 ERROR UserUtils:92 - Thrift exception when saving the user
 org.apache.thrift.protocol.TProtocolException: Required field 'id' was not present! Struct: User(id:null, type:user, email:xxx, userGroup:USER, externalid:, fullname:xxx, givenname:xxx, lastname:xxx, department:xxx)
	at org.eclipse.sw360.datahandler.thrift.users.User.validate(User.java:1409)
	at org.eclipse.sw360.datahandler.thrift.users.UserService$addUser_args.validate(UserService.java:5349)
	at org.eclipse.sw360.datahandler.thrift.users.UserService$addUser_args$addUser_argsStandardScheme.write(UserService.java:5408)
[...]

PR #278, opened 2018-05-28
fix(user): Change user id field to optional

maierthomas:
- change the user id field to optional

Note: Only `email` and `department` is required for a `User` object.

closes eclipse/sw360#277

(requires new sw360 release version?)


maxhbr:
It is now approved, changes were requested by accidentially clicking the wrong radio button

maierthomas:
I am just confused because github shows still `requested changes`.
Normally it should not more appear since you have approved this PR.

Issue #282, opened 2018-06-01
Clearing Team in project will be reverted in case of moderation requests


maierthomas:
The field `Clearing Team` is empty or `Unknown` after project moderation request.

_Note: default value for `clearingTeam` in project initialization_

**Thrift**
```
  public Project() {
    this.type = "project";

    this.state = org.eclipse.sw360.datahandler.thrift.projects.ProjectState.ACTIVE;

    this.projectType = org.eclipse.sw360.datahandler.thrift.projects.ProjectType.CUSTOMER;

    this.moderators = new HashSet<String>();

    this.contributors = new HashSet<String>();

    this.visbility = org.eclipse.sw360.datahandler.thrift.Visibility.BUISNESSUNIT_AND_MODERATORS;

    this.securityResponsibles = new HashSet<String>();

    this.clearingTeam = "Unknown";

  }
```

[This](https://github.com/eclipse/sw360/blob/master/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ProjectModerationRequestGenerator.java) will lead to an addition/deletion of the project `clearingTeam`!

PR #283, opened 2018-06-01
chore(projects): Remove default value for clearingTeam in projects

maierthomas:
- remove default value for `clearingTeam` = `Unknown`

closes eclipse/sw360#282


maierthomas:
_note: verify null value of clearing process_

Issue #287, opened 2018-06-07
ReadMe OSS DOCX Disclore Report Exception


maierthomas:
The ReadMe OSS disclosure can not be generated because of an exception:
```

2018-04-17 08:50:22 ERROR Sw360ThriftServlet:44 - uncaught
 java.lang.NullPointerException
        at org.ektorp.util.Documents.getAccessor(Documents.java:97)
        at org.ektorp.util.Documents.getId(Documents.java:59)
        at com.google.common.collect.Maps.uniqueIndex(Maps.java:1379)
        at com.google.common.collect.Maps.uniqueIndex(Maps.java:1341)
        at org.eclipse.sw360.datahandler.thrift.ThriftUtils.getIdMap(ThriftUtils.java:142)
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.fillTodos(LicenseDatabaseHandler.java:689)
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.getTodosByIds(LicenseDatabaseHandler.java:677)
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.fillLicenseForOrganisation(LicenseDatabaseHan$
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.getLicenseForOrganisation(LicenseDatabaseHand$
2018-04-17 08:50:22 ERROR RestControllerHelper:151 - cannot create self link for license with id: Apache-2.0
 2018-04-17 08:50:31 ERROR Sw360ThriftServlet:44 - uncaught
 java.lang.NullPointerException
        at org.ektorp.util.Documents.getAccessor(Documents.java:97)
        at org.ektorp.util.Documents.getId(Documents.java:59)
        at com.google.common.collect.Maps.uniqueIndex(Maps.java:1379)
        at com.google.common.collect.Maps.uniqueIndex(Maps.java:1341)
        at org.eclipse.sw360.datahandler.thrift.ThriftUtils.getIdMap(ThriftUtils.java:142)
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.fillTodos(LicenseDatabaseHandler.java:689)
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.getTodosByIds(LicenseDatabaseHandler.java:677)
        at org.eclipse.sw360.licenses.db.LicenseDatabaseHandler.fillLicenseForOrganisation(LicenseDatabaseHan$
[...]

PR #288, opened 2018-06-07
fix(licenses): Validate obligation list in LicenseDatabaseHandler

maierthomas:
- validates the obligation object before `ThriftUtils.getIdMap`

closes eclipse/sw360#287


kallesoranko:
Code seems to be okay

Issue #239, opened 2018-04-26
Prevent overwriting REST API configuration files


maierthomas:
<a href="https://github.com/maierthomas"><img src="https://avatars1.githubusercontent.com/u/29916928?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [maierthomas](https://github.com/maierthomas)**
_Tuesday Mar 20, 2018 at 13:44 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/789_

----

In the current setup we store the REST API configuration in the following files:
- application.yml  (auhtorization server)
- application.yml (resource server)

A new deployment of these packages will overwrite the rest configuration settings.
It should be possible to manage the rest settings outside of the webapps (e.g. sw360.properties.)

PR #292, opened 2018-06-15
feat(rest): Read client id and client secret from configuration file

maierthomas:
- enable to store client id and client secret in the sw360.properties file

**keys:**
- "rest.security.client.id"
- "rest.security.client.secret"

closes eclipse/sw360#239

### DIFF
--- a/rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/Sw360AuthorizationServer.java
+++ b/rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/Sw360AuthorizationServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2017-2018. Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
  *
@@ -11,6 +11,7 @@
 
 package org.eclipse.sw360.rest.authserver;
 
+import org.apache.log4j.Logger;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.thrift.users.UserGroup;
 import org.springframework.boot.SpringApplication;
@@ -18,25 +19,40 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.web.support.SpringBootServletInitializer;
 
+import javax.crypto.SealedObject;
+import java.io.IOException;
 import java.util.Properties;
+
+import static org.eclipse.sw360.rest.authserver.security.Sw360SecurityEncryptor.encrypt;
 
 @SpringBootApplication
 public class Sw360AuthorizationServer extends SpringBootServletInitializer {
 
+    private static final Logger log = Logger.getLogger(Sw360AuthorizationServer.class);
+
     private static final String PROPERTIES_FILE_PATH = "/sw360.properties";
-    private static final String DEFAULT_ACCESS_TIME_IN_SECONDS = "3600";
     private static final String DEFAULT_WRITE_ACCESS_USERGROUP = UserGroup.SW360_ADMIN.name();
 
-    public static final int TOKEN_ACCESS_VALIDITY;
-    public static final UserGroup WRITE_ACCESS_USERGROUP;
+    public static final String CONFIG_ACCESS_TOKEN_VALIDITY_SECONDS;
+    public static final UserGroup CONFIG_WRITE_ACCESS_USERGROUP;
+    public static final String CONFIG_CLIENT_ID;
+    public static final SealedObject CONFIG_CLIENT_SECRET;
 
     static {
         Properties props = CommonUtils.loadProperties(Sw360AuthorizationServer.class, PROPERTIES_FILE_PATH);
+        CONFIG_WRITE_ACCESS_USERGROUP = UserGroup.valueOf(props.getProperty("rest.write.access.usergroup", DEFAULT_WRITE_ACCESS_USERGROUP));
+        CONFIG_ACCESS_TOKEN_VALIDITY_SECONDS = props.getProperty("rest.access.token.validity.seconds", null);
+        CONFIG_CLIENT_ID = props.getProperty("rest.security.client.id", null);
+        CONFIG_CLIENT_SECRET = getConfigClientSecret(props);
+    }
 
-        TOKEN_ACCESS_VALIDITY = Integer.parseInt(props.getProperty(
-                "rest.token.access.validity", DEFAULT_ACCESS_TIME_IN_SECONDS));
-        WRITE_ACCESS_USERGROUP = UserGroup.valueOf(props.getProperty(
-                "rest.write.access.usergroup", DEFAULT_WRITE_ACCESS_USERGROUP));
+    private static SealedObject getConfigClientSecret(Properties props) {
+        try {
+            return encrypt(props.getProperty("rest.security.client.secret", null));
+        } catch (IOException e) {
+            log.error("Error occured while encrypting client password", e);
+            return null;
+        }
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


maierthomas:
needs unit tests

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/Sw360AuthorizationServer.java
```diff
@@ -11,32 +11,51 @@
...
+    public static final SealedObject CONFIG_CLIENT_SECRET;
+
+    private static final Logger log = Logger.getLogger(Sw360AuthorizationServer.class);
 
     static {
         Properties props = CommonUtils.loadProperties(Sw360AuthorizationServer.class, PROPERTIES_FILE_PATH);
-
-        TOKEN_ACCESS_VALIDITY = Integer.parseInt(props.getProperty(
+        CONFIG_TOKEN_ACCESS_VALIDITY = Integer.parseInt(props.getProperty(
                 "rest.token.access.validity", DEFAULT_ACCESS_TIME_IN_SECONDS));
```
maxhbr: Variable names do not match. If the variable is called `CONFIG_TOKEN_ACCESS_VALIDITY` I would expect the default value to be called `DEFAULT_TOKEN_ACCESS_VALIDITY`.

But then you loos the `_IN _SECONDS` postfix. and the name is less descriptive. Maybe needs a comment.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/Sw360AuthorizationServer.java
```diff
@@ -11,32 +11,51 @@
...
-        WRITE_ACCESS_USERGROUP = UserGroup.valueOf(props.getProperty(
+        CONFIG_WRITE_ACCESS_USERGROUP = UserGroup.valueOf(props.getProperty(
                 "rest.write.access.usergroup", DEFAULT_WRITE_ACCESS_USERGROUP));
+        CONFIG_CLIENT_ID = props.getProperty(
+                "rest.security.client.id", null);
+
+        SealedObject clientSecret;
```
maxhbr: This variable is unnecessary.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360AuthorizationServerConfiguration.java
```diff
@@ -70,13 +71,13 @@ public void configure(AuthorizationServerSecurityConfigurer oauthServer) throws
     @Override
     public void configure(ClientDetailsServiceConfigurer clients) throws Exception {
         clients.inMemory()
-                .withClient(clientId)
+                .withClient(CONFIG_CLIENT_ID == null ? clientId : CONFIG_CLIENT_ID)
```
maxhbr: I am not sure whether this is the correct point to inject the overriding configuration. It does not modify the field in the class and might lead to confusing behaviour if the field is used somewhere else. Maybe you can overwrite it on the `@Value` annotation level?

(Same for the secret)

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360SecurityEncryptor.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class Sw360SecurityEncryptor {
+
+    private static final Logger log = Logger.getLogger(Sw360SecurityEncryptor.class);
+
+    private static final byte[] SECRET_SPEC_KEY = "sw360-rest-super-secret-password".getBytes();
+    private static final String SECRET_SPEC_TRANSFORMATION = "AES";
+
+    public static SealedObject encrypt(String password) throws IOException {
+        if (password != null && !password.isEmpty()) {
```
maxhbr: Why don't you support the empty password? Looks like an unnecessary restriction to me.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360SecurityEncryptor.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.apache.log4j.Logger;
+
+import javax.crypto.*;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.IOException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+
+public class Sw360SecurityEncryptor {
```
maxhbr: Encrypting with a hard coded password is only obfuscation? Maybe the class could better be called `Sw360SecurityObfuscator`? :wink:

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360SecurityEncryptor.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private static final Logger log = Logger.getLogger(Sw360SecurityEncryptor.class);
+
+    private static final byte[] SECRET_SPEC_KEY = "sw360-rest-super-secret-password".getBytes();
```
maxhbr: might be better called `PUBLIC_SPEC_KEY` :wink:

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/Sw360AuthorizationServer.java
maierthomas: Changed to `DEFAULT_TOKEN_ACCESS_VALIDITY` and added a comment (validity time in seconds)

maierthomas: `CONFIG_CLIENT_SECRET` is defined as final constant.
I can only initialize this constant value for one time. (in constructor)
To consider a correct exception handling I need to create this temp variable `clientSecret`.
I renamed `clientSecret` to `tmpClientSecret`.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/Sw360AuthorizationServer.java
maierthomas: `CONFIG_CLIENT_SECRET` is defined as final constant.
I can only initialize this constant value for one time. (in constructor)
To consider a correct exception handling I need to create this temp variable `clientSecret`.
I renamed `clientSecret` to `tmpClientSecret`.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360AuthorizationServerConfiguration.java
maierthomas: I tried to overwrite the @value annotation directly.
But this is not possible.

As result I created a new @PostConstruct method `postSw360AuthorizationServerConfiguration` to overwrite the @value annotation fields.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360SecurityEncryptor.java
maierthomas: Yes it should support also a empty password.
Removed `!password.isEmpty()` condition.

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360SecurityEncryptor.java
maierthomas: changed to `PUBLIC_SPEC_KEY`.
Your are right, its not really secret :stuck_out_tongue_winking_eye:

rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360SecurityEncryptor.java
maierthomas: hmm.. I am absolutely not familiar with `Sw360SecurityObfuscator` :hear_no_evil:

adrian-evo:
Tested and works as expected. Client id and secret taken from properties file.

Issue #297, opened 2018-06-19
Moderation footer width is not equal to table widh


maierthomas:
The length of the footer line is not equal to the moderation datatable (columns).

PR #298, opened 2018-06-19
chore(moderation): Adjust footer length with datatable width (columns)

maierthomas:
fix footer length (width) in moderation view

closes eclipse/sw360#297

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
@@ -56,36 +56,18 @@
             <div class="tab-content span10">
                 <div id="tab-Open" class="tab-pane">
                     <table id="moderationsTable" cellpadding="0" cellspacing="0" border="0" class="display" style="width:100%">
-                        <colgroup>
-                            <col style="width: 5%;" />
-                            <col style="width: 10%;" />
-                            <col style="width: 20%;" />
-                            <col style="width: 15%;" />
-                            <col style="width: 15%;" />
-                            <col style="width: 30%;" />
-                            <col style="width: 5%;" />
-                        </colgroup>
                         <tfoot>
                         <tr>
-                            <th colspan="7"></th>
+                            <th colspan="8"></th>
                         </tr>
                         </tfoot>
                     </table>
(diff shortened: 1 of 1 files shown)


frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
```diff
@@ -57,35 +57,37 @@
                 <div id="tab-Open" class="tab-pane">
                     <table id="moderationsTable" cellpadding="0" cellspacing="0" border="0" class="display" style="width:100%">
                         <colgroup>
-                            <col style="width: 5%;" />
+                            <col style="width: 90px;" />
```
maxhbr: mixing `px` and `%` is normally not a good idea

frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
```diff
@@ -209,6 +211,12 @@
                 {title: "Moderators", render: {display: renderModeratorsListExpandable}},
                 {title: "State"},
                 {title: "Actions"}
+            ],
+            columnDefs: [
+                {
+                    width: "90px", "targets": [0],
```
maxhbr: Why does datables needs the width here? does it ignoire the width stated in the html above?

frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
maxhbr: How does it behave when the window is scaled?

maierthomas: yes mixing of px and % is a bad idea.
Anyway I don't like to specify each column width.
I removed this code, thus the browser calculates the optimal column width.
I looks much better, especially the date and type columns are now calculates in one line.

frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
maierthomas: just a test, removed code.

adrian-evo:
Tested. Footer adjusted successfully.

Issue #299, opened 2018-06-20
File download abort for big attachments (100mb) in Chrome


maierthomas:
Download of big attachments (>100mb) is not possible in Chrome.
-> `Network error` occurs!

maierthomas:
reference #146

PR #306, opened 2018-06-27
fix(attachments): Set content encoding to identity in case of gzip files

maierthomas:
Set content encoding to identity for gzip files.

closes eclipse/sw360#299

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/AttachmentPortletUtils.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/AttachmentPortletUtils.java
@@ -129,8 +129,7 @@ public class AttachmentPortletUtils extends AttachmentFrontendUtils {
         if(attachments.size() == 1){
             filename = attachments.get(0).getFilename();
             contentType = attachments.get(0).getContentType();
-            if (contentType.equalsIgnoreCase("application/gzip")) ;
-            {
+            if (contentType.equalsIgnoreCase("application/gzip")) {
                 // In case of downloads with gzip file extension (e.g. *.tar.gz)
                 // the client should receive the origin file.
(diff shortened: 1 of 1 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/AttachmentPortletUtils.java
```diff
@@ -129,6 +129,14 @@ private void serveAttachmentBundle(List<AttachmentContent> attachments, Resource
         if(attachments.size() == 1){
             filename = attachments.get(0).getFilename();
             contentType = attachments.get(0).getContentType();
+            if (contentType.equalsIgnoreCase("application/gzip")) ;
```
kallesoranko: is this if-statement in correct syntax? Is it correct to have semicolon in the end?

alexbrdn: Definitely a mistake, but I'm not sure whether it will cause any wrong behavior. Should be corrected anyways

adrian-evo:
Unfortunately it is not working for me. E.g. a simple attachment of type text (Readme-test.txt) is downloaded with content changed (Readme-test(1).txt). On the master branch the file is downloaded correctly.

[Readme-test.txt](https://github.com/eclipse/sw360/files/2144595/Readme-test.txt)

[Readme-test(1).txt](https://github.com/eclipse/sw360/files/2144598/Readme-test.1.txt)

adrian-evo:
Tested with several file types, including gzip, and download works fine. Also for big attachments.

adrian-evo:
Unfortunately it seems big attachment download works in Chrome, however auto unpack does not seems to be fixed.

PR #311, opened 2018-07-03
fix(search): impose the defined search limit on all searches by default

alexbrdn:
the search limit used to be not applied to release, user, and vendor search


mcjaeger:
right. itis now at 4 occasions, maybe two of them are redundant:

https://github.com/eclipse/sw360/search?q=setResultLimit%28DatabaseSettings.LUCENE_SEARCH_LIMIT%29&unscoped_q=setResultLimit%28DatabaseSettings.LUCENE_SEARCH_LIMIT%29

Issue #315, opened 2018-07-11
License information could not be generated


maierthomas:
Readme has only the following text: License information could not be generated.
**An exception occured: java.lang.NullPointerException**

_Trace:_
```
 java.lang.NullPointerException
	at org.eclipse.sw360.licenseinfo.outputGenerators.OutputGenerator.mergeLicenseInfoParsingResults(OutputGenerator.java:124)
	at java.util.HashMap.merge(HashMap.java:1253)
	at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)
	at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1374)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
```

PR #316, opened 2018-07-11
fix(licenseinfo): Update merge handling for licenseInfo objects

maierthomas:
- fix the null pointer exception in case of merging licenseInfo objects.

closes eclipse/sw360#315

### DIFF
--- a/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
+++ b/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
@@ -32,7 +32,9 @@ import org.jetbrains.annotations.NotNull;
 import java.io.StringWriter;
 import java.util.*;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public abstract class OutputGenerator<T> {
     protected static final String VELOCITY_TOOLS_FILE = "velocity-tools.xml";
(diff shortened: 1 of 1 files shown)


backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
```diff
@@ -119,10 +119,29 @@ protected LicenseInfoParsingResult mergeLicenseInfoParsingResults(LicenseInfoPar
                 !getComponentLongName(r1).equals(getComponentLongName(r2))){
             throw new IllegalArgumentException("Only successful parsing results for the same release can be merged");
         }
+
         LicenseInfoParsingResult r = new LicenseInfoParsingResult(r1);
-        r.getLicenseInfo().getLicenseNamesWithTexts().addAll(r2.getLicenseInfo().getLicenseNamesWithTexts());
-        r.getLicenseInfo().getCopyrights().addAll(r2.getLicenseInfo().getCopyrights());
-        r.getLicenseInfo().getFilenames().addAll(r2.getLicenseInfo().getFilenames());
+        if (r2.getLicenseInfo().isSetLicenseNamesWithTexts()) {
```
alexbrdn: I would posit that this option is easier to read:
```java
        Optional.of(r).map(LicenseInfoParsingResult::getLicenseInfo).map(LicenseInfo::getLicenseNamesWithTexts)
                .ifPresent(set -> set.addAll(Optional.of(r2)
                        .map(LicenseInfoParsingResult::getLicenseInfo)
                        .map(LicenseInfo::getLicenseNamesWithTexts)
                        .orElse(Collections.emptySet())));
```

What do you think?

maierthomas: Hmmm I had a similar solution with a lambda function to solve that problem.
But I thought its not really readable.
I would say that your solution is not more clearly. :)

alexbrdn: let's get a second opinion. @maxhbr, what do you say?

maxhbr: @alexbrdn : I see two problems with your code:
- it contains duplicated code: `.map(LicenseInfoParsingResult::getLicenseInfo).map(LicenseInfo::getLicenseNamesWithTexts)`
- it only works if `r.getLicenseInfo().isSetLicenseNamesWithTexts() == true` (otherwise nothing is done, due to the first `ifPresent`)

I would write
```
r.getLicenseInfo().setLicenseNamesWithTexts().addAll(
    nullToEmptyCollection(
        r2.getLicenseInfo().getLicenseNamesWithTexts()));
```
But that also expects that `r.getLicenseInfo().isSetLicenseNamesWithTexts() == true`.

maierthomas: r.getLicenseInfo().setLicenseNamesWithTexts().addAll(
    nullToEmptyCollection(
        r2.getLicenseInfo().getLicenseNamesWithTexts()));
This will create an empty collection in case of null.
Why we should create a empty collection if the origin is null.

Yes both of the LicenseInfo objects can be null
`r.getLicenseInfo().getXXX` and `r2.getLicenseInfo().getXXX` can be null.

- 1. I am verify if i need to merge something `if (r2.getLicenseInfo().isSetLicenseNamesWithTexts())`
- 2. Verify If the origin object == null (set list to origin)
- 3. If the origin object != null just add the items which we want to merge

maierthomas:
tested successfully in our siemens test system (foss360).

PR #317, opened 2018-07-11
feat(rest) Batch-Deletion for components

blaumeiser-at-bosch:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
@@ -1,6 +1,6 @@
 /*
  * Copyright Siemens AG, 2017-2018.
- * Copyright Bosch Software Innovations GmbH, 2017.
+ * Copyright Bosch Software Innovations GmbH, 2017-2018.
  * Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
@@ -10,18 +10,21 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
+
 package org.eclipse.sw360.rest.resourceserver.component;
 
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.thrift.TException;
+import org.eclipse.sw360.datahandler.thrift.RequestStatus;
 import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
 import org.eclipse.sw360.datahandler.thrift.components.Component;
 import org.eclipse.sw360.datahandler.thrift.components.Release;
 import org.eclipse.sw360.datahandler.thrift.users.User;
 import org.eclipse.sw360.datahandler.thrift.vendors.Vendor;
 import org.eclipse.sw360.rest.resourceserver.attachment.Sw360AttachmentService;
+import org.eclipse.sw360.rest.resourceserver.core.MultiStatus;
 import org.eclipse.sw360.rest.resourceserver.core.HalResource;
 import org.eclipse.sw360.rest.resourceserver.core.RestControllerHelper;
 import org.eclipse.sw360.rest.resourceserver.release.Sw360ReleaseService;
@@ -117,6 +120,25 @@ public class ComponentController implements ResourceProcessor<RepositoryLinksRes
         componentService.updateComponent(sw360Component, user);
         HalResource<Component> userHalResource = createHalComponent(sw360Component, user);
         return new ResponseEntity<>(userHalResource, HttpStatus.OK);
+    }
+
+    @PreAuthorize("hasAuthority('WRITE')")
+    @RequestMapping(value = COMPONENTS_URL + "/{ids}", method = RequestMethod.DELETE)
+    public ResponseEntity<List<MultiStatus>> deleteComponents(
+            @PathVariable("ids") List<String> idsToDelete, OAuth2Authentication oAuth2Authentication) throws TException {
+        User user = restControllerHelper.getSw360UserFromAuthentication(oAuth2Authentication);
+        List<MultiStatus> results = new ArrayList<>();
+        for(String id:idsToDelete) {
+            RequestStatus requestStatus = componentService.deleteComponent(id, user);
+            if(requestStatus == RequestStatus.SUCCESS) {
+                results.add(new MultiStatus(id, HttpStatus.OK));
+            } else if(requestStatus == RequestStatus.IN_USE) {
+                results.add(new MultiStatus(id, HttpStatus.CONFLICT));
+            } else {
+                results.add(new MultiStatus(id, HttpStatus.INTERNAL_SERVER_ERROR));
+            }
+        }
+        return new ResponseEntity<>(results, HttpStatus.MULTI_STATUS);
     }
 
     @PreAuthorize("hasAuthority('WRITE')")
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/Sw360ComponentService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/Sw360ComponentService.java
@@ -1,6 +1,6 @@
 /*
  * Copyright Siemens AG, 2017-2018.
- * Copyright Bosch Software Innovations GmbH, 2017.
+ * Copyright Bosch Software Innovations GmbH, 2017-2018.
  * Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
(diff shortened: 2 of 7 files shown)


hengel2810:
This pull-request implements the functionality to delete multiple components with one request and get a response which contains an error code for each component. The request response status code is always 207 (Multi-Status)

Request: DELETE .../resource/api/components/componentId1,componentId2

Response:
```
[
    {
        "resourceId":"componentId1",
        "status":500
    },
    {
        "resourceId":"componentId2",
        "status":200
    }
]
```

mcjaeger:
Hi, I was just wondering (and I do not know): will it delete releases? How is the handling if there are releases tied to the component?

hengel2810:
It will also delete releases and attachments which belong to the component. I'm using the already existing `deleteComponent`-method in `backend/src-common/.../datahandler/db/ComponentDatabaseHandler` (line 781) for this.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -104,6 +107,25 @@
...
+            @PathVariable("ids") List<String> idsToDelete, OAuth2Authentication oAuth2Authentication) throws TException {
+        User user = restControllerHelper.getSw360UserFromAuthentication(oAuth2Authentication);
+        List<BatchOperationResult> results = new ArrayList<>();
+        for(String id:idsToDelete) {
+            RequestStatus requestStatus = componentService.deleteComponent(id, user);
+            HttpStatus status;
+            if(requestStatus == RequestStatus.SUCCESS) {
+                status = HttpStatus.OK;
+            } else {
+                status = HttpStatus.INTERNAL_SERVER_ERROR;
```
maierthomas: If a component contains releases (release.size >0) the method call `componentService.deleteComponent`
will return `RequestStatus.IN_USE`. 
Instead of `500 - internal server error` we could throw `409 conflict`, what do you think?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/Sw360ComponentService.java
```diff
@@ -69,6 +69,11 @@ public RequestStatus updateComponent(Component component, User sw360User) throws
         return requestStatus;
     }
 
+    public RequestStatus deleteComponent(String componentId, User sw360User) throws TException {
```
maierthomas: Maybe update license header to `Copyright Bosch Software Innovations GmbH, 2018`

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/BatchOperationResult.java
```diff
@@ -0,0 +1,49 @@
+/*
```
maierthomas: _Suggestions?_

Maybe only return the component delete failures to the user?
We could reuse `RestExceptionHandler.java` to handle this scenario.

1. `public ErrorMessage(String message, HttpStatus httpStatus)` (overload the constructor)
2.  `results.add(new ErrorMessage("Could not delete component id...", HttpStatus.CONFLICT));`

What do you think about this?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/BatchOperationResult.java
```diff
@@ -0,0 +1,49 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.sw360.rest.resourceserver.core;
+
+import org.springframework.http.HttpStatus;
+
+public class BatchOperationResult {
```
maierthomas: Maybe rename to a more 'generic' name like `MultiStatus` ?

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/TestHelper.java
```diff
@@ -1,38 +1,22 @@
 /*
  * Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
+ * Copyright Bosch Software Innovations GmbH, 2018.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
-
```
maierthomas: :stuck_out_tongue_winking_eye:

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -104,6 +107,25 @@
...
+    @RequestMapping(value = COMPONENTS_URL + "/{ids}", method = RequestMethod.DELETE)
+    public ResponseEntity<List<BatchOperationResult>> deleteComponents(
+            @PathVariable("ids") List<String> idsToDelete, OAuth2Authentication oAuth2Authentication) throws TException {
+        User user = restControllerHelper.getSw360UserFromAuthentication(oAuth2Authentication);
+        List<BatchOperationResult> results = new ArrayList<>();
+        for(String id:idsToDelete) {
+            RequestStatus requestStatus = componentService.deleteComponent(id, user);
+            HttpStatus status;
+            if(requestStatus == RequestStatus.SUCCESS) {
+                status = HttpStatus.OK;
```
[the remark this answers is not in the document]
hengel2810: I've changed that, status is added directly now.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
hengel2810: I'm now using `409 conflict` in this case

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/BatchOperationResult.java
hengel2810: Changed the class name to `MultiStatus`

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/BatchOperationResult.java
hengel2810: Do you have a specific reason for that ? 
I would prefer that the succeeded deletions are also in the response. I think it's not so much overhead that it's necessary to prune the response. 

Maybe we can discuss this in the round during the next sw360 meeting.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/Sw360ComponentService.java
hengel2810: License header is updated.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/BatchOperationResult.java
maierthomas: No I don't have a specific reason for that.
Of course you are absolutely right here and it was just a suggestions.
The idea was to reuse some existing class for that but it would be better to have a complete result with all entries in the response.

maierthomas:
To clarification: 

What is the basic use case of this pull request.
Currently only empty components (without linked releases) can be deleted over this batch operation.
Should this pull request also delete attached releases?

adrian-evo:
Tested and I could delete multiple components at once, including attachments. Also it works only with admin user rights and not normal user which is as expected.

hengel2810:
Initially it was planned to delete all releases linked to the component, but there's still the problem with projects linked to these releases. 
The `deleteComponent` method in `ComponentDatabaseHandler` checks if a component has releases and wouldn't delete the component if there are releases, but there's also a method called `removeReleaseAndCleanUp` which is never called. In this method there is the following comment: `//TODO notify using projects!?? Or stop if there are any`. 

I would suggest that we first use this version where releases aren't deleted and I'll implement a future version where it's possible to add a `force` parameter to the request which deletes all linked releases of a component and also remove these releases from the projects.

maierthomas:
yes I also had the idea to add a parameter to force the deletion of all releases.
Its okay for me to start with that implementation and enhance this parameter later in another pull request.
But we should describe the current behavior in the Rest API doc (only component without releases can be deleted).

mcjaeger:
How about
* merging this PR now
* component delete deletes components, if there are no references, which means that there are releases
* another endpoint deletes releases if there are no references, which means projects using these
* another endpoint deletes projects, if no references are there (being used as subproject)

maierthomas:
@hengel2810 please rebase on the current master and fix the conflicted files! thank you!

hengel2810:
Rebased with current master and conflicts resolved.

PR #318, opened 2018-07-11
feat(rest) Route PATCH /component/{id} added to update an existing component

blaumeiser-at-bosch:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
@@ -106,6 +106,20 @@ public class ComponentController implements ResourceProcessor<RepositoryLinksRes
     }
 
     @PreAuthorize("hasAuthority('WRITE')")
+    @RequestMapping(value = COMPONENTS_URL + "/{id}", method = RequestMethod.PATCH)
+    public ResponseEntity<Resource<Component>> patchComponent(
+            @PathVariable("id") String id,
+            @RequestBody Component updateComponent,
+            OAuth2Authentication oAuth2Authentication) throws TException {
+        User user = restControllerHelper.getSw360UserFromAuthentication(oAuth2Authentication);
+        Component sw360Component = componentService.getComponentForUserById(id, user);
+        sw360Component = this.restControllerHelper.updateComponent(sw360Component, updateComponent);
+        componentService.updateComponent(sw360Component, user);
+        HalResource<Component> userHalResource = createHalComponent(sw360Component, user);
+        return new ResponseEntity<>(userHalResource, HttpStatus.OK);
+    }
+
+    @PreAuthorize("hasAuthority('WRITE')")
     @RequestMapping(value = COMPONENTS_URL, method = RequestMethod.POST)
     public ResponseEntity<Resource<Component>> createComponent(
             OAuth2Authentication oAuth2Authentication,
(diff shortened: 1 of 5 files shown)


hengel2810:
This pull-request implements the functionality to update an existing component. All properties in the PATCH request body should be updated with this request.

Request: PATCH .../resource/api/components/componentId

Body:
```
{
    "name": "Updated Name"
}
```

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -104,6 +104,20 @@
         return new ResponseEntity<>(userHalResource, HttpStatus.OK);
     }
 
+    @PreAuthorize("hasAuthority('WRITE')")
+    @RequestMapping(value = COMPONENTS_URL + "/{id}", method = RequestMethod.PATCH)
+    public ResponseEntity<Resource<Component>> patchComponent(
+            @PathVariable("id") String id,
+            @RequestBody Component updateComponent,
```
maierthomas: Do we ignore wrong fields?
In case of wrong body request (component) which contains irregular component data do we just skip them or maybe throw an exception? 
e.g. `400 bad request` or `422 unprocessable entity`

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -222,6 +221,16 @@ private void addEmbeddedProject(HalResource halResource, Project project) {
         halResource.addEmbeddedResource("sw360:projects", halProject);
     }
 
+    public Component updateComponent(Component componentToUpdate, Component requestBodyComponent) {
+        for(Component._Fields field:Component._Fields.values()) {
+            Object fieldValue = requestBodyComponent.getFieldValue(field);
+            if(fieldValue != null) {
+                componentToUpdate.setFieldValue(field, fieldValue);
```
maierthomas: This will allow to overwrite all component fields especially name, id..
Should we restrict some fields here?

Also there is no validation for some fields like links to another resources (e.g. users).
I think a validation is not necessary here but we should keep that in mind.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/ComponentTest.java
```diff
@@ -75,4 +87,43 @@ public void should_get_all_components() throws IOException {
...
+                        String.class);
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+        JsonNode responseBodyJsonNode = new ObjectMapper().readTree(response.getBody());
+        assertEquals(responseBodyJsonNode.get("name").textValue(), updatedComponentName);
+        assertNull(responseBodyJsonNode.get("invalid_property"));
+
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
```
maierthomas: Why  `SuppressWarnings` annotation here?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
hengel2810: At the moment jackson is parsing the body to a `component` object and all non-component properties will be automatically ignored by jackson without throwing an exception.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
hengel2810: If the request body contains a new `id` the `updateComponent` method throws an exception and the request returns `500`, so I think it's necessary to restrict the field `id` there and simply ignore it. 

I don't know whether there should be more restrictions, I think it's more a strategic question than a technical one if it's necessary to update for example the name of a component.

adrian-evo:
Tested and I could update different component's fields successfully.

maierthomas:
@hengel2810 please rebase onto current master and fix the conflict. thank you!

hengel2810:
Rebased with current master and conflicts resolved.

PR #319, opened 2018-07-11
feat(rest) Specify required fields for components in GET /components

blaumeiser-at-bosch:


hengel2810:
This pull-request implements the functionality to define a `fields` parameter in the GET .../components request. All properties in this `fields` parameter are returned for the components in the response if they have these properties. This allows a more dynamic use of this route for clients using this API. 


Request: GET .../resources/api/components?fields=description,createdBy

Response:
```
{
    "_embedded": {
        "sw360:components": [
            {
                "name": "component 1",
                "description": "Component for testing",
                "createdBy": "John Doe",
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/resource/api/components/0ab383d191df0e69bb6a2c69e700096a"
                    }
                }
            }
        ]
    },
    "_links": {
        "curies": [
            {
                "href": "http://localhost:8080/resource/docs/html5/{rel}.html",
                "name": "sw360",
                "templated": true
            }
        ]
    }
}
```

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/PropertyKeyMapping.java
```diff
@@ -0,0 +1,17 @@
+package org.eclipse.sw360.rest.resourceserver.core;
```
maierthomas: add license header

maierthomas: _to understand the usage_
This file is only used in case of changing thrift field names with jackson by using `@JsonProperty`?

hengel2810: The problem with the license header is fixed now and I've added the documentation to the `components.adoc`

hengel2810: Yes, the PropertyKeyMapping-Class only contains the logic and keys to map between thrift and jackson field names.

hengel2810:
Yes, `createdBy` is one of the fields which is listed in `@JsonIgnoreProperties` in the `JacksonCustomizations`-File. All the listed properties are ignored explicitly during the parsing process.

PR #330, opened 2018-07-24
chore(rakefile): Drop the Rakefile, since it is unused

maxhbr:


alexbrdn:
why doesn't the commit actually delete the rakefile?

maxhbr:
Sorry, I have pushed to the wrong remote. Is fixed now.

maierthomas:
:+1:

PR #332, opened 2018-08-01
feat(rest) Paging/Sorting for GET /components, response contains pagi…

blaumeiser-at-bosch:
…nation links and metadata

feat(datahandler) Added Paging/Sorting functionality


hengel2810:
The GET /components route should use pagination. By default each page should have 10 items per page.
The first page has the index 0. It should be possible to use one of the following sort parameter: name, createdOn, createdBy, componentType

The result should contain meta data for the pagination. (links, itemCounts)

GET .../api/components?page=1&page_entries=5&sort=name,desc

adrian-evo:
Tested, works as expected.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/resourcelists/ResourceComparatorGenerator.java
```diff
@@ -0,0 +1,57 @@
+/*
```
maierthomas: I like this solution with this `ResourceComperatorGenerator` class.
We can easily enhance it by adding the other sw360 resource types.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/resourcelists/WrongPaginationParameterException.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.sw360.datahandler.resourcelists;
+
+public class WrongPaginationParameterException extends Exception {
```
maierthomas: I would prefer to name it like `PaginationParameterException` or just `PaginationException`

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/resourcelists/ResourceListConstants.java
```diff
@@ -0,0 +1,29 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.sw360.datahandler.resourcelists;
+
+public class ResourceListConstants {
+
+    public static final String COMPONENT_CLASS_NAME = "component";
```
maierthomas: We should use these type names in `SW360Constants`

```
    // Proper values of the "type" member to deserialize to CouchDB
    public static final String TYPE_OBLIGATION = "obligation";
    public static final String TYPE_TODO = "todo";
    public static final String TYPE_RISKCATEGORY = "riskCategory";
    public static final String TYPE_RISK = "risk";
    public static final String TYPE_LICENSETYPE = "licenseType";
    public static final String TYPE_LICENSE = "license";
    public static final String TYPE_VENDOR = "vendor";
    public static final String TYPE_USER = "user";
    public static final String TYPE_COMPONENT = "component";
    public static final String TYPE_RELEASE = "release";
    public static final String TYPE_ATTACHMENT = "attachment";
    public static final String TYPE_PROJECT = "project";
    public static final String TYPE_MODERATION = "moderation";
```

Or rename and reuse this `MAP_DOCUMENT_TYPE_STRING` (ThriftEnumUtils)

```
    private static final ImmutableMap<DocumentType, String> MAP_DOCUMENT_TYPE_STRING = ImmutableMap.of(
            DocumentType.COMPONENT, "component" ,
            DocumentType.RELEASE, "release" ,
            DocumentType.PROJECT, "project",
            DocumentType.LICENSE, "license",
            DocumentType.USER, "user"
    );
```

libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/resourcelists/ResourceComparatorGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class ResourceComparatorGeneratorTest {
+
+    ResourceComparatorGenerator resourceComparatorGenerator;
+    String componentResourceType;
+
+    @Before
+    public void setUp() throws Exception {
+        resourceComparatorGenerator = new ResourceComparatorGenerator();
+        componentResourceType = new Component().getType();
```
maierthomas: Please use type name in `SW360Constants`

libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/resourcelists/ResourceComparatorGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Test
+    public void checkComponentComparators() throws ResourceClassNotFoundException{
+        assertNotNull(resourceComparatorGenerator.generateComparator(componentResourceType, "name"));
```
maierthomas: Check also if this comparator sorts by Component::getName

libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/resourcelists/ResourceListControllerTest.java
```diff
@@ -0,0 +1,123 @@
...
+    ResourceComparatorGenerator resourceComparatorGenerator;
+    List<Component> unsortedComponents;
+    String componentResourceType;
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Before
+    public void setUp() throws Exception {
+        componentResourceType = new Component().getType();
```
maierthomas: `SW360Constants`

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/Sw360ResourceServer.java
```diff
@@ -37,4 +40,15 @@ protected SpringApplicationBuilder configure(SpringApplicationBuilder builder) {
     public static void main(String[] args) {
         SpringApplication.run(Sw360ResourceServer.class, args);
     }
+
+    @Bean
+    public RepositoryRestConfigurer repositoryRestConfigurer() {
+        return new RepositoryRestConfigurerAdapter() {
+            @Override
+            public void configureRepositoryRestConfiguration(RepositoryRestConfiguration config) {
+                config.setLimitParamName("page_entries");
```
maierthomas: Please extract this string to a constant.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/Sw360ResourceServer.java
```diff
@@ -37,4 +40,15 @@ protected SpringApplicationBuilder configure(SpringApplicationBuilder builder) {
...
+                config.setDefaultPageSize(10);
```
maierthomas: Should we make this value configurable?
e.g. add this to rest .yml or sw360.properties?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -73,20 +78,29 @@
...
+        List<Component> allComponents = new ArrayList<Component>();
         if (name != null && !name.isEmpty()) {
-            sw360Components.addAll(componentService.searchComponentByName(name));
+            allComponents.addAll(componentService.searchComponentByName(name));
         } else {
-            sw360Components.addAll(componentService.getComponentsForUser(sw360User));
+            allComponents.addAll(componentService.getComponentsForUser(sw360User));
         }
 
+        PaginationOptions paginationOptions = restControllerHelper.paginationOptionsFromPageable(pageable, new Component().getType());
```
maierthomas: If `pageable` is null you will call the complete method sequence for pagination.
You could verify here if you need to page / filter the result.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -58,13 +68,86 @@
...
+        paginationLinks.add(new Link(createPaginationLink(baseUrl, paginationResult.getTotalPageCount() - 1, paginationOptions.getPageSize()),"last"));
+
+        return paginationLinks;
+    }
+
+    private String createPaginationLink(String baseUrl, int page, int pageSize) {
+        return baseUrl + "?page=" + page + "&page_entries=" + pageSize;
+    }
+
+    private String getAPIBaseUrl() throws URISyntaxException {
```
maierthomas: I like this.. We could reuse it in our different ResourceProcessor's

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -58,13 +68,86 @@
...
+                paginationResult.getTotalCount(),
+                paginationResult.getTotalPageCount());
+        return new PagedResources<>(resources, pageMetadata, pagingLinks);
+    }
+
+    private List<Link> getPaginationLinks(PaginationResult paginationResult, String baseUrl) {
+        PaginationOptions paginationOptions = paginationResult.getPaginationOptions();
+        List<Link> paginationLinks = new ArrayList<>();
+
+        paginationLinks.add(new Link(createPaginationLink(baseUrl, 0, paginationOptions.getPageSize()),"first"));
```
maierthomas: Maybe extract the String previos, next, last... ?page and &page_entries to const.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestExceptionHandler.java
```diff
@@ -43,8 +45,8 @@
         return new ResponseEntity<>(new ErrorMessage(e, HttpStatus.FORBIDDEN), HttpStatus.FORBIDDEN);
     }
 
-    @ExceptionHandler(ResourceNotFoundException.class)
-    public ResponseEntity<ErrorMessage> handleResourceNotFound(ResourceNotFoundException e) {
+    @ExceptionHandler({ResourceNotFoundException.class, WrongPaginationParameterException.class})
```
maierthomas: Hmm I am unhappy to send 404 in this case.
If we want to send a Client Error we should use 406/409 otherwise we should use 500.

hengel2810: What do you think about `400 Bad Request` ?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
hengel2810: I think `pageable` is always `!=NULL` at this point, because there's a defined default value for `page_entries` in `Sw360ResourceServer`. Even if there are no paging parameters in the request, the pageable object will contain `page=0` and `page_entries=10`

hengel2810:
I've updated the PR based on your review.

PR #333, opened 2018-08-01
feat(rest) Route PATCH /release/{id} added to update an existing release

blaumeiser-at-bosch:


hengel2810:
This pull-request implements the functionality to update an existing release. All properties in the PATCH request body should be updated with this request. Unknown properties or properties annotated as ignored in the `JacksonCustomizations.java` will be ignored.

Request: PATCH .../resource/api/releases/releaseId

Body:
```
{
    "name": "Updated Name"
}
```

PR #334, opened 2018-08-01
feat(rest) Added route DELETE /releases/{ID},{ID2} to delete releases

blaumeiser-at-bosch:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/release/ReleaseController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/release/ReleaseController.java
@@ -17,12 +17,14 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.log4j.Logger;
 import org.apache.thrift.TException;
+import org.eclipse.sw360.datahandler.thrift.RequestStatus;
 import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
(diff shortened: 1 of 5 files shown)


hengel2810:
This pull-request implements the functionality to delete multiple releases with one request and get a response which contains an error code for each release. The request response status code is always 207 (Multi-Status). It's only possible to delete releases which aren't linked to other resources.

Request: DELETE .../resource/api/releases/releaseId1,releaseId2

Response:
```
[
    {
        "resourceId":"releaseId1",
        "status":500
    },
    {
        "resourceId":"releaseId2",
        "status":200
    },
    {
        "resourceId":"releaseId3",
        "status":409
    }   
]
```

maierthomas:
@hengel2810 please rebase your branch onto the current master! thank you

alexbrdn:
Shouldn't wrong release id return 404 instead of 500?

maierthomas:
Yes this would be the best solution to return 404 NOT_FOUND in this case.
Currently the REST API returns always 500 in this case because our thrift services throw a TException.

hengel2810:
That's true, 404 would be a better error code for that. I'll fix that and rebase the branch afterwards.

maierthomas:
@hengel2810 I would say 500 is okay in this case.
Because you have to change the thrift service implementation If you want to return 404.
You also should change the DELETE Batch method for component.
If you want we can write a small issue for this.

hengel2810:
@maierthomas Ok, I've rebased the branch and it would be good to create an issue for the status code problem.

Issue #340, opened 2018-08-14
Show object identifier (id) in REST resources


maierthomas:
Currently in the SW360 REST API all identifiers (ids) are ignored by jackson customization.
The id should be visible in the embedded object or at least in the detail information 'GET' request of a sw360 resource.

mcjaeger:
actually, there is also the idea of showing the object identifier in the Web UI.

mcjaeger:
and that has also been covered by now, for example:

<img width="842" alt="Screenshot 2021-03-25 at 22 29 10" src="https://user-images.githubusercontent.com/3418977/112546315-9698e200-8db9-11eb-857b-1cd7f6f013f1.png">

PR #341, opened 2018-08-15
chore(rest): Show object identifier in REST resources

maierthomas:
- shows object ids for REST resources

closes eclipse/sw360#340


mcjaeger:
(from discussion), rather preferred to have the couchdb id out of the APi and use in the del function the translator instead.

Issue #337, opened 2018-08-06
Advanced Search for Group in Project View is not working probably


maierthomas:
If you search for Groups or / Tags in Advanced Search for projects the result is not correct.
It seems the advanced search for groups is added by wildcards.
Please verify the search logic for groups and tags.

PR #345, opened 2018-08-17
fix(projects): Enable phrase search for group and tag in projects

maierthomas:
For advanced search with `group` or `tag` phrase searches should be used.

closes eclipse/sw360#337


maierthomas:
tested successfully on my local setup

PR #347, opened 2018-08-20
fix(licenseInfo): remove the hidden license text input field from license info generation page

alexbrdn:
Sending license texts in the request to generate license info file is unnecessary. In some cases, it makes the request body so large that the tomcat refuses to serve it and the user receives HTTP 502 from the reverse proxy server.


adrian-evo:
Tested and license info generation works.

PR #348, opened 2018-08-21
fix(rest): Show attachments as embedded resource list

maierthomas:
Shows again sw360:attachments array for releases (also projects and components)


adrian-evo:
Tested and attachments list returned successfully.

Issue #346, opened 2018-08-20
Attachment Usages Seem to throw Exception at Project Page


mcjaeger:
On various instances, when laoding the project details page, I found an exception in the log - effect is currently unknown, just found it, but it looks not so good:

``` 
2018-08-20 08:16:55,392 ERROR AttachmentAwarePortlet:140 - Cannot load attachment usages
 org.apache.thrift.TApplicationException: Internal error processing getAttachmentsUsages
        at org.apache.thrift.TApplicationException.read(TApplicationException.java:111)
        at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:79)
        at org.eclipse.sw360.datahandler.thrift.attachments.AttachmentService$Client.recv_getAttachmentsUsages(AttachmentService.java:728)
        at org.eclipse.sw360.datahandler.thrift.attachments.AttachmentService$Client.getAttachmentsUsages(AttachmentService.java:713)
        at org.eclipse.sw360.portal.portlets.AttachmentAwarePortlet.setAttachmentsInRequest(AttachmentAwarePortlet.java:130)
        at org.eclipse.sw360.portal.portlets.AttachmentAwarePortlet.setAttachmentsInRequest(AttachmentAwarePortlet.java:192)
        at org.eclipse.sw360.portal.portlets.projects.ProjectPortlet.prepareDetailView(ProjectPortlet.java:844)
        at org.eclipse.sw360.portal.portlets.projects.ProjectPortlet.doView(ProjectPortlet.java:738)
        at com.liferay.portal.kernel.portlet.LiferayPortlet.doDispatch(LiferayPortlet.java:235)

```

PR #349, opened 2018-08-21
fix(projects): fix assertion exception when requesting attachment uses with empty ids set

alexbrdn:
fix assertion exception when requesting attachment uses with empty ids set

closes eclipse/sw360#346


maierthomas:
tested successfully on my setup.

No error message appears in log for projects with attachments and without attachments.

Issue #247, opened 2018-04-26
REST API: provide external id and get back document


maierthomas:
<a href="https://github.com/mcjaeger"><img src="https://avatars1.githubusercontent.com/u/3418977?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mcjaeger](https://github.com/mcjaeger)**
_Tuesday Apr 03, 2018 at 13:19 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/804_

----

For integration with other systems, following endpoint required:
* provide external Id as query
* get the document back

use case is to take the maven artifact Id for example and then be able to get the release.

But it could be also relevant for projects or components

PR #353, opened 2018-08-28
feat(rest): Add externalId endpoint (projects) to REST API

maierthomas:
- new REST API endpoint to list projects based on external ids
- added documentation (project page)

closes eclipse/sw360#247

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -453,6 +453,10 @@ public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
         return repository.getCountByProjectId(id);
     }
 
+    public Set<Project> searchByExternalIds(Map<String, Set<String>> externalIds, User user) {
+        return repository.searchByExternalIds(externalIds, user);
+    }
+
     public Set<Project> getAccessibleProjects(User user) {
         return repository.getAccessibleProjects(user);
     }
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
@@ -132,6 +132,15 @@ public class ProjectRepository extends SummaryAwareRepository<Project> {
                     "  }" +
                     "}";
 
+    private static final String BY_EXTERNAL_IDS =
+            "function(doc) {" +
+                    "  if (doc.type == 'project') {" +
+                    "    for (var externalId in doc.externalIds) {" +
+                    "       emit( [externalId, doc.externalIds[externalId]] , doc._id);" +
+                    "    }" +
+                    "  }" +
+                    "}";
+
     public ProjectRepository(DatabaseConnector db) {
         super(Project.class, db, new ProjectSummary());
         initStandardDesignDocument();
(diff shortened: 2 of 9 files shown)


maierthomas:
URL: https://localhost:8443/resource/api/projects/externalIds
Method: POST
Request Body:
```
{
"key1": ["value"]
"key2":["value1","value"]
}
```

alexbrdn:
I object to the endpoint URL. You don't create an object representing external ids with this POST request. Rather, you create a search job by external ids. IMO, the URL should reflect that. Something like https://localhost:8443/resource/api/projects/searchJobs/byExternalIds would be more appropriate, I believe.

hengel2810:
At the moment we are already using simple filtering for some resources, for example if you want to list all components with a specific name:
GET `.../api/components?name=abc`

I don't know whether it's the common way to use the same principle for more complex filters, but maybe we can use something like this:
GET `.../api/components?externalIds.key1=value&externalIds.key2=value1,value2`

This might be a more RESTful solution.

maierthomas:
Yes you are right @hengel2810.
Maybe we can use `javax.ws.rs.core.MultivaluedMap`
http://..../?key1=value1&key1=value2&key2=value3
-> receive: key1 values value1, value2; key2 value3.

But I want to prevent to search for all components and loop the external ids.
I think this will lead to a performance issue.
In this case we should call the couchdb view with the specific external ids

hengel2810:
I think, and you've already mentioned that, the `MultivaluedMap` makes it unpossible to determine if an entry is a filter for externalIds. This could be problematic with future filter implementations.

I'm not sure if the `externalIds.key1=value&externalIds.key2=value1,value2` solution is a common way to handle this, but that's the best idea that comes to my mind at the moment. 

An advantage of this would be that it's easy to convert into a `Map<String, Set<String>>`, so you can use it with your current implementation with the defined CouchDB View.

adrian-evo:
Tested and with current implementation it works, projects with specified ext ids returned successfully.

maierthomas:
I think the best solution would be a `GET` endpoint like
`https://localhost:8443/resource/api/projects/externalIds` or
`https://localhost:8443/resource/api/projects/byExternalIds` or
`https://localhost:8443/resource/api/projects/searchByExternalIds` ...

The endpoint supports a request parameter `List<String>`
This request parameter represents the key value structure of the external id map.
Example:

`https://localhost:8443/resource/api/projects/searchByExternalIds?key1=value1&key2=value2&key3=value3`
Or maybe comma separated:
`https://localhost:8443/resource/api/projects/searchByExternalIds?key1=value1,value2&key2=value3`

maierthomas:
**Update**

Added two endpoints to list projects by external ids: <br>

_GET_ 
URL: /api/projects/searchByExternalIds?key1=value1&key1=value2&key2=value3
Example:

> $ curl 'https://sw360.org/api/projects/searchByExternalIds?project-ext=515432&project-ext=7657&portal-id=13319-XX3' -i -H 'Content-Type: application/hal+json' -H 'Authorization: Bearer eyJ...'

_POST_ 
URL: /api/projects/searchByExternalIds
Example:

> $ curl 'https://sw360.org/api/projects/searchByExternalIds' -i -X POST -H 'Content-Type: application/hal+json' -H 'Authorization: Bearer eyJ...' -d '{
>   "project-ext" : [ "515432", "7657" ],
>   "portal-id" : [ "13319-XX3" ]
> }'
>

innerjoin:
@maierthomas and colleagues: I was just looking for this kind of query capability for SW360 right now, so good to see that you are already into that. I have one use case which I'm not sure you cover yet and one change request:
- Is it possible to query for the existence of an external ID? e.g. `api/projects/searchByExternalIds?project-ext` or `api/projects/searchByExternalIds?project-ext=*` would return all projects that have an external ID called `project-ext`, no matter what value is set for it. This would allow me to get rid of requesting all projects and filtering them on my side. Is this possible with the current implementation and if not, could you enhance the implementation to cover that?
- I would suggest to rename the endpoint from `searchByExternalIds` to `externalIds`. This would allow for future enhancements of the endpoint for scenarios like: _"Update all projects that have the external ID `X` with the value`Y` and change it's value to `Z`_".

Issue #355, opened 2018-08-28
REST API error message in log for moderators


maierthomas:
Error Message in Log:

` 2018-08-28 13:55:12 ERROR RestControllerHelper:192 - cannot create embedded user with email: test@mustermann.de`

Reason:
No "ID" is present for moderation user objects

>             User user = new User();
>             user.setEmail(moderatorEmail);
>             addEmbeddedUser(halResource, user, "sw360:moderators");

PR #356, opened 2018-08-28
fix(rest): Correct REST embedded User to prevent error messages

maierthomas:
Fix `createEmbeddedUser` (moderators) logic to prevent error messages

closes eclipse/sw360#355


maierthomas:
@hengel2810 normally you only can select a existing user as moderator.
But in the case that an user does not exists I want to prevent the exception.

adrian-evo:
Tested and error message is not displayed anymore.

PR #364, opened 2018-09-18
feat(wsimport): lar file

kallesoranko:
Signed-off-by: Kalle Soranko <ksoranko@verifa.io>


maierthomas:
tested successfully, I was able to Import the 'WS Import' as private page.

PR #365, opened 2018-09-18
fix(rest): Trying to sort components by an unsupported property causes NPE

hengel2810:
Currently the following request will throw an NPE:
`GET /components?page=0&sort=ownerCountry`

This PR will fix this bug.


maierthomas:
tested successfully.

unknowing sort properties will not result in a nullpointer exception

PR #366, opened 2018-09-18
fix(rest): GET /{resources} response will contain all resources if there are no paging parameters in the request

hengel2810:
This PR will remove the default paging if there are no paging parameters in the `GET /{resources}` request.

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/resourcelists/PaginationOptions.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/resourcelists/PaginationOptions.java
@@ -14,13 +14,13 @@ package org.eclipse.sw360.datahandler.resourcelists;
 
 import java.util.Comparator;
 
-public class PaginationOptions {
+public class PaginationOptions<T> {
 
     private final int pageNumber;
     private final int pageSize;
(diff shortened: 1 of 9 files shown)


mcjaeger:
(from discussion) seems not to be limited to components (opposed to what the title suggests)

hengel2810:
Sorry about the misleading title. This PR removes the default paging for all resource lists generated by the `ResourceListController`. At the moment it only concerns `Components` but in future implementations it will also concern other resources. I've updated the PR.

maierthomas:
test result:

1. 'GET' /resource/api/components = all components are visible, without pagination parameters
2. 'GET' /resource/api/components?page=1&page_entries=5&sort=name,desc = all components are visible, with pagination parameters.

I think something is wrong, because in the 2. example I receive all components while I set the pagination parameters (page=1 and page_entries=5).

See my test result:

image

hengel2810:
I've updated the PR, the problem should be fixed now.

maierthomas:
@hengel2810 please sqash these 2 commits into 1. thank you.

hengel2810:
@maierthomas commits are updated

maierthomas:
tested successfully.

- Without pagination parameters = All values are visible
- /resource/api/components?page=1&page_entries=5&sort=name,desc = only 5 components are visible

PR #374, opened 2018-09-27
fix(projects): prevent duplicate attachment usages from crashing project display

alexbrdn:
fix(projects): prevent duplicate attachment usages from crashing project display
try to prevent duplicate attachment usages in the first place


adrian-evo:
Tested with duplicated attachments, source code or license info, set or unset usage multiple times. No issues found.

Issue #376, opened 2018-10-02
Reporting: Make separate template document for License Report


mcjaeger:
Currently, the license report (see screenshot) is a modified form of the License Disclosure Document:

![image](https://user-images.githubusercontent.com/3418977/46344918-fc6dde00-c642-11e8-8cdd-e8429a9034e0.png)

It should be a separate template document, copied from the existing template, allowing to have both documents in parallel.

This document should have - as with the current template documents - its premier location (overriding all other locations) in /etc/sw360. Otherwise, it shall be in the resources folder of the license info service so it is part of the *.war file.

mcjaeger:
to be clarfied: contents in the footer.

dreh23:
This can be closed

mcjaeger:
right.

PR #382, opened 2018-10-11
Chore(licenseinfo): Add separate DOCX template for reports

dreh23:
To address different docx templates for license report generation the switch was refactored. Additionally the template was duplicated, but is still a duplicate from the old template.


dreh23:
#376

dreh23:
I am not quite sure why the ip-validation fails. See attached picture below:

eca

mcjaeger:
maybe two things: 1) propagation time and 2) this is not a "git commit -s" ... which facilitates the signed-off-by in the commit message which is missing in your commit message

dreh23:
Forced push with commit -s. IP-validation seems to work now.

maierthomas:
@dreh23 please change the PR title and commit message name.
Suggestion: chore(licenseinfo): Add separate DOCX template for reports

maierthomas:
tested successfully.

Both reports could be generated with that implementation.

PR #390, opened 2018-10-26
fix(license): fix problem in editing licenses

maxhbr:
currently it is not possible to modify and update licenses via the UI, due to
the problem that the `disabled` checked-flag is not send back to the server in
the post request. Changing it to `readonly` fixes the problem.


maierthomas:
tested successfully, I was able to update a license.

PR #405, opened 2018-12-03
chore(ui): Some fixes for UI regarding search and filters

### LINKED ISSUE #384: Search in Project Overview (Table) View Corrections
In the left section when searching for project names, to have list of the according projects. The search term is also used to search inside the description. Actually this is confusing, so the search should be limited to the name when the user enters a string in the name field.

Search terms the shall be  AND conjunction instead of OR conjunction, which is confusing. Currently, searching for "my project" and "1.0" for version, the result includes project with version "1.0", but with different name, and vice versa

## Summary (Updated)

* Search terms entered in project name or component name in the list views shall be AND conjuncted, not OR, also possibly this shall be applied with all other search field to be consistent. If that is not possible and OR semantics is better, the results which have all terms in common shall appear first in the list.

* When a user enters search terms in multiple fields, the multiple fields shall be searched in AND conjunctions and not OR.

* When a user enters a name for project or components, the search should be only in the name field of the records (and not description or other fields) - just using the field where the user has entered the search term).

* If the user searches for a release when adding a release to a project, then the results should be same as with the component search on the component list view. Of course, there will be more, because the releases search lists releases, but the considered components will be the same.

maierthomas:
- add drop down menu for projects view (project state, project type)
- limitation for similar component names in dialog
- some style changes for the sw360 content1 (advanced search...)
- style changes for search view
- removed search.css

Search optimizations:
- remove datatable autosort for components view and search release dialog
- AND concatenation for lucene search text (e.g. Project name or component name)

closes #384 


maierthomas:
Advanced Search modifications
image

Search view modifications
image

frontend/sw360-portlet/src/main/webapp/html/projects/view.jsp
```diff
@@ -139,9 +142,11 @@
             </tr>
             <tr>
                 <td>
-                    <label for="state">State</label>
-                    <input type="text" class="searchbar" name="<portlet:namespace/><%=Project._Fields.STATE%>"
-                           value="<sw360:out value="${state}"/>" id="state" class="filterInput">
+                    <label for="project_state">Project Type</label>
```
greimela-si:
```suggestion
                    <label for="project_state">Project State</label>
```

frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
```diff
@@ -298,7 +295,6 @@
                         {"title": "Component Type", data: "cType"},
                         {"title": "Actions", data: "id", render: {display: renderComponentActions}}
                     ];
-                    order = [[2, 'asc']];
                     columnDefs = [{ "orderable": false, "targets": [0, 4, 5, 6 ,7] }];
```
lepokle: This line should be removed as well, since its declaration was removed and it is not used anymore

maierthomas: was my bad, `columnDefs` should not be removed... only `order`

lepokle: Are you sure? I did not found any other reference to columnDefs.

PR #419, opened 2018-12-17
feat(rest): Allow to search only by externalId-Key (without specific value)

### LINKED ISSUE #361: REST API extension: support more queries by externalIds, rename ressource path
Extracted comment from #353 

@maierthomas and colleagues: I was just looking for this kind of query capability for SW360 right now, so good to see that you are already into that. I have one use case which I'm not sure you cover yet and one change request:
- Is it possible to query for the existence of an external ID? e.g. `api/projects/searchByExternalIds?project-ext` or `api/projects/searchByExternalIds?project-ext=*` would return all projects that have an external ID called `project-ext`, no matter what value is set for it. This would allow me to get rid of requesting all projects and filtering them on my side. Is this possible with the current implementation and if not, could you enhance the implementation to cover that?
- I would suggest to rename the endpoint from `searchByExternalIds` to `externalIds`. This would allow for future enhancements of the endpoint for scenarios like: _"Update all projects that have the external ID `X` with the value`Y` and change it's value to `Z`_".

maierthomas:
- allows to search for externalId keys without values like (.../searchByExternalIds?external-id-key or .../searchByExternalIds?external-id-key=)
- added some tests for missing API docs (create project, create release, externalIds for components...)
- some small bugfixes regarding vulnerability/vendor ids (URL selflink was null)

closes #361 

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/couchdb/DatabaseRepository.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/couchdb/DatabaseRepository.java
@@ -104,6 +104,22 @@ public class DatabaseRepository<T> extends CouchDbRepositorySupport<T> {
         return queryForIds(query);
     }
 
+    public Set<String> queryForIdsOnlyComplexKey(String queryName, String key) {
+        return queryForIdsOnlyComplexKeys(queryName, Collections.singleton(key));
+    }
+
+    public Set<String> queryForIdsOnlyComplexKeys(String queryName, Set<String> keys) {
+        Set<String> queryResult = new HashSet<>();
+        for (String key : keys) {
+            // If there is no value for the key just search for the key occurrence
+            // \ufff0 is used to ignore all other complex keys
+            ComplexKey startKeys = ComplexKey.of(key);
+            ComplexKey endKeys = ComplexKey.of(key, "\ufff0");
+            queryResult.addAll(queryForIds(queryName, startKeys, endKeys));
+        }
+        return queryResult;
+    }
+
     private static Set<ComplexKey> createComplexKeys(Map.Entry<String, Set<String>> key) {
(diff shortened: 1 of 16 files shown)


innerjoin:
@maierthomas Great Christmas gift, thanks a lot 🥇

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/couchdb/DatabaseRepository.java
```diff
@@ -97,11 +97,28 @@ public DatabaseRepository(Class<T> type, DatabaseConnector databaseConnector) {
...
-                .map(DatabaseRepository::createComplexKeys)
-                .flatMap(Collection::stream).collect(Collectors.toSet());
-        ViewQuery query = createQuery(queryName).keys(complexKeys);
-        return queryForIds(query);
+        Set<String> queryResult = new HashSet<>();
+        Set<ComplexKey> complexKeys = new HashSet<>();
+
+        for (Map.Entry<String, Set<String>> externalId : keys.entrySet()) {
+            Set<String> externalValues = externalId.getValue();
+            if (externalValues.isEmpty() || (externalValues.size() == 1 && externalValues.iterator().next().isEmpty())) {
```
maxhbr: But now one can no longer search for all entries with key type `someExternalIdName` and empty value?

I think this implicit behaviour would better be moved to another explicit function.

innerjoin: I agree with the first bit about searching for empty values, this is valuable and should not be _lost_.

However I would suggest an alternative solution approach:
- _**.../searchByExternalIds?external-id-key**_ - search only by key
- _**.../searchByExternalIds?external-id-key=**_ - search by key and value (in this case, an empty value)

mcjaeger:
(from discussion of the telco today) the ampersand is meant to be a delimiter of key value pairs and does not imply an conjunction for queries.

PR #421, opened 2018-12-18
feat(rest) Added CORS module to fix CORS problems with JS clients

blaumeiser-at-bosch:


blaumeiser-at-bosch:
Error pull request, will rethink the strategy

PR #422, opened 2018-12-18
hengel2810/fix cors problem with client

blaumeiser-at-bosch:


mcjaeger:
(from discussion) no easy to test because it is an infrastructure issue at the JSclient (which has not been published so far). merging should be OK if review is OK. By default it is disabled

rest/rest-common/src/main/java/org/eclipse/sw360/rest/common/Sw360CORSFilter.java
```diff
@@ -0,0 +1,93 @@
...
+        httpMethods.add(HttpMethod.DELETE.name());
+        httpMethods.add(HttpMethod.PATCH.name());
+        return valueListStringBuilder(httpMethods);
+    }
+
+    private static String allowedHttpHeaders() {
+        List<String> httpHeaders = new ArrayList<>();
+        httpHeaders.add(HttpHeaders.CONTENT_TYPE);
+        httpHeaders.add(HttpHeaders.AUTHORIZATION);
+        return valueListStringBuilder(httpHeaders);
```
maierthomas: You only want to have a comma separated string? you can use Java8- String joins
`return String.join(",", httpHeaders);`

maierthomas:
please dont kill me, the colon is missing in the commit title :innocent:

PR #423, opened 2018-12-18
feat(rest): Updated response for GET requests on resource lists if there are no resources available

blaumeiser-at-bosch:


mcjaeger:
(from discussion) test when empty response is expected

PR #425, opened 2018-12-20
feat(licenseinfo): Use property for controlling license info generation

lepokle:
This PR adds a property which controls whether information from external license files should be taken on reading SPDX files.


backend/src-common/src/main/resources/sw360.properties
```diff
@@ -9,6 +9,8 @@
 # common property file for the backend services
 backend.url= http://localhost:8080
 
+licenseinfo.spdxparser.use-license-info-from-files=true
```
maierthomas: maybe move this property to `backend/src/src-licenseinfo/src/main/resources`

PR #430, opened 2019-01-09
Fix for Bug Null Ptr Exec at Generate License Report

dreh23:
Fix for https://github.com/siemens/sw360/issues/12

Signed-off-by: Johannes Amorosa <johannes.amorosa@endocode.com>


backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
```diff
@@ -40,7 +41,7 @@
 import static org.eclipse.sw360.licenseinfo.outputGenerators.DocxUtils.*;
 
 public class DocxGenerator extends OutputGenerator<byte[]> {
-
+    private static final Logger LOGGER = Logger.getLogger(DocxGenerator.class);
```
maierthomas: `LOGGER` is not used, can be removed

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
```diff
@@ -293,13 +294,16 @@ private void fillOverview3rdPartyComponentTable(XWPFDocument document, Collectio
...
             row.addNewTableCell().setText(licenseInfo.getComponentName());
 
-            String globalLicense = "";
-            for(LicenseNameWithText l : licenseInfo.getLicenseNamesWithTexts()) {
-                if(l != null && l.getType().equals("global")) {
-                    globalLicense = l.getLicenseName();
-                    break;
-                }
-            }
+            Optional<String> opGlobalLicense = licenseInfo.getLicenseNamesWithTexts().stream()
```
maierthomas: What about `if(l != null && "global".equals(l.getType()) {`
The root cause is just the NPE of `.getType`.
I think your implementation is working quite good but somehow not really readable.

maierthomas:
please change commit msg: fix(licenseinfo): NPE at....

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
dreh23: A null pointer check is not very elegant. But of course your suggestion is more readable.

PR #431, opened 2019-01-09
chore(thrift): update thrift version to 0.11.0

maxhbr:
It would also possible to update to thrift 0.12.0, but the corresponding libthrift is not yet available on maven central.

This also implies that one needs to update the thrift compiler on the build host to version 0.11.0, this can be done via the script `scripts/install-thrift.sh` (e.g. on ubuntu).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,12 @@ sudo: required
 dist: trusty
 language: java
 jdk: openjdk8
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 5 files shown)


maierthomas:
successfully tested on my local vagrant setup with upgraded Apache Thrift version 0.11.0

maierthomas:
The code review is partly done but I am not really familiar with the docker file.
code looks good.

Issue #434, opened 2019-01-10
REST: Create project needs linked releases and visibility is not editable


maierthomas:
For creating a project over REST API its necessary to add linked releases or at least a empty linked releases list. Its also not possible to set the visibility of a project.

PR #435, opened 2019-01-10
chore(vulnerabilities): Linked releases can be empty or null (rest create project)

maierthomas:
- fix bug if user creates a project without linked releases over rest
- added visibility to rest interface (project mixin)

closes eclipse/sw360#434


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
```diff
@@ -175,6 +176,10 @@ public Sw360Module() {
             @JsonProperty("linkedReleases")
             abstract public Map<String, ProjectReleaseRelationship> getReleaseIdToUsage();
 
+            @Override
+            @JsonProperty("visibility")
+            abstract public Visibility getVisbility();
```
kallesoranko: typo getVisbility

maierthomas: the typo is correct here, because the db field is named `visbility`......
So this code will override the db field `visbility` to `visiblity` by using the getter `getVisbility`.

PR #439, opened 2019-01-16
chores(quick-deploy): add quick deploy for portlet

lepokle:
This commit is an addition to the quick deploy PR sw360/sw360-chores#55

In order to prevent the Tomcat from creating a temp directory for the webapp
from which the webapp is service, the following context settings are
necessary:
  - antiJARLocking=false
  - antiResourceLocking=false

If the webapp provides no context file, liferay generates such a file with
both settings set to true.

This commit introduces an own context file which is filled based on properties.
These propterties are "true" by default but can be set to "false" by using
the maven profile "develop" on building.

### DIFF
--- a/backend/svc/pom.xml
+++ b/backend/svc/pom.xml
@@ -79,6 +79,7 @@
             <plugin>
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-war-plugin</artifactId>
+              <version>2.2</version>
               <configuration>
                 <outputDirectory>${webapps.dir}</outputDirectory>
               </configuration>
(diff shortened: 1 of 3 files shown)


maxhbr:
The file `frontend/sw360-portlet/src/main/webapp/META-INF/context.xml` needs an license header

mcjaeger:
> The file `frontend/sw360-portlet/src/main/webapp/META-INF/context.xml` needs a license header

```
<!--
  ~ Copyright Siemens AG, 2013-2017. Part of the SW360 Portal Project.
  ~
  ~ SPDX-License-Identifier: EPL-1.0
  ~
  ~ All rights reserved. This program and the accompanying materials
  ~ are made available under the terms of the Eclipse Public License v1.0
  ~ which accompanies this distribution, and is available at
  ~ http://www.eclipse.org/legal/epl-v10.html
  -->
```

PR #446, opened 2019-01-22
feat(rest): Add filter functionality for GET /components route

blaumeiser-at-bosch:
feat(datahandler): Added filter functionality for resources

Signed-off-by: Henrik Engelbrink <henrik.engelbrink@bosch-si.com>

PR #447, opened 2019-01-22
Paging and sorting for Releases

blaumeiser-at-bosch:
This PR adds sorting and paging functionality for releases, which was already present for components

### how to test:
try to access releases via rest and try to sort and page them (in the same way as it is done for components)

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -16,7 +16,6 @@ import org.apache.log4j.Logger;
 import org.apache.thrift.TBase;
 import org.apache.thrift.TException;
 import org.apache.thrift.TFieldIdEnum;
-import org.eclipse.sw360.datahandler.resourcelists.*;
 import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
 import org.eclipse.sw360.datahandler.thrift.components.Component;
 import org.eclipse.sw360.datahandler.thrift.components.Release;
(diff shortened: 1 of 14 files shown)


mcjaeger:
@blaumeiser-at-bosch I know that you are just transferring the branches of Mr. Engel, but it is really helpful for others to have
* description
* how to test

imaykay:
I am going to do the code review.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ReleaseSpecTest.java
```diff
@@ -157,38 +159,70 @@ public void should_document_get_releases() throws Exception {
...
+                .param("page","0")
+                .param("page_entries", "5")
+                .param("sort","name,desc")
                 .accept(MediaTypes.HAL_JSON))
                 .andExpect(status().isOk())
                 .andDo(this.documentationHandler.document(
+                        requestParameters(
+                                parameterWithName("page").description("Page of releases"),
+                                parameterWithName("page_entries").description("Amount of releases per page"),
+                                parameterWithName("sort").description("Defines order of the releases")
```
imaykay: Maybe it would be good to at least link a documentation on how to properly influence the sorting. So what are the possible values of this param.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -176,7 +176,7 @@ private String getAPIBaseUrl() throws URISyntaxException {
         return new PaginationOptions<T>(pageable.getPageNumber(), pageable.getPageSize(), comparator);
     }
 
-    private Comparator<T> comparatorFromPageable(Pageable pageable,  String resourceClassName) throws ResourceClassNotFoundException {
+    private Comparator<T> comparatorFromPageable(Pageable pageable, String resourceClassName) throws ResourceClassNotFoundException {
         Sort.Order order = firstOrderFromPageable(pageable);
```
imaykay: I am sure if my understood is correct, but would this line imply that sorting only works for one attribute?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/release/ReleaseController.java
```diff
@@ -55,10 +61,10 @@
 import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
```
imaykay: Unused imports

maxhbr:
There is the test 
https://github.com/eclipse/sw360/blob/00ffcf8e9dbe0bf22a8f7cb74cf0d98f12a9b678/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/ComponentTest.java#L96-L108
which asserts, that if one requests the list of components, then the returned json still needs to have the part
```
{
  [...],
  "_embedded": { "sw360:components": [] },
  "_links": {
    "curries": [{ 
        "href": "[...]docs/components.html",
        "name": "sw360",
       "templated": true
    }]
  }
}
```

##
The corresponding assertions are defined in:
https://github.com/eclipse/sw360/blob/00ffcf8e9dbe0bf22a8f7cb74cf0d98f12a9b678/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/TestHelper.java#L41-L68

maxhbr:
I am not able to generate empty responses, e.g. see https://github.com/eclipse/sw360/pull/447/files#diff-66ac9c1a1af383185e05150c943c5effR77

This causes two tests to fail

maxhbr:
This PR will have conflicts with #446

maxhbr:
Related links:
- https://stackoverflow.com/questions/30286795/how-to-force-spring-hateoas-resources-to-render-an-empty-embedded-array
- https://github.com/spring-projects/spring-hateoas/issues/522

mcjaeger:
would it be Ok to close?

mcjaeger:
as discussed, let's close it.

PR #448, opened 2019-01-22
chore(thrift): build thrift with `--without-java` flag

maxhbr:
This minor change. We do not need to build with the `--with-java` flag, since we get the java libraries from maven. We only need the core compiler.
If this java is not disabled, the thrift build step tries to download a jar dependency while compiling.

##

This can be tested by running 
```
$ mvn clean
$ scripts/compileWithDocker.sh
```
and is also used by travis, so travis also tests the changes here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ cache:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -y couchdb
-  - sudo ./scripts/install-thrift.sh --no-cleanup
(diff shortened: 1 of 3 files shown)


maxhbr:
There are problems with one travis job due to this change here:

maxhbr:
If I push, the travis tests are not actually rerun. If one clicks on the `Details` link one lands on the old execution for commit 660825e:

and the Build History does not contain any new entries

PR #450, opened 2019-01-22
chore(common): Mail service sends notifications asynchronous

### LINKED ISSUE #449: MailUtils: E-Mail notifications are send synchronous
The E-Mail notification send is synchronous. But it should be asynchronous.

maierthomas:
- MailUtil: `transport.send` is now called by an Executor
- Singleton (or should be ThreadPoolExecutor with a specific number of controlled threads?)

closes #449


backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
```diff
@@ -113,6 +118,19 @@ public void sendMail(Set<String> recipients, String excludedRecipient, String su
         sendMail(recipients, excludedRecipient, subjectNameInPropertiesFile, textNameInPropertiesFile, notificationClass, roleName, true, textParameters);
     }
 
+    public void sendMailAsync(MimeMessage message) {
+        ExecutorService mailExecutor = Executors.newSingleThreadExecutor();
```
maxhbr: Instead of creating a new Executor every time it might be better to have static or persistent one.

backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
```diff
@@ -113,6 +118,19 @@ public void sendMail(Set<String> recipients, String excludedRecipient, String su
...
+        mailExecutor.execute(() -> {
+            try {
+                Transport.send(message);
+                log.info("Sent message successfully to user " + message.getRecipients(Message.RecipientType.TO));
```
maxhbr: Since this is run asynchronously it might be good to have a more descriptive log message here and in the catch clause

backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
```diff
@@ -113,6 +118,19 @@ public void sendMail(Set<String> recipients, String excludedRecipient, String su
...
+            } catch (MessagingException e) {
+                log.error("Could not sent E-Mail notification via SMTP", e);
```
maxhbr: Maybe also log the reason, why the sending failed?

mcjaeger: but it would be good to have the e-mail and/or the STMP server address in the log

backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
```diff
@@ -191,10 +207,10 @@ private void sendMailWithSubjectAndText(String recipient, MimeMessage message) {
     private void writeMessageToLog(MimeMessage message) {
         try {
             log.info(String.format("E-Mail message dumped to log, because mailing is not configured [correctly]:\n"+
-            "From: %s\n"+
-            "To: %s\n"+
-            "Subject: %s\n"+
-            "Text: %s\n",
+                            "From: %s\n"+
```
mcjaeger: one good application of this would be to have the mails sent instead manually - so an admin could pull it out of the log. But in this case I am not sure to have the entire message printed.

backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
maierthomas: changed to static

backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
maierthomas: added host

backend/src-common/src/main/java/org/eclipse/sw360/mail/MailUtil.java
```diff
@@ -35,6 +40,7 @@
 public class MailUtil {
 
     private static final Logger log = Logger.getLogger(MailUtil.class);
+    private static final ExecutorService mailExecutor = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
maierthomas: So we should shutdown `ExecutorService` after every mail notification?
Does it make any sense to declare this object as static if we close/shutdown it every time?

maierthomas:
tested successfully on our stage/test environment

PR #428, opened 2019-01-07
refactor(webjars): update versions of webjars

lepokle:
Closes #412

Signed-off-by: Leo von Klenze <leo.vonklenze@scansation.de>

PR #437, opened 2019-01-11
refactor(velocity): update to new version

lepokle:
* the new version of tools does not contain dom4j as dependency anymore. As we use it in the tests it is now directly declared with scope test
* the template for text license info files was modified to ensure that every release is on its own line
* deprecated and removed jquery api was fixed (error replaced by fail)

Closes #128

Signed-off-by: Leo von Klenze <leo.vonklenze@scansation.de>

PR #453, opened 2019-01-25
fix(projects): fix (de)select all checkbox and expanding license files on Generate License Information page

alexbrdn:
Regression due to jquery version bump of #428. Deprecated method `error()` is no longer available on jqXHR object.


alexbrdn:
Turns out, #437 has already fixed this between the time I have pulled the master and pushed the fix.

PR #454, opened 2019-01-25
fix(treetables): fix inconsistent indentation in treetables

alexbrdn:
`<sw360:out>` tag adds a `<span>` around the text when it is necessary to abbreviate it. Padding such `span`s inside treetables leads to inconsistent and misleading indentation.

Removing the padding seems to have no drawbacks.


maierthomas:
successfully tested on siemens stage/test environment

PR #460, opened 2019-02-01
fix(vul-scheduler): fixed an issue where vulnerabilities were stored in the wrong db

imaykay:
* fixed the storage db of the cve search which stored vuls and their relations in the standard db while the thrift service tried to read them from the vm db

fixes eclipse/sw360/#401

Testing: On a clean installation one should be able to schedule the cve search (Admin -> Schedule) and after a successful run there should be vulnerabilities in the application that are visible at the releases or on the Vulnerabilities page.
The bad thing: I guess the start time for the scheduler is only changeable via code. So maybe one want to change org.eclipse.sw360.schedule.timer.Scheduler.getNextSyncDate(int, int) before deploying the branch (careful: the time entered is GMT).

### DIFF
--- a/backend/src/src-vulnerabilities/src/main/java/org/eclipse/sw360/vulnerabilities/db/VulnerabilityDatabaseHandler.java
+++ b/backend/src/src-vulnerabilities/src/main/java/org/eclipse/sw360/vulnerabilities/db/VulnerabilityDatabaseHandler.java
@@ -11,20 +11,21 @@
 package org.eclipse.sw360.vulnerabilities.db;
 
 import com.google.common.collect.ImmutableList;
-import org.apache.commons.lang.StringUtils;
-import org.apache.log4j.Logger;
-import org.apache.thrift.TBase;
+
+import org.eclipse.sw360.datahandler.common.DatabaseSettings;
 import org.eclipse.sw360.datahandler.couchdb.DatabaseConnector;
 import org.eclipse.sw360.datahandler.thrift.RequestStatus;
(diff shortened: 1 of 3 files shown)


backend/src/src-vulnerabilities/src/main/java/org/eclipse/sw360/vulnerabilities/db/VulnerabilityDatabaseHandler.java
```diff
@@ -35,22 +36,13 @@
...
-        relationRepo = new VulnerabilityRelationRepository(db);
-    }
+    public VulnerabilityDatabaseHandler() throws MalformedURLException {
+        DatabaseConnector db = new DatabaseConnector(DatabaseSettings.getConfiguredHttpClient(),
+                DatabaseSettings.COUCH_DB_VM);
 
-    public VulnerabilityDatabaseHandler(DatabaseConnector db) throws MalformedURLException {
-        this.db = db;
         vulRepo = new VulnerabilityRepository(db);
         relationRepo = new VulnerabilityRelationRepository(db);
```
maxhbr: Does that now mean, that the `VulnerabilityRealtions` are written to the `COUCH_DB_VM`? They should be written to the default database.

imaykay: Exactly. I documented the reasons in the bug ticket. That is a bit of a problem when having issue tickets and PRs :-)

PR #461, opened 2019-02-01
fix(Attachments-UI):Restrict from duplicate attachment upload

### LINKED ISSUE #417: There should not be duplicate attachment names
Currently, a user could make an upload of attachments with the same name multiple times.

The system should / could enforce different names

nutanv1contr:
#### Summary:
Restrict users from uploading multiple attachments with same file name.
#### Changes:
If user tries to upload multiple attachments with same file name , then during upload it will restrict the user from uploading  with warning message "Attachment with <i> filename</i> already exist"
#### Steps to test:
1.Try to upload an attachment with same file name and extension , listed on the attachment list.
#### Expected result: 
It should give an warning message "Attachment with  <i> filename</i> already exist"

closes #417

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -400,9 +400,10 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
         // Get actual document for members that should not change
         Component actual = componentRepository.get(component.getId());
         assertNotNull(actual, "Could not find component to update!");
-
         if (changeWouldResultInDuplicate(actual, component)) {
             return RequestStatus.DUPLICATE;
+        } else if (duplicateAttachmentExist(component)) {
+            return RequestStatus.DUPLICATE_ATTACHMENT;
         } else if (makePermission(actual, user).isActionAllowed(RequestedAction.WRITE)) {
             // Nested releases and attachments should not be updated by this method
             if (actual.isSetReleaseIds()) {
@@ -428,6 +429,14 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
         return isDuplicate(after);
     }
+
+    private boolean duplicateAttachmentExist(Component component) {
+    	if(component.attachments != null && !component.attachments.isEmpty()) {
+            return AttachmentConnector.isDuplicateAttachment(component.attachments);
+        }
+        return false;
+    }
+
 
     private void updateComponentInternal(Component updated, Component current, User user) {
         // Update the database with the component
@@ -621,6 +630,8 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
         if (actual.equals(release)) {
             return RequestStatus.SUCCESS;
+        } else if (duplicateAttachmentExist(release)) {
+            return RequestStatus.DUPLICATE_ATTACHMENT;
         } else if (changeWouldResultInDuplicate(actual, release)) {
             return RequestStatus.DUPLICATE;
         } else {
@@ -669,6 +680,13 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
         }
 
         return isDuplicate(after);
+       }
+
+    private boolean duplicateAttachmentExist(Release release) {
+        if (release.attachments != null && !release.attachments.isEmpty()) {
+            return AttachmentConnector.isDuplicateAttachment(release.attachments);
+        }
+        return false;
     }
 
     private void deleteAttachmentUsagesOfUnlinkedReleases(Release updated, Release actual) throws SW360Exception {
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -194,7 +194,9 @@ public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
         if (changeWouldResultInDuplicate(actual, project)) {
             return RequestStatus.DUPLICATE;
-        } else if (!changePassesSanityCheck(project, actual)) {
+        } else if (duplicateAttachmentExist(project)) {
+            return RequestStatus.DUPLICATE_ATTACHMENT;
+        } else if (!changePassesSanityCheck(project, actual)){
             return RequestStatus.FAILED_SANITY_CHECK;
         } else if (makePermission(actual, user).isActionAllowed(RequestedAction.WRITE)) {
             copyImmutableFields(project,actual);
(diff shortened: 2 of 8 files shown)


imaykay:
I am not sure, if the basic idea on how to solve this issue is the best. I have two issues with this approach:
1. I am able to upload two or more attachments with the same name in one edit cycle because the javascript only compares new names with already stored names (all attachments added in this cycle are not yet stored and linked to the object so they are not part of the comparison).
2. If we really want to achieve a better data situation in the database, the check should definitely happen on the server side. Best approach would be in the thrift services so that other clients like REST would also get this restriction.
For a fail-fast feedback, the javascript is still a good thing, but I think it is not sufficient for solving the problem.

nutanv1contr:
I have committed the latest changes. @imaykay please review.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -609,42 +619,53 @@ public RequestStatus updateRelease(Release release, User user, Iterable<Release.
...
         return RequestStatus.SUCCESS;
+       }
     }
 
-    private void deleteAttachmentUsagesOfUnlinkedReleases(Release updated, Release actual) throws SW360Exception {
+    private boolean duplicateAttachmentExist(Release release) {
+    	if(release.attachments != null && !release.attachments.isEmpty()) {
+            return RepositoryUtils.isDuplicateAttachment(release.attachments);
+        }
+		return false;
```
imaykay: Please check spaces vs. tabs.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
```diff
@@ -208,7 +210,14 @@ public RequestStatus updateProject(Project project, User user) throws SW360Excep
         }
     }
 
-    private void deleteAttachmentUsagesOfUnlinkedReleases(Project updated, Project actual) throws SW360Exception {
+    private boolean duplicateAttachmentExist(Project project) {
+        if(project.attachments != null && !project.attachments.isEmpty()) {
+            return RepositoryUtils.isDuplicateAttachment(project.attachments);
+        }
+        return false;
+	}
```
imaykay: Please check spaces vs. tabs.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/Sw360Portlet.java
```diff
@@ -274,6 +274,9 @@ public void setSessionMessage(PortletRequest request, RequestStatus requestStatu
             case FAILED_SANITY_CHECK:
                 setSW360SessionError(request, ErrorMessages.UPDATE_FAILED_SANITY_CHECK);
                 break;
+            case DUPLICATE_ATTACHMENT:
+            	setSW360SessionError(request, ErrorMessages.DUPLICATE_ATTACHMENT);
```
imaykay: Please check spaces vs. tabs.

frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
```diff
@@ -168,8 +168,9 @@
         // register event handlers
         $('#upload-attachment-button').on('click', function() {
             attachmentUploader.showUploadDialog(function(addedAttachment) {
-                attachmentTable.row.add(addedAttachment).draw(false);
-                $(attachmentTable.row('#' + addedAttachment.attachmentContentId).node()).addClass('added');
+            	attachmentTable.row.add(addedAttachment).draw(false);
+            	$(attachmentTable.row('#' + addedAttachment.attachmentContentId).node()).addClass('added');
+
```
imaykay: Maybe you could reset this file as nothing valuable got changed?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/ErrorMessages.java
```diff
@@ -23,6 +23,7 @@
     public static final String COMPONENT_DUPLICATE ="Component could not be added, since a component with the same name already exists.";
     public static final String RELEASE_NOT_ADDED = "Release could not be added.";
     public static final String RELEASE_DUPLICATE ="Release could not be added, since a release with the same name and version already exists.";
+    public static final String DUPLICATE_ATTACHMENT ="Multiple Attachments with same name cannot be present in attachment list.";
```
imaykay: I am a fan of proper formatting as you might have noticed :-) Maybe add a space after the equal sign (I know, the others did not do it consistently, but we can do it better :D ) and write "attachments" with a small A.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/RepositoryUtils.java
```diff
@@ -59,4 +61,8 @@ public static RequestSummary doBulk(Collection<?> objects, User user, DatabaseRe
 
         return searchIds;
     }
+
+    public static boolean isDuplicateAttachment(Set<Attachment> attachments) {
```
imaykay: Maybe place this method into the AttachmentConnector, which seems to be used for attachment related methods? As this method has no functionality that can be used for different data types, I am not sure if this universal class is a good place.

mcjaeger:
should be in for `sw360-3.4.0-M1`

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
nutanv1contr: Thanks @imaykay  for reviewing the changes.  I have added logic to check the sha1 aloong with file name check. 
As discussed about the second point we will restrict users from updating the object if there is a duplicate attachment present on the list and application will display an error message which will indicate the reason for  update failure. So I am not making any change for this as the logic is already added.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
nutanv1contr: Done.

frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
nutanv1contr: Done

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/ErrorMessages.java
nutanv1contr: Done also updated the error message as we will be checking file name along with sha1.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/RepositoryUtils.java
nutanv1contr: Done.Thanks for mentioning that.

imaykay:
I couldn't test the final changes, just looked at the code, because my data already has the defaultvendor field which causes issues in older branches :-/ @mcjaeger could you do a quick retest?

mcjaeger:
@imaykay yes

nutanv1contr:
Thanks @mcjaeger for your reviewing the changes. I have made necessary changes. Now application will stay in edit mode along with the error message.

mcjaeger:
for my review, I think it would have been better to stay at the attachments tab and not switch to the general details tab in case of the error message. It is apparent at project and release attachment, for component attachments, it is OK.

mcjaeger:
sorry, was mistakenly putting my review e-mail into the commit message, was the one of Andi of course ...

Issue #426, opened 2019-01-04
Visibility of attachment usage without any WRITE permission


maierthomas:
Currently all users can try to change the attachment usage for projects in the view mode.
Therefore users can check/uncheck attachment usages and the everyone can see the "save attachment usage" button. In case that users have no admin privileges or in general no write access an error occurs as liferay banner after saving.

For users without write access the button should not be visible and all checkbox should be disabled.

smrutis1:
I would like to work on it.

PR #462, opened 2019-02-04
feat(attachment-usage): Restrict modification of the attachment usage

smrutis1:
**Summary:**
Restrict users, without any WRITE access from modifying the attachment usage.

**Changes:**
"Save Usages" button is hidden and all the checkboxes are disabled for the users, without any WRITE access.

Closes: #426 

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
@@ -865,7 +865,7 @@ public class ProjectPortlet extends FossologyAwarePortlet {
                 putVulnerabilitiesInRequest(request, id, user);
                 putAttachmentUsagesInRequest(request, id);
                 request.setAttribute(
-                        VULNERABILITY_RATINGS_EDITABLE,
+                        WRITE_ACCESS_USER,
                         PermissionUtils.makePermission(project, user).isActionAllowed(RequestedAction.WRITE));
 
                 addProjectBreadcrumb(request, response, project);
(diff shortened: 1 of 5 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
```diff
@@ -865,6 +865,9 @@ private void prepareDetailView(RenderRequest request, RenderResponse response) t
                         VULNERABILITY_RATINGS_EDITABLE,
                         PermissionUtils.makePermission(project, user).isActionAllowed(RequestedAction.WRITE));
 
+                request.setAttribute(
+                        ATTCHMENT_USAGE_EDITABLE,
+                        PermissionUtils.makePermission(project, user).isActionAllowed(RequestedAction.ATTACHMENTS));
```
imaykay: The update action that is used when hitting the "Save Usages" button is org.eclipse.sw360.portal.portlets.projects.ProjectPortlet.saveAttachmentUsages(ResourceRequest, ResourceResponse) which checks only the WRITE permission. Do we really want to check the ATTACHMENTS permission here for client side en-/disabling?

smrutis1: Currently, ATTACHMENT and WRITE serves the same purpose for project level permission. But as suggested, I will rename "VULNERABILITY_RATINGS_EDITABLE" to a generic one "WRITE_ACCESS_USERS" and make the corresponding changes where ever applicable and use the same variable for attachment usages as well.

imaykay:
Thanks for the adjustments, code looks good to me.

PR #466, opened 2019-02-06
feat(wsimport): import sha1 values for releases as external id

kallesoranko:
When using wsimport create external id with key "sha1" and copy there library's sha1 value from Whitesource as value


kallesoranko:
closing this PR for now, since the actual use case is not so clear and sha1 is not an unique identifier

Issue #410, opened 2018-12-13
Self links should be forbidden at all


mcjaeger:
generally, the self links in the UI are almost extinct, however, in the "linked releases hierarchy" view in the project details, you can open endlessly the hierarchy. Also the same for the releases side:

![image](https://user-images.githubusercontent.com/3418977/49930269-84365b00-fec4-11e8-93a7-f4ebbd81f6b6.png)

Maybe the self-link should be forbidden at creation of the reference.

mcjaeger:
I think it would be good to stop users from creating self link for this first.

nutan1st:
I would like to look into this issue

mcjaeger:
actually, it seems to be there for linked projects too. https://github.com/eclipse/sw360/pull/469

nutanv1contr:
Now self link link is forbidden for both release and project.However cyclic link still exists for project.I have created a separate issue for the #546

mcjaeger:
second part was here: https://github.com/eclipse/sw360/pull/709

PR #469, opened 2019-02-11
fix(ReleaseLink): Remove self link from LinkedReleases hierarchy

nutanv1contr:
#### Summary: 
Self links should be forbidden at all
#### Changes: 
Added logic to remove the release from the LinkedReleases list ,if the parent and child release Id is same.
#### Steps to Test:
1. From the "Components" tab select any of the listed component.
2. Goto Release Overview and click on  Release Revision of any of the listed release.
3. Goto to Linked Releases tab of the selected Release and try to expand the Linked Releases Hierarchy and observe the structure. 
#### Expected result:
Self link should  not be present under same Release. 
Closes: #410

### DIFF
--- a/backend/src/src-components/src/main/java/org/eclipse/sw360/components/ComponentHandler.java
+++ b/backend/src/src-components/src/main/java/org/eclipse/sw360/components/ComponentHandler.java
@@ -273,8 +273,14 @@ public class ComponentHandler implements ComponentService.Iface {
         assertNotNull(release);
         assertId(release.getId());
         assertUser(user);
-
+        removeSelfLink(release);
         return handler.updateRelease(release, user, ThriftUtils.IMMUTABLE_OF_RELEASE);
+    }
+
+    private void removeSelfLink(Release release) {
+        if(release.releaseIdToRelationship != null && !release.releaseIdToRelationship.isEmpty()) {
+            release.releaseIdToRelationship.remove(release.id);
+        }
     }
 
     @Override


mcjaeger:
@imaykay indeed, I think I wrote "I think it would be good to stop users from creating self link for this first."

mcjaeger:
@imaykay indeed, I think I wrote "I think it would be good to stop users from creating self link for this first."

I think solving this on portlet level would bear the risk for me that for example, spreadsheet export and REST API will show still faulty data as well. therefore forbidding self links at creation seems to be the better strategy here.

nutanv1contr:
As discussed, I have added the logic at back end to stop creation of release self link.Now during release link creation if user has selected the self link then during save operation self link will be ignored.

backend/src/src-components/src/main/java/org/eclipse/sw360/components/ComponentHandler.java
```diff
@@ -273,10 +273,16 @@ public RequestStatus updateRelease(Release release, User user) throws TException
         assertNotNull(release);
         assertId(release.getId());
         assertUser(user);
-
+        removeSelftLink(release);
         return handler.updateRelease(release, user, ThriftUtils.IMMUTABLE_OF_RELEASE);
     }
 
+    private void removeSelftLink(Release release) {
```
imaykay: Could you please fix the typo real quick?

nutanv1contr: Removed the typo. Sorry for that. 
1. I do not see any ways to add self link during creation as the release is not created yet.
2.As I tested it, if I try to update the release from moderation it redirects me to the same update release flow so the changes are reflecting for me.
3.I will have to check on  fossology releases.
As suggested if I add my code inside updateRelease() then the unit test case will fail due to the immutability check.

imaykay: 1. Oh, of course *facepalm* - my bad :D 
2. Sounds good!
3. That would be a question for @mcjaeger what would be correct. If this self link should be excluded from immutability or what's correct here. Then we could just fix the unit test.

mcjaeger:
@imaykay, @nutanv1contr  approved or not?

imaykay:
I personally still would put the changes one layer deeper as this class is in my opinions more like a proxy that is only validating null constraints. But I will not block this issue longer.

@mcjaeger pls be aware that we were not sure about the fossology code path - do you have any input on that one regarding this issue (see also other comments)?

Issue #249, opened 2018-04-26
Multiple attachment upload is not supported


maierthomas:
<a href="https://github.com/maierthomas"><img src="https://avatars1.githubusercontent.com/u/29916928?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [maierthomas](https://github.com/maierthomas)**
_Tuesday Apr 17, 2018 at 09:41 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/817_

----

Currently users cant upload multiple attachments at the same time.
This behavior is reproducible for
- drag and drop
- select attachment(s) 

In the 'Select Attachment' view users can select several attachments 
thus multiple upload should be supported.

smrutis1:
As per my observation, currently multiple attachment upload is supported, but due to some concurrency fault, sometimes the upload gets stalled in between. I have fixed this issue by applying synchronization mechanism. Now, we could select multiple files and upload the same.

PR #470, opened 2019-02-13
fix(attachment): Multiple attachment upload stall issue

smrutis1:
**Description:**
Currently multiple attachment upload is supported, but due to some concurrency fault, sometimes the upload gets stalled in between. 

**Changes:**
I have fixed this issue by applying synchronization mechanism. Now, we could select multiple attachments and upload the same.

**Testing Steps:**
1) Select/drag and drop multiple attachments and upload the same.
2) Upload attachments from multiple browsers at the same time(to simulate the multi-threaded environment).

closes: #249 

### DIFF
--- a/libraries/commonIO/src/main/java/org/eclipse/sw360/commonIO/AttachmentFrontendUtils.java
+++ b/libraries/commonIO/src/main/java/org/eclipse/sw360/commonIO/AttachmentFrontendUtils.java
@@ -34,18 +34,17 @@ import java.util.concurrent.TimeUnit;
 public class AttachmentFrontendUtils {
 
     private static final Logger log = Logger.getLogger(AttachmentFrontendUtils.class);
-    protected final AttachmentService.Iface client;
 
     private AttachmentStreamConnector connector;
     // TODO add Config class and DI
     private final Duration downloadTimeout = Duration.durationOf(30, TimeUnit.SECONDS);
 
-    public AttachmentFrontendUtils() {
-        this(new ThriftClients());
-    }
+    protected final ThreadLocal<AttachmentService.Iface> attchmntClient = ThreadLocal.<AttachmentService.Iface>withInitial(
+            () -> {
+                return new ThriftClients().makeAttachmentClient();
+            });
 
-    public AttachmentFrontendUtils(ThriftClients thriftClients) {
-        client = thriftClients.makeAttachmentClient();
+    public AttachmentFrontendUtils() {
     }
 
     public InputStream getStreamToServeAFile(Collection<AttachmentContent> attachments, User user, Object context)
@@ -77,11 +76,11 @@ public class AttachmentFrontendUtils {
     }
 
     public AttachmentContent getAttachmentContent(String id) throws TException {
-        return client.getAttachmentContent(id);
+        return attchmntClient.get().getAttachmentContent(id);
     }
 
     public AttachmentContent makeAttachmentContent(AttachmentContent attachmentContent) throws TException {
-        return client.makeAttachmentContent(attachmentContent);
+        return attchmntClient.get().makeAttachmentContent(attachmentContent);
     }
 
     public Attachment getAttachmentForDisplay(User user, String attachmentContentId) {
@@ -97,7 +96,7 @@ public class AttachmentFrontendUtils {
     public void deleteAttachments(Set<String> attachmentContentIds){
         try {
             for(String id: attachmentContentIds) {
-                client.deleteAttachmentContent(id);
+                attchmntClient.get().deleteAttachmentContent(id);
             }
         } catch (TException e){
             log.error("Could not delete attachments from database.",e);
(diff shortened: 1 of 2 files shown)


mcjaeger:
so this means now that until one upload on the entire sw360 server takes place, all other users attempting an upload at the same time are blocked and need to wait until this is finished?

smrutis1:
I have made the necessary changes to fix this issue. Kindly review it and share your views.

**Suggestions:**
For uploading attachments, currently we are using resumable js library which does not support uploading multiple files at once. There is a suggestion to use Plupload [(ref)](https://www.plupload.com) to achieve the same. Please share your views.

maxhbr:
FYI: plupload is licensed under AGPL

mcjaeger:
@maxhbr thanks for pointing this out -> rules it out

imaykay:
I totally agree, that your adjustments are making the code better, because Thrift clients are not thread safe! But they should be reusable.

So, another improvement would be to store the client in a ThreadLocal so that we only need one per thread and not one for each and every call.
* This is suggested as one possible solution in https://grokbase.com/t/thrift/user/1667d5gt24/how-to-make-a-thrift-client-for-multiple-threads
* Netflix is doing something similar in https://github.com/Netflix/CassJMeter/blob/master/src/main/java/com/netflix/jmeter/connections/thrift/ThriftConnection.java

As far as I saw, our *Utils classes are often newly instantiated when needed, only in AttachmentAwarePortlet not. So reusing would be limited either way and usage of the util classes should not be your problem in this ticket. But it might be a good starting point for a cleanup to have the correct implementation inside the util classes.

What I do not understand yet is the general bug description. It said that one single user cannot upload more then one attachment in one edit cycle. But you said in your last comment that the client is not initiating uploads in parallel. Then there shouldn't be any issues on the server side with Thrift clients because the requests should be handled one after another. Can you help me here so that I understand whether I got the issue and the fix right?
And of course, for different users this is obviously a necessary fix, I am just having trouble to understand the single user problem.

smrutis1:
> So, another improvement would be to store the client in a ThreadLocal so that we only need one per thread and not one for each and every call.

Thanks @imaykay for suggestion. I will store the attachment thrift client in the ThreadLocal for now. Later we can implement for other clients as well.

> What I do not understand yet is the general bug description. It said that one single user cannot upload more then one attachment in one edit cycle. But you said in your last comment that the client is not initiating uploads in parallel. Then there shouldn't be any issues on the server side with Thrift clients because the requests should be handled one after another. Can you help me here so that I understand whether I got the issue and the fix right?

Clarification for single user issue:
1) When a user browses and selects multiple files at a time, for each file selection, a separate call is made which was using the same instance of thrift client.
2) While uploading, each file is divided into multiple chunks and multiple requests are being made for each chunk simultaneously, which eventually uses the same instance of thrift client.

Hope, this will clarify your doubts. Please let me know if any further clarification is needed.

smrutis1:
Hi @imaykay , I have updated the PR with the changes. Kindly review it and let me know.

libraries/commonIO/src/main/java/org/eclipse/sw360/commonIO/AttachmentFrontendUtils.java
```diff
@@ -40,6 +41,11 @@
     // TODO add Config class and DI
     private final Duration downloadTimeout = Duration.durationOf(30, TimeUnit.SECONDS);
 
+    protected final ThreadLocal<AttachmentService.Iface> attchmntClient = ThreadLocal.<AttachmentService.Iface>withInitial(
```
imaykay: I think you can delete the line
`protected final AttachmentService.Iface client;`
and cleanup the constructors because we don't need one with ThriftClients as parameter.

smrutis1:
> One question: was there a special reason for not building the ThreadLocal into the ThriftClients class directly?

I thought we can test this one first in productive systems and then we can make the changes in the ThriftClient class for all the clients.

libraries/commonIO/src/main/java/org/eclipse/sw360/commonIO/AttachmentFrontendUtils.java
```diff
@@ -34,18 +34,17 @@
...
-        this(new ThriftClients());
-    }
+    protected final ThreadLocal<AttachmentService.Iface> attchmntClient = ThreadLocal.<AttachmentService.Iface>withInitial(
+            () -> {
+                return new ThriftClients().makeAttachmentClient();
+            });
 
-    public AttachmentFrontendUtils(ThriftClients thriftClients) {
-        client = thriftClients.makeAttachmentClient();
+    public AttachmentFrontendUtils() {
```
imaykay: This one might not be needed since this default constructor is the only one. But it's okay for now :-)

imaykay:
> > One question: was there a special reason for not building the ThreadLocal into the ThriftClients class directly?
> 
> I thought we can test this one first in productive systems and then we can make the changes in the ThriftClient class for all the clients.

My point was not to change it for all clients. But to test the final setup (ThreadLocal in ThriftClients class) for this AttachmentClient only.
Let's go with your solution now so that we can merge the branch after tests.

mcjaeger:
parallel upload seems to work with small attachments (< 20MB), however not larger. I was wondering if we would have some web server defaults on some of the deployments?

mcjaeger:
retested on local deployment ... seem to work now also with 300MB uploads and doing some other uploads in parallel in two browsers using two accounts ...

PR #471, opened 2019-02-13
fix(scripts): add missing dependency to scripts/install-thrift.sh

maxhbr:
the script was missing the dependency `build-essential` which provides `make`

### DIFF
--- a/scripts/install-thrift.sh
+++ b/scripts/install-thrift.sh
@@ -74,7 +74,7 @@ if [[ ! -f "./compiler/cpp/thrift" ]]; then
 
     if has "apt-get" ; then
         $SUDO_CMD apt-get update
-        $SUDO_CMD apt-get install -y libboost-dev libboost-test-dev libboost-program-options-dev libevent-dev automake libtool flex bison pkg-config g++ libssl-dev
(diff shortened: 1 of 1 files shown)


mcjaeger:
when executing I was a little bit confused by the following output:
```
...
thrift 0.11.0

Building C (GLib) Library .... : no
Building C# (Mono) Library ... : no
Building C++ Library ......... : no
Building D Library ........... : no
Building Dart Library ........ : no
Building dotnetcore Library .. : no
Building Erlang Library ...... : no
Building Go Library .......... : no
Building Haskell Library ..... : no
Building Haxe Library ........ : no
Building Java Library ........ : no
Building Lua Library ......... : no
Building NodeJS Library ...... : no
Building Perl Library ........ : no
Building PHP Library ......... : no
Building Plugin Support ...... : no
Building Python Library ...... : no
Building Py3 Library ......... : yes
Building Ruby Library ........ : no
Building Rust Library ........ : no
...
```

maxhbr:
```
Building Java Library ........ : no
```
This is correct, we do not want to build the java library. We pull that library via maven.

```
Building Py3 Library ......... : yes
```
It might be good to fix this, we do not need these libraries. (**Edit:** this fix is now contained in this PR)

Issue #468, opened 2019-02-11
UR errors / glitches when editing


mcjaeger:
Stuff when working on the component portlet when testing https://github.com/eclipse/sw360/pull/458

1. Components:

Each time user tab out of component name field , application shows a message "To Avoid duplicate components, check there similar ones! Does yours already"

2. Releases:

If user tries to add a new release with any of the existing release name then on click of Add release , application gives an error message "Components is temporarily unavailable." and shows a blank screen.

nutanv1contr:
I would like to work on this issue.

mcjaeger:
(from discussion) let's focus on the second point only.

PR #473, opened 2019-02-19
fix(UI-Release): UI error on duplicate release creation

nutanv1contr:

#### Issue:
Vendor Id null check was missing for new release .As exception handling was not proper so there was a abnormal termination from code flow which was causing the issue. 

#### Steps to test:
1.Create a new release with any of the existing version name.

#### Expected result:
1.It should give a error message for duplicate release and it should display the same new release window.
Closes  : #468

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
@@ -1138,8 +1138,8 @@ public class ComponentPortlet extends FossologyAwarePortlet {
     }
 
     private void fillVendor(Release release) throws TException {
-        VendorService.Iface client = thriftClients.makeVendorClient();
-        if(release.isSetVendorId()) {
+        if(!isNullOrEmpty(release.getVendorId()) && release.isSetVendorId()) {
+            VendorService.Iface client = thriftClients.makeVendorClient();
             Vendor vendor = client.getByID(release.getVendorId());
             release.setVendor(vendor);
         }


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
```diff
@@ -1128,7 +1128,9 @@ public void updateRelease(ActionRequest request, ActionResponse response) throws
     }
 
     private void prepareRequestForReleaseEditAfterDuplicateError(ActionRequest request, Release release) throws TException {
-        fillVendor(release);
+        if(!isNullOrEmpty(release.getVendorId())) {
```
imaykay: I am sorry for having to say something in every ticket. But I was just asking myself, if the fillVendor method itself is not badly implemented. Because this one tries to check its precondition but might just do it wrong. Because it only checks if the vendorid has been set and not that it is set different from null or empty.
Even though I am not a 100% sure if the vendor should be removable once it has been set (for which setting it to null or empty explicitly would be necessary), but this code paths would break in this case as well, wouldn't it?
So maybe better fix fillVendor in case that anyone else might want to use this method in this class.

nutanv1contr: Thanks @imaykay for the suggestion, As of now there is only one execution flow for fillVendor() so my idea was to not enter into the flow if vendorId is missing which is happening for new release .Yes you are right if tomorrow someone else tries to use the method it may break.I will make changes inside fillVendor() method before thrift call (thriftClients.makeVendorClient()) to save the flow from crashing.

nutanv1contr: I have added a check inside fillVendor() method. I have not removed any of the existing check as I am not sure about the implication of removing them.

imaykay: Hi @nutanv1contr and again sorry for being picky! But why don't we go with a combination of both if clauses like
`if(release.isSetVendorId() && !isNullOrEmpty(release.getVendorId())) {`
as the first line in the method? I think this would be the optimal solution because
* we do only create a Thrift client when we really will use it
* if the vendorid has not been set it cannot be null or empty, so it fails faster
* we still keep the existing check

But these are so minor improvements that I would also approve your change if you don't have the time to do another adjustment.

nutanv1contr: Thanks for the suggestion.I have pushed the suggested changes.

imaykay: Thanks a lot @nutanv1contr

mcjaeger:
(stays on release details page and says in red colour: " Release could not be added, since a release with the same name and version already exists. ")

PR #459, opened 2019-01-29
fix(licenseinfo): Exclude old commons-lang3 dependencies

maierthomas:
- fix velocity dependencies problem

`
NoSuchMethodError: org.apache.commons.lang3.StringEscapeUtils.escapeXml
`

Signed-off-by: Thomas Maier <thomas.maier@evosoft.com>

PR #474, opened 2019-02-22
fix(license-import): add missing dependency

maxhbr:
Due to the changes in https://github.com/eclipse/sw360/commit/8fee82570cc54018540dc1e53457cfa08216c8b3#diff-522eb230e2bc520bfed39d0abf54dd21 (for https://github.com/eclipse/sw360/pull/459) the **import all SPDX licenses** feature is broken:

This PR adds a replacement for the excluded dependency.

The added dependency `org.apache.commons:commons-lang3:3.8.1` is already in the eclipse system, and we can piggyback of `[CQ18249] Apache commons-lang3 3.8.1` (see: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19043)

### DIFF
--- a/backend/src/src-licenseinfo/pom.xml
+++ b/backend/src/src-licenseinfo/pom.xml
@@ -53,6 +53,11 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.8.1</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.velocity</groupId>
             <artifactId>velocity-engine-core</artifactId>
             <version>2.0</version>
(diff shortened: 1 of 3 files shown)


maxhbr:
Until this feature is fixed, one can get the same result with the following steps:
1. download the attached file ([SPDX.zip](https://github.com/eclipse/sw360/files/2894528/SPDX.zip))
2. rename the file from `SPDX.zip` to `SPDX.lics`
3. import the result as a license archive

(The attached file was generated on a fresh instance with this branch by importing all SPDX licenses and exporting the resulting DB as lics archive. The renaming is necessary, since only specific extensions are allowed to be uploaded)

mcjaeger:
should be in for `sw360-3.4.0-M1`

lepokle:
Looks good but we should add the explicit dependency in src-licenses as well. Afterwards it can be merged.

PR #478, opened 2019-02-27
fix(users): write screenname into externalID field

### LINKED ISSUE #477: SW360 can not handle users created in liferay or via LDAP import
If one starts a fresh SW360, with deployed war files and lar files, one can observe the following behaviour:

SW360 uses **only one single user in the CouchDB** for every liferay user, which authenticates. This means that this user gets lots of `formerEmailAddresses` and the same UID is used on every interaction.

# Why?

The problem might be, that the `externalId` is not filled, on creation / update of the SW360 user.

Since the `externalId` of a user is taken from `liferayUser.getOpenId()` which returns `""`, we have that `externalId==""`. That matches the setup user, and thus we have found a matching user, which we update.

# How to reproduce
## Step 1: login as `setup@sw360.org`
After that, in the couchdb there is an entry created, corresponding to the setup user

## The result:
On login, the new user is written synchronized to the CouchDB, but sw360 identifies the wrong matching account. Instead it reuses the setup user account. The result looks like:

maxhbr:
This tries to fix #477, but it fails to fix existing instances. The problem is that in existing instances there is already a user with multiple email addresses. There is no direct way to differentiate that user back into individual users.

This might also collide with other approaches, where the `openId` of a liferay user is actually used, for storing the external ID.

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/UserRepository.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/UserRepository.java
@@ -34,7 +34,7 @@ import java.util.Set;
         @View(name = "all",
                 map = "function(doc) { if (doc.type == 'user') emit(null, doc._id) }"),
         @View(name = "byExternalId",
-                map = "function(doc) { if (doc.type == 'user') emit(doc.externalid.toLowerCase(), doc._id) }"),
+                map = "function(doc) { if (doc.type == 'user' && doc.externalid) emit(doc.externalid.toLowerCase(), doc._id) }"),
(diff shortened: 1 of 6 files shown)


maxhbr:
This PR now fixes the issue actually in three times, but all changes look good and beneficial.

maxhbr:
This might have no implications to users created with the following code:
https://github.com/eclipse/sw360/blob/0a104fe9e79a221561bbc91ba433dd7934d8a288/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/users/UserPortletUtils.java#L82-L109

This code adds the `externalId` to both fields, `screenName` and `openId`.

PR #496, opened 2019-03-25
feat(thrift): add http proxy for thrift clients

bs-matil:
This adds the capability to set a proxy for the thrift client which enables liferay or other clients to reach a thrift service behind a proxy.

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftClients.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftClients.java
@@ -12,6 +12,10 @@
  */
 package org.eclipse.sw360.datahandler.thrift;
 
+import org.apache.http.HttpHost;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.impl.conn.DefaultProxyRoutePlanner;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentService;
 import org.eclipse.sw360.datahandler.thrift.components.ComponentService;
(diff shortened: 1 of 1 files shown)


libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftClients.java
```diff
@@ -88,9 +96,19 @@ private static TProtocol makeProtocol(String url, String service) {
...
+                DefaultProxyRoutePlanner routePlanner = new DefaultProxyRoutePlanner(proxy);
+                CloseableHttpClient httpClient = HttpClients.custom().setRoutePlanner(routePlanner).build();
+                thriftClient = new THttpClient(destinationAddress, httpClient);
+            } else {
+                thriftClient = new THttpClient(destinationAddress);
+            }
         } catch (TTransportException e) {
             log.error("cannot connect to backend on " + destinationAddress, e);
+        } catch (MalformedURLException e) {
+            log.error("cannot connect via http proxy (REASON:MalformedURLException) to thrift backend:  " + BACKEND_PROXY_URL, e);
```
maxhbr: does the `BACKEND_PROXY_URL` maybe contain basic auth credentials (in some configurations)? In that case it might better not to log it.

mcjaeger:
tested on our side was just finished. @bs-matil, why did you close it?

bs-matil:
extentialy

nutanv1contr:
I performed a test on the PR.Below are the scenarios tested and there result.
1.Deploy frontend and backend in different servers and access backend by setting backend.url property in sw360.properties file in the server which has frontend deployed.
**Result:** Success
 Able to fetch the data from backend and display it in frontend. 
2.Tried to perform above scenario with wrong backend url in w360.properties.
**Result:** Success
Application data is not getting fetched into frontend.
3.Configure proxy server (Squid is used for testing) and specify the proxy server url with backend.proxy.url in sw360.proeprties and try to access application.
**Result:** Success
 Able to access and view the application data in fronend.
4.Shutdown the proxy server or give wrong proxy url in sw360.properties file  and try to access the application.
**Result:** Success:
Not able to fetch application data .

PR #497, opened 2019-03-25
Bs matil/implement keycloak oauth token mapper

bs-matil:
This feature adds the capability to use keycloak as authorization server


mcjaeger:
@maxhbr I think it could be Ok to test it this way. Please let me know if you prefer more testing.

mcjaeger:
it is ready

PR #501, opened 2019-03-27
fix(ui): MyComponents and MyProjects does not list in Home page if email is changed

### LINKED ISSUE #302: MyComponents and MyProjects does not look for old e-mails
If a user changes the e-mail, the view of the projects (and components) within the `Home` page is not showing the projects and components referring to the old e-mail.

The query should consider all e-mails of the user (actual and former) to perform this query.

nutanv1contr:
#### Summary:
 If a user changes the e-mail, the view of the projects (and components) within the Home page is not showing the projects and components referring to the old e-mail.
### Root Cause:
In home page earlier the search was based on  user email because of which when we change the email application is not able to find any result for that.As per latest changes projects and components are searched based on user id.
#### Steps to reproduce:
1.Login to the application and check the projects and components list on the home page.
2.Now go to "My Account" and change the email address for the logged in user and save the changes.
3.Now go back to home page and refresh the page.

#### Result:
1.It should show the same list of  all projects and components associated to the user. 

Closes #302

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/homepage/MyComponentsPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/homepage/MyComponentsPortlet.java
@@ -46,7 +46,7 @@ public class MyComponentsPortlet extends Sw360Portlet {
         List<Component> components;
         try {
             final User user = UserCacheHolder.getUserFromRequest(request);
-            components = thriftClients.makeComponentClient().getMyComponents(user);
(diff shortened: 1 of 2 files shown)


smrutis1:
@nutanv1contr  I followed the mentioned test steps, but unable to view the list of projects or components associated with the user after coming back to homepage post modifying the user email. However, after login with the modified email address, the same is showing properly. 

This might be because after modifying the user email in "MyAccount" section, the user details are getting removed from the cache. Please have a look.

mcjaeger:
(from discussion) actually, it is Ok to relogin in this case. (because changing)

PR #507, opened 2019-04-02
feat(report): add common rules table

hemarkus:
Add and fill the common rules table in the report. Therefore use Todo to
hold the actual common rule (project level obligation).

Signed-off-by: Markus Herpich <markus@endocode.com>


libraries/lib-datahandler/src/main/thrift/projects.thrift
```diff
@@ -14,6 +14,8 @@ include "attachments.thrift"
 include "vendors.thrift"
 include "components.thrift"
 include "sw360.thrift"
+include "commonobligations.thrift"
```
imaykay: why is this and the additions in line 18 and 37 necessary? (disclaimer: i am no thrift expert)

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
```diff
@@ -197,6 +239,26 @@ public static String printFullname(Release release) {
         return getReleaseFullname(vendorName, release.getName(), release.getVersion());
     }
 
+    public static Map<Todo, TodoInfo> getProjectObligations(Project project) {
```
imaykay: just a quick question regarding the naming: is todo and obligation the same or not?

libraries/lib-datahandler/src/main/thrift/projects.thrift
```diff
@@ -156,6 +160,13 @@ struct ProjectWithReleaseRelationTuple {
     2: required ProjectReleaseRelationship relation,
 }
 
+struct ProjectTodo {
+    1: required string todoId;
+    2: required string userId;
+    3: required string updated;
```
imaykay: The rest of the code suggests that this is a timestamp? Maybe use another data type? Some thrift definitions uses something like `i64` which seems to become a long in the java object then.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortletUtils.java
```diff
@@ -107,12 +108,53 @@ public static void updateProjectFromRequest(PortletRequest request, Project proj
...
+                    break;
                 default:
                     setFieldValue(request, project, field);
             }
         }
     }
 
+    private static void updateProjectTodosFromRequest(String[] ids, String userId, Project project) {
+        Set<String> idSet = ids != null ? Arrays.stream(ids).collect(Collectors.toSet()) : Collections.emptySet();
+        Set<ProjectTodo> currentTodods = project.getTodosSize() > 0 ? project.getTodos() : Collections.emptySet();
```
imaykay: Just a little type in todos. If you should push some changes anyway, you could fix this as well.

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
```diff
@@ -177,6 +182,8 @@ private void fillReportDocument(
             fillOverview3rdPartyComponentTable(document, projectLicenseInfoResults);
             fillAdditionalRequirementsTable(document, obligationResults);
 
+            fillCommonRulesTable(document, project);
```
imaykay: Looking at the table indices above, I would suggest, that the common rules table should be before the additional requirements table in the final document. Again, it would be easier for other developers if the method order could reflect the order in the document...

libraries/lib-datahandler/src/main/thrift/projects.thrift
hemarkus: this is just remains of another approach. good catch

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
hemarkus: A todo is an obligation, but not the other way around.

libraries/lib-datahandler/src/main/thrift/projects.thrift
hemarkus: Actually all thift definitions I found use a string, that's why I chose a string.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortletUtils.java
hemarkus: oops will fix

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
hemarkus: Yeah, right. I will fix this

libraries/lib-datahandler/src/main/thrift/projects.thrift
imaykay: Yeah, it seems we have both versions in the code base. It's fine then.

mcjaeger:
tested-by:michael.c.jaeger@siemens.com

Issue #483, opened 2019-03-08
License Info from Subprojects


mcjaeger:
License Info from Subprojects is saved at the top project. Downside is that in case the subproject is important multiple times, the license (file) selection needs to be repeated. The main point is to allow the user to pick also a license (file) selection from the subproject rather than creating an own one.

mcjaeger:
(from discussion)

Actually there was a discussion between:

* taking over the data and copy to local project data structure.
* or, using the "call by reference" from the sub project.

-> we decided to copy the data and then the user can make adaptations. Reason for this is primarily because the project record cannot be preserved fully as it was at the time when the reference would be made.

another topic: What happens, if the same release and attachment is referenced twice in different subprojects?

Then, the system should store selection individually, also considering the project id, where the relase is directly linked to.

mcjaeger:
(from discussion)

There is a discussion that on copy of the selections, the user should be warned if existing selections will be overwritten. Showing of the error message should be connected when a usage was found - as the user is shown the message telling that a setting has been restored.

PR #508, opened 2019-04-04
feat(subproject licenses): added possibility to take over license selection from subproject

imaykay:
* changed page behavior: same attachments of same releases are no longer
coupled in the tree if they belong to different subprojects
* license file contains the union of all selected licenses (so if the
same license is not selected on one leaf but is selected on another, it
will be part of the license file)
* added new field in LicenseInfoUsage to store the project path to this
attachment
* added button to take over the license selection of a subproject in a
fire and forget manner
* added loading spinner row
* source code bundle generation should still be the same

closes eclipse/sw360#483

The basic idea of the new license info file generation page is, that licenses are identified by the pattern
`projectid:subprojectid1:subprodjectid2:releaseid:attachmentcontentid:license`.
So the chain of (sub-)projectids has been added as an identifier in this PR. This was necessary to decouple the licenses in the same attachment at the same release in different project tree parts, so that one can take over the license selections for a subproject subtree without changing anything outside of this subtree.

Regarding testing, there are a lot of possibilities to test. Especially combinations with bigger subproject trees, where the same release (and therefore the same license relevant attachments) is available in different parts of this tree. Not only the UI is interesting in these cases when (un-)folding different tree parts first, then take over subproject selections or doing it the other way around - but also the license file generation. Especially the cases where a license is selected in one part of the tree and not selected in another branch.
In addition, loading of the last selection can be tested when entering the process again.

And as said before the source code bundle generation should be tested because it shares a lot of code with the license file generation. But it should be untouched by these changes.

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/AttachmentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/AttachmentDatabaseHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2016. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2016, 2019. Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
  *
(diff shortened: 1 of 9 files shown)


imaykay:
@lepokle pls re-review

Issue #519, opened 2019-05-02
Project spreadsheet export failure


nutanv1contr:
**Summary:**
During "Export to spreadsheet" for project.if  any of the cell value which to be exported  exceeds 
32767 (max allowed)  value then application shows an error page "Internal Server Error".

**Steps to reproduce:**
1.Create a project and and link around 460 releases to the project so that when user user select export to spreadsheet for the project then "releases   with usage" column will have more then 32767 characters.
2.Now try to export all projects from projects tab by selecting "Export to spread sheet".
**Expected Result:**
If the cell value exceeds 32767 (max allowed) characters then when just display an error message inside #cell has exceeded max number of characters.
**Actual Result:**
Application gives an error page  "Internal Server Error : An error occurred while accessing the requested resource.

PR #520, opened 2019-05-02
fix:Project spreadsheet export failure due to maximum cell length reached

nutanv1contr:
**Summary:**
If the cell value exceeds 32,767 characters then application should not display error page
Instead it should put value as "Export to spread sheet"

Added check for maximum cell limit during writing the values into row. 

**Steps to reproduce:**
1.Create a project and and link around 460 releases to the project so that when user user select export to spreadsheet for the project then "releases with usage" column will have more then 32767 characters.
2.Now try to export all projects from projects tab by selecting "Export to spread sheet".
**Actual Result:**
If the cell value exceeds 32767 (max allowed) characters then when just display an error message inside #cell has exceeded max number of characters.

Closes: #519

### DIFF
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
@@ -11,6 +11,7 @@
 package org.eclipse.sw360.exporter;
 
 
+import org.apache.poi.ss.SpreadsheetVersion;
 import org.apache.poi.ss.usermodel.*;
 import org.apache.poi.xssf.streaming.SXSSFSheet;
 import org.apache.poi.xssf.streaming.SXSSFWorkbook;
@@ -96,7 +97,11 @@ public class ExcelExporter<T, U extends ExporterHelper<T>> {
         }
         for (int column = 0; column < helper.getColumns(); column++) {
             Cell cell = row.createCell(column);
-            cell.setCellValue(values.get(column));
+            if(values.get(column).length() >= SpreadsheetVersion.EXCEL2007.getMaxTextLength()) {
+                cell.setCellValue("cell has exceeded max number of characters");
+            }else {
+                cell.setCellValue(values.get(column));
+            }
             cell.setCellStyle(style);
         }
     }


libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
```diff
@@ -96,7 +97,11 @@ private void fillRow(Row row, List<String> values, CellStyle style) {
         }
         for (int column = 0; column < helper.getColumns(); column++) {
             Cell cell = row.createCell(column);
-            cell.setCellValue(values.get(column));
+            if(values.get(column) != null && (values.get(column).length() >= SpreadsheetVersion.EXCEL2007.getMaxTextLength())) {
+                cell.setCellValue("cell has exceeded max number of characters");
+            }else {
```
mcjaeger: please correct space

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
```diff
@@ -96,7 +97,11 @@ private void fillRow(Row row, List<String> values, CellStyle style) {
         }
         for (int column = 0; column < helper.getColumns(); column++) {
             Cell cell = row.createCell(column);
-            cell.setCellValue(values.get(column));
+            if(values.get(column) != null && (values.get(column).length() >= SpreadsheetVersion.EXCEL2007.getMaxTextLength())) {
```
mcjaeger: please correct space

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
```diff
@@ -96,7 +97,11 @@ private void fillRow(Row row, List<String> values, CellStyle style) {
...
+                cell.setCellValue("cell has exceeded max number of characters");
```
mcjaeger: it would be better visible in the generated spreadsheet if the would be a leading hashmark like "#cell has exceed may number of characters"

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
nutanv1contr: Corrected the same. thanks

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
nutanv1contr: Thanks for the suggestion @mcjaeger . I have made the suggested changes.

mcjaeger:
@nutanv1contr can you please make the formatting of the commit message also conventional changelog style? instead of

Screen Shot 2019-05-03 at 13 52 13

it should be like

fix(report): adding coverage if content exceeds the max number of characters in cell

mcjaeger:
test ran successfully locally on my machine I am not sure if makes sense to wait hours for travis ...

mcjaeger:
premature merge! I am sorry for that. at commit `ccbfeb3 ` the file looked like 

```
    /**
     * Write the values into the row, setting the cells to the given style
     */
    private void fillRow(Row row, List<String> values, CellStyle style) {
        if(values.size() < helper.getColumns()){
            throw new IllegalArgumentException("List of row values is too short.");
        }
        for (int column = 0; column < helper.getColumns(); column++) {
            Cell cell = row.createCell(column);
            if (values.get(column) != null && (values.get(column).length() >= SpreadsheetVersion.EXCEL2007.getMaxTextLength())) {
                cell.setCellValue("#cell has exceeded max number of characters");
            } else {
                cell.setCellValue(values.get(column));
            }
            cell.setCellStyle(style);
        }
    }
```

PR #523, opened 2019-05-06
feat(spdx-import): added functionality to view and use spdx information

imaykay:
* added additional infos to existing licenseinfo object (filled only from spdx parsers)
* added portlet methods to get an array of concluded license ids and to use this info as input for a release update
  * implemented in a way that can be extended if other information should be handled as well
* added some javascript handling in clearing details tab of releases to offer actions and show results
* added POST-redirect-GET pattern to release update to avoid duplicate release updates when triggering a reload because of the new functionality after coming from edit mode
* improved error message for not found licenses

closes eclipse/sw360#92

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
@@ -13,6 +13,7 @@ package org.eclipse.sw360.portal.portlets.components;
 
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
@@ -34,10 +35,13 @@ import org.eclipse.sw360.datahandler.couchdb.lucene.LuceneAwareDatabaseConnector
 import org.eclipse.sw360.datahandler.permissions.PermissionUtils;
 import org.eclipse.sw360.datahandler.thrift.*;
 import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
+import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentType;
 import org.eclipse.sw360.datahandler.thrift.codescoop.CodescoopService;
 import org.eclipse.sw360.datahandler.thrift.components.*;
 import org.eclipse.sw360.datahandler.thrift.cvesearch.CveSearchService;
 import org.eclipse.sw360.datahandler.thrift.cvesearch.VulnerabilityUpdateStatus;
+import org.eclipse.sw360.datahandler.thrift.licenseinfo.LicenseInfoParsingResult;
+import org.eclipse.sw360.datahandler.thrift.licenseinfo.LicenseInfoService;
 import org.eclipse.sw360.datahandler.thrift.projects.Project;
 import org.eclipse.sw360.datahandler.thrift.projects.ProjectService;
 import org.eclipse.sw360.datahandler.thrift.users.RequestedAction;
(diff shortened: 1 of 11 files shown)


imaykay:
Hi @mcjaeger - would you mind to do a pre-test and give some feedback on the current version of this PR?  A working round trip is implemented with some open points in mind:
* Is UI/UX okay for the users from your point of view?
* I am currently showing attachments of types license-info-combined and license-info-xml, is that correct?
* Currently we might get LicenseRef IDs instead of the final license ID - do we need to resolve them? Are these references used often in SPDX documents?
* Do we need check whether this sw360 instance knows the license denoted by each ID?

maxhbr:
@imaykay can you make screenshots and present them here?

imaykay:
@maxhbr The current workflow is as follows:

Before
---
image

Step 1
---
image

Step 2
---
image

Step 3
---
image

After
---
image

/cc @mcjaeger

mcjaeger:
(from discussion)
* maybe cut off prefix `LicenseRef-`
* from testing, let have a look to see if POST-redirect-GET technique helps
* if license is not in DB: maybe change behavior from license portlet, instead of showing "Error fetching license from backend": "License ID $id not found in database."

imaykay:
Pushed an improved version that addresses the three points:
> maybe cut off prefix LicenseRef-

done
> from testing, let have a look to see if POST-redirect-GET technique helps

done (currently only for update release path - lets see how this behaves and then do it for all POSTs)
> if license is not in DB: maybe change behavior from license portlet, instead of showing "Error fetching license from backend": "License ID $id not found in database."

done (the id cannot be written easily into the error message because of our very generic way to use error messages. but it is available in the logs)

lepokle:
Result from testing:
- If the license file is empty one can click add anyway but nothing happens. Maybe better write "no licenses found" and disable the button
- the license is always added to the field, is that intended?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
```diff
@@ -499,6 +500,76 @@ private void writeJsonResponse(String json, ResourceResponse response) throws IO
...
+            log.error("Cannot write JSON response for attachment id " + attachmentContentId + " in release " + releaseId
+                    + ".", e);
+            response.setProperty(ResourceResponse.HTTP_STATUS_CODE, "500");
+        }
+    }
+
+    private void writeSpdxLicenseInfoIntoRelease(ResourceRequest request, ResourceResponse response) {
+        User user = UserCacheHolder.getUserFromRequest(request);
+        String releaseId = request.getParameter(PortalConstants.RELEASE_ID);
+        ObjectMapper om = new ObjectMapper();
```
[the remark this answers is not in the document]
imaykay: changed

imaykay:
>     * If the license file is empty, one can click add anyway, a success message is shown but nothing happens. Maybe better write "no licenses found" and disable the button

changed
>     * the license is always added to the field, is that intended?
I cannot tell for sure if this is the correct behavior. But I would suggest to not delete anything in this workflow step - especially when not warning the user about possible data loss.

frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/clearingDetails.jspf
```diff
@@ -167,6 +201,115 @@
...
+                url.setParameter(tableData.attachmentIdParameterName, attachmentId);
+
+                replaceContentWithSpinner($btnCell);
+
+                $.ajax({
+                    type: 'GET',
+                    dataType: 'json',
+                    url: url.toString(),
+                }).done(function(result) {
+                    if (!result || result.length == 0 || Object.getOwnPropertyNames(result) == 0) {
```
lepokle: You should add an `.length` to getOwnPropertyNames as it returns an array.

imaykay: done :-)

PR #525, opened 2019-05-07
feat(REST-Doc): Updated the REST API Documentation

smrutis1:
- Added the dcumentation for linking releases to the project.
- Updated the documentation for creating the project with linked project and linked release.

### DIFF
--- a/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
+++ b/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
@@ -41,6 +41,7 @@ import java.text.SimpleDateFormat;
 import java.util.*;
 
 import static org.eclipse.sw360.datahandler.thrift.MainlineState.MAINLINE;
+import static org.eclipse.sw360.datahandler.thrift.MainlineState.OPEN;
 import static org.eclipse.sw360.datahandler.thrift.ReleaseRelationship.CONTAINED;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Matchers.anyObject;
@@ -452,12 +453,18 @@ public class ProjectSpecTest extends TestRestDocsSpecBase {
 
     @Test
     public void should_document_create_project() throws Exception {
-        Map<String, String> project = new HashMap<>();
+        Map<String, Object> project = new HashMap<>();
         project.put("name", "Test Project");
         project.put("version", "1.0");
         project.put("visibility", "PRIVATE");
         project.put("description", "This is the description of my Test Project");
         project.put("projectType", ProjectType.PRODUCT.toString());
+        Map<String, ProjectReleaseRelationship> releaseIdToUsage = new HashMap<>();
+        releaseIdToUsage.put("3765276512", new ProjectReleaseRelationship(CONTAINED, OPEN));
+        project.put("linkedReleases", releaseIdToUsage);
+        Map<String, ProjectRelationship> linkedProjects = new HashMap<String, ProjectRelationship>();
+        linkedProjects.put("376576", ProjectRelationship.CONTAINED);
+        project.put("linkedProjects", linkedProjects);
 
         String accessToken = TestHelper.getAccessToken(mockMvc, testUserId, testUserPassword);
         this.mockMvc.perform(post("/api/projects")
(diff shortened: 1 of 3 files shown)


mcjaeger:
(from discussion) should be instead of "MAINLINE" rather "OPEN" according to user feedback for the project release relation.

smrutis1:
@mcjaeger Have updated the PR with mentioned changes.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
```diff
@@ -483,4 +492,14 @@ public void should_document_create_project() throws Exception {
...
     }
+
+	@Test
+	public void should_document_create_releases() throws Exception {
+		List<String> releaseIds = Arrays.asList("3765276512", "5578999", "3765276513");
+
+		String accessToken = TestHelper.getAccessToken(mockMvc, testUserId, testUserPassword);
+		this.mockMvc.perform(post("/api/projects/" + project.getId() + "/releases").contentType(MediaTypes.HAL_JSON)
+				.content(this.objectMapper.writeValueAsString(releaseIds))
+				.header("Authorization", "Bearer " + accessToken)).andExpect(status().isCreated());
```
imaykay: Shouldn't the `.andDo(this.documentationHandler.document(...` part be coming after `.andExpect()` so that all the input request params and response params get explained?

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
```diff
@@ -483,4 +492,14 @@ public void should_document_create_project() throws Exception {
                                 fieldWithPath("_embedded.createdBy").description("The user who created this project")
                         )));
     }
+
+	@Test
+	public void should_document_create_releases() throws Exception {
```
imaykay: I know, that the method in the rest controller is also called 'create releases', but I find that confusing since no one creates releases. A release is linked, but not created. Anyway, since the controller uses this strange naming, we can leave it here I think (or we should change both, the controller and the documentation here).

smrutis1: Updated the method names in both the places.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
smrutis1: This API only accepts an array of release ids as request body, there won't be any other params/fields as such and returns only http status and header information as response, there won't be any other response params either. So I think no need to add the documentation for the same. However to make it clearer for the user, I have updated the description section mentioning about what should be the request body. Please let me know, if it is ok.

nutanv1contr:
Tested the PR. 
1.Rest API document is updated with new end point to link release to project.
2.Create Project request fields documentation is listing two new fields linkedReleases.linkedProjects.
3.For "Linked Releases to Project" the  release relation is "Contained" and project mainline state is "Open".
**Concerns**
1.The sub properties under linkedReleases and linkedProjects should be documented or not? @mcjaeger 
2.While testing the curl request for new fields if I give the linked Projects as "CONTAINED" then it is not reflecting in UI.the value for Project Relation is "Is a Project".Is this the expected behavior?

mcjaeger:
@hieupvtsdv thanks for your feedback, please help us to see your post by considering to create an issue in such cases rather than adding a comment to a closed PR

smrutis1:
@hieupvtsdv , Could you please share the request payload. It will be helpful to identify and correct the issue, if any?

hieupvtsdv:
Request:
curl -X POST -H "Authorization: Bearer ${access_token}" -H "Content-Type: application/json"  http://localhost:8080/resource/api/projects/02eb80266df89e883f1af93c98003fc5/releases -d '{"creatBy" : "test@liferay.com" , "releaseIds" : "02eb80266df89e883f1af93c98000e54"}'
---------------
Response:
{"status":400,"error":"Bad Request","message":"JSON parse error: Can not deserialize instance of java.util.ArrayList out of START_OBJECT token; nested exception is com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.util.ArrayList out of START_OBJECT token\n at [Source: java.io.PushbackInputStream@73668b2d; line: 1, column: 1]","timestamp":"2019-05-21T03:34:17.857Z"}
Please help me!

PR #527, opened 2019-05-09
Revert "fix(ui): MyComponents and MyProjects does not list in Home page if email is changed"

mcjaeger:
Reverts eclipse/sw360#501 because it turned out not to be compatible with productive deployments.


nutanv1contr:
Performed a test .All the functionalities are working as expected in local.

Issue #528, opened 2019-05-10
Component edit looses ExternalId and Attachment changes


imaykay:
Since #267 the component edit mode looses changes to ExternalId and Attachments silently. 

Reason is a form inside the edit form, which makes the browser to end the edit form when the inner form begins so that all input fields after the inner form are not sent to the server.

Solution is to place the inner form after the edit form, which should be possible since it is only for a popup anyway.

PR #529, opened 2019-05-10
fix(component edit): fixed an issue where external id and attachment …

imaykay:
... changes were not saved

fixes eclipse/sw360#528


nutanv1contr:
Tested the PR by creating new components and updating  existing components with external id and attachments linked to the component.
Result:
Data is persisting after save operation and attachments are not getting deleted.

PR #531, opened 2019-05-15
feat(rest): Whitelisting Fields in the REST API Response

### LINKED ISSUE #487: REST API: components: field "homepage" missing
The REST API endpoint GET components doesn't provide component's homepage.

Plus, it seems to be impossible to update it using the PATCH request.

I consider this as rather important field to clearly describe publicly available components.

### LINKED ISSUE #516: REST API: Include Vulnerability flags when requesting a single project
When requesting a single project through the `api/projects/:id` endpoint, the two flags related to vulnerabilities: 

- Enable Security Vulnerability Monitoring
- Enable Displaying Vulnerabilities

should be included in the API response.

smrutis1:
* White listed "homepage" field at component level.
* White listed "enableSvm" and   "enableVulnerabilitiesDIsplay" at project level.

closes #487 , closes #516

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
@@ -34,7 +34,7 @@ import org.eclipse.sw360.datahandler.thrift.vulnerabilities.Vulnerability;
 import org.eclipse.sw360.datahandler.thrift.vulnerabilities.VulnerabilityDTO;
 import org.eclipse.sw360.rest.resourceserver.core.serializer.JsonProjectRelationSerializer;
 import org.eclipse.sw360.rest.resourceserver.core.serializer.JsonReleaseRelationSerializer;
-
+import org.eclipse.sw360.rest.resourceserver.project.EmbeddedProject;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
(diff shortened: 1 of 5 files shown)


smrutis1:
@maxhbr I have added the same in the post request. Please re-review.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/EmbeddedProject.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * 
+ * This class is used for json customization for embedded project.
+ * 
+ * @author smruti.sahoo@siemens.com
+ *
+ */
+@Relation(collectionRelation = "sw360:projects")
+public class EmbeddedProject extends Project {
```
maxhbr: Why does a project need this `EmbeddedProject` construction, but no other resource has something similar (although embedded users or licenses are implemented?)

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
```diff
@@ -182,6 +181,97 @@ public Sw360Module() {
             @JsonProperty("id")
             abstract public String getId();
         }
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+        @JsonIgnoreProperties({
```
maxhbr: This is mostly duplicated code. Is there maybe a option to say, that this the list for `Project` but with additional entries? Maybe by extending the `ProjectMixin`

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/EmbeddedProject.java
smrutis1: With the current implementation, its little difficult to blacklist the boolean fields (like `enableSvm,` `enableVulnerabilitiesDisplay` etc.) only in the project listing through the `api/projects` endpoint and white list it, when requesting for a single project through the `api/projects/:id` endpoint . That's why I have to add a subclass for the project to achieve the same. We can add the same implementation for other resources like users or licences, may be later, when we face same type of situation. Please let me know , if you have any other idea.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
smrutis1: Done, Thanks for the suggestion.

nutanv1contr:
Tested the PR. Functionality is working as expected.Fields are now behaving correctly on rest request and response.

Issue #530, opened 2019-05-15
component category to spreadsheet export


nutanv1contr:
For some reason, the component record field "Categories" is not part of the spreadsheet export. It should be added in the Spreadsheet export "projects with linked releases".

PR #532, opened 2019-05-15
feat(spreadsheet): Added component categories field on spreadsheet export of Project with linked releases

nutanv1contr:
**Summary:** Added "Categories" field of component to spreadsheet export of "Project with linked releases".

**Steps to reproduce:**
1. Go to Projects tab and select "Projects with linked releases" from the drop own and click on "Export Spreadsheet".

**Result:**
1.Categories column should be present on the spreadsheet with value as corresponding component categories of the linked release to project.

closes : #530

### DIFF
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseExporter.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseExporter.java
@@ -109,6 +109,7 @@ public class ReleaseExporter extends ExcelExporter<Release, ReleaseHelper> {
             case COMPONENT_ID:
                 headers.add(displayNameFor(field.getFieldName(), nameToDisplayName));
                 headers.add(displayNameFor("component type", nameToDisplayName));
+                headers.add(displayNameFor("categories", nameToDisplayName));
                 break;
             case VENDOR:
                 Vendor.metaDataMap.keySet().stream()
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
@@ -117,7 +117,7 @@ public class ReleaseHelper implements ExporterHelper<Release> {
                 // second, add joined data, remark that headers have already been added
                 // accordingly
 
-                // add component type in every case
+                // add component type and categories in every case
                 Component component = this.preloadedComponents.get(release.componentId);
                 if (component == null) {
                     // maybe cache was not initialized properly, so try to load manually
@@ -132,8 +132,14 @@ public class ReleaseHelper implements ExporterHelper<Release> {
                 // check again and add value
                 if (component == null) {
                     row.add("");
+                    row.add("");
                 } else {
                     row.add(ThriftEnumUtils.enumToString(component.getComponentType()));
+                    if (component.getCategories() == null) {
+                        row.add("");
+                    } else {
+                        row.add(component.getCategories().toString());
+                    }
                 }
 
                 // project origin and project mainline state only if wanted
--- a/backend/src-common/src/main/java/org/eclipse/sw360/components/summary/ComponentSummary.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/components/summary/ComponentSummary.java
@@ -70,6 +70,7 @@ public class ComponentSummary extends DocumentSummary<Component> {
         copyField(document, copy, Component._Fields.NAME);
         copyField(document, copy, Component._Fields.VENDOR_NAMES);
         copyField(document, copy, Component._Fields.COMPONENT_TYPE);
+        copyField(document, copy, Component._Fields.CATEGORIES);
 
         if (type == SummaryType.SUMMARY) {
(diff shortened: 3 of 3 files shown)


hieupvtsdv:
How to linked releases to projects using REST API?

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -134,6 +134,11 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
                     row.add("");
```
imaykay: Do we need a second `row.add("");` here? So that we do fill the new column with an empty value?

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -134,6 +134,11 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
...
                 } else {
                     row.add(ThriftEnumUtils.enumToString(component.getComponentType()));
+                    if (component.getCategories() == null) {
+                        row.add("");
+                    } else {
+                        row.add(component.getCategories().toString());
```
imaykay: Is it possible that categories are null so that you can get a NPE here? Probably not because I had a component without categories and it did not break - if I remember correctly. But I am not sure, if this is guaranteed.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseExporter.java
```diff
@@ -109,6 +109,7 @@ private static void addToHeaders(List<String> headers, Release._Fields field) {
             case COMPONENT_ID:
                 headers.add(displayNameFor(field.getFieldName(), nameToDisplayName));
                 headers.add(displayNameFor("component type", nameToDisplayName));
+                headers.add(displayNameFor("Categories", nameToDisplayName));
```
imaykay: All other header names are starting with a small letter (at least if they are no acronym)

nutanv1contr:
Hi @hieupvtsdv ,
Please refer https://github.com/eclipse/sw360/pull/525.

nutanv1contr:
> In the way the new column has been added, it is now also added to a release in the export of the component list export (go to tab "Components" and below the list of components choose "Components with releases" and click "Export Spreadsheet"). So in this export, we now have the column "categories" twice.
> Not sure if this is really a problem, but at least something unpleasant.

Thanks for pointing that out @imaykay . The same behavior is there for "component type also".The reason is , there is a common implementation for both project and release that is why the changes are getting applied for both the places.i have created a separate issue  #541 to address the same.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
nutanv1contr: In case the categories field is not set then we will have to set it with empty value otherwise application will through an exception for in consistency in number of rows and columns.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
nutanv1contr: Yes.I tried to create a component without setting the categories field and when I linked any of the release to project from the same category then in that case application crashes.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseExporter.java
nutanv1contr: Thanks for pointing it out.I have made the necessary changes.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
imaykay: Yes, that was also my impression. And that's why I asked if we need a second of these statements here. Because the first one is for the component type, so you probably should add a second for the categories, shouldn't you?

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
imaykay: Wait, it crashed or it didn't crash? If it crashed then you should probably do a null check before you call `toString()` on `getCategories()`.

imaykay: Sorry, my bad, you already added the null check!

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
nutanv1contr: Yes got it thanks. You are right,Even though the chance of component being null is minimal still if component is null we will need another row.add("").I have added the same

smrutis1:
Tested the PR successfully. The component categories field added in the project spreadsheet export.

Issue #539, opened 2019-05-20
Projects:New field required in "Add Project" : Domain


nutanv1contr:
A new field is required in "Add Project" under tab "Summary" in category "General"

New Field Details: 
Field Name : **Domain** 
Field Type : Dropdown 
Field Values : Application Software, Documentation, Embedded Software, Hardware, Test and Diagnostics.
* So the idea is to have a drop down field as for type and others while the value can be defined in sw360.properties.
* This field needs to be supported by the REST API. (see white listing of fields)

PR #540, opened 2019-05-20
feat(Projects): Added new field "domain" to project summary

nutanv1contr:
Summary:
Added new field "domain" to project summary
Field domain" will be a dropdown and the values will be fetched from sw360.properties file
Added rest api support for the field.
Added the field to rest api documentation

Steps to test:
-Create new project and look for a field with name "domain" under project summary.
-The field should of type dropdown and the values should be fetched from sw360.properties file.
-User should be able to select and save the field value.
-User should be able to add the field values by create project from rest end point.
-On entering invalid values rest end point should give error response. 

closes : #539

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/basicInfo.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/basicInfo.jspf
@@ -18,6 +18,7 @@
 
 <core_rt:set var="clearingTeamsStringSet" value='<%=PortalConstants.SET_CLEARING_TEAMS_STRING%>'/>
 <core_rt:set var="preferredCountryCodes" value='<%=PortalConstants.PREFERRED_COUNTRY_CODES%>'/>
+<core_rt:set var="domain" value='<%=PortalConstants.DOMAIN%>'/>
 
 <table class="table info_table" id="ProjectBasicInfo">
     <thead>
@@ -94,6 +95,29 @@
                       name="<portlet:namespace/><%=Project._Fields.DESCRIPTION%>" rows="4" cols="30"
                       style="width: 200px; height: 25px; resize:both;"
                       placeholder="Enter Description"><sw360:out value="${project.description}" stripNewlines="false"/></textarea>
+        </td>
+    </tr>
+    <tr>
+        <td width="33%">
+            <label class="textlabel stackedLabel" for="proj_domain">Domain</label>
+             <select class="toplabelledInput" id="domain" name="<portlet:namespace/><%=Project._Fields.DOMAIN%>" data-selected="${project.domain}"
+                style="min-width: 162px; min-height: 28px;">
+                <option value>-- Select Domain --</option>
+                <core_rt:forEach items="${domain}" var="entry">
+                    <c:choose>
+                        <c:when test="${entry eq project.domain}">
+                            <option selected="selected" value="<sw360:out value="${entry}"/>" class="textlabel stackedLabel"> ${entry} </option>
+                        </c:when>
+                        <c:otherwise>
+                            <option value="<sw360:out value="${entry}"/>" class="textlabel stackedLabel"> ${entry} </option>
+                        </c:otherwise>
+                    </c:choose>
+                </core_rt:forEach>
+            </select>
+        </td>
+        <td width="33%">
+        </td>
+        <td width="33%">
         </td>
     </tr>
     <tr>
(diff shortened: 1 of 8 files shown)


rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
```diff
@@ -8,6 +8,34 @@
  */
 package org.eclipse.sw360.rest.resourceserver.restdocs;
 
+import static org.eclipse.sw360.datahandler.thrift.MainlineState.MAINLINE;
```
maxhbr: Resorting the imports looks like an unnecessary refactoring for me. And what IDE do you use, I thin we usually sort the static imports to the bottom of the list and AFAIK this is also the way IntelliJ handles the imports?

akapti:
Tested the changes. working as expected.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectValidator.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.sw360.datahandler.thrift.projects.Project;
+import org.eclipse.sw360.rest.resourceserver.Sw360ResourceServer;
+import org.springframework.stereotype.Component;
+
+/*
+ * This class is used to validate the project input fields.
+ * @author nutan.das@siemens.com
+ * */
+@Component
+public class ProjectValidator {
```
imaykay: I like it a lot that you started to use input validation! Do you see any possibility to achieve the same by using more standardized techniques (based on JSRs) like mentioned in https://www.mkyong.com/spring-boot/spring-rest-validation-example/, https://www.baeldung.com/spring-boot-bean-validation or https://www.baeldung.com/spring-data-rest-validators?

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
nutanv1contr: For this case I have reverted the import reordering.

frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/basicInfo.jspf
```diff
@@ -96,6 +97,24 @@
                       placeholder="Enter Description"><sw360:out value="${project.description}" stripNewlines="false"/></textarea>
         </td>
     </tr>
+    <tr><
```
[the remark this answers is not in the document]
nutanv1contr: Sorry for that.I missed a small typo. Thanks for pointing that out.

imaykay:
Ah, one more thing: could you please double check the styling? It seems that the new row is missing its right border.
image

nutanv1contr:
> Ah, one more thing: could you please double check the styling? It seems that the new row is missing its right border.
> image

nutanv1contr:
> > Ah, one more thing: could you please double check the styling? It seems that the new row is missing its right border.
> > image

Thanks, made necessary changes.This issue seems to be there in chrome.I tested this in fire fox it was working fine.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectValidator.java
nutanv1contr: Hi @imaykay ,

Currently JSR validation is not working for @BasePathAwareController.(
https://github.com/spring-projects/spring-boot/issues/12863
https://jira.spring.io/browse/DATAREST-593
).
And I don't see any better way to do the validation here.Please suggest if there is any better solution to it.

mcjaeger:
(from discussion) move validation out here and do a clean approach for all of the fields that require validation

nutanv1contr:
Created a separate issue for rest end point field validation. https://github.com/eclipse/sw360/issues/584

mcjaeger:
failure at cvs-search such as also in other branches

lepokle:
@nutanv1contr 
The GUI is implemented in a way, that opening existing projects and saving them will automatically set the first domain value as new domain.

I think it would be better if domain remains unset until it is explicitly selected during project edit.

nutanv1contr:
> @nutanv1contr
> The GUI is implemented in a way, that opening existing projects and saving them will automatically set the first domain value as new domain.
> 
> I think it would be better if domain remains unset until it is explicitly selected during project edit.

Done.

PR #550, opened 2019-05-26
feat(codescoop): support release v0.5.0

sweetca:
Signed-off-by: Alex Skorohod <alexey.skorohod@gmail.com>

Support latest data model & auth changes

### DIFF
--- a/backend/src/src-codescoop/src/main/java/org/eclipse/sw360/codescoop/CodescoopHandler.java
+++ b/backend/src/src-codescoop/src/main/java/org/eclipse/sw360/codescoop/CodescoopHandler.java
@@ -48,27 +48,41 @@ public class CodescoopHandler implements CodescoopService.Iface {
     public CodescoopHandler() {
         Properties props = CommonUtils.loadProperties(CodescoopHandler.class, PROPERTY_FILE);
         String host = props.getProperty(PROPERTY_HOST, null);
-        String client = props.getProperty(PROPERTY_CLIENT, null);
+        String user = props.getProperty(PROPERTY_CLIENT, null);
         String credentials = props.getProperty(PROPERTY_CREDENTIALS, null);
         String proxy = props.getProperty(PROPERTY_PROXY, null);
 
         log.info("Codescoop client construct : host=" + host + ", client=" + client
                 + ", credentials=" + credentials + ", proxy=" + proxy);
 
-        try {
-            if (isNotBlank(host) && isNotBlank(client) && isNotBlank(credentials)) {
-                this.client = ClosableCodescoopClient
-                        .build()
-                        .withApiUrl(host)
-                        .withApiUser(client)
-                        .withApiKey(credentials)
-                        .withProxyHost(proxy)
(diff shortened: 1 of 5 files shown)


backend/src/src-codescoop/src/main/java/org/eclipse/sw360/codescoop/CodescoopHandler.java
```diff
@@ -48,27 +48,41 @@
...
-        String client = props.getProperty(PROPERTY_CLIENT, null);
+        String user = props.getProperty(PROPERTY_CLIENT, null);
         String credentials = props.getProperty(PROPERTY_CREDENTIALS, null);
         String proxy = props.getProperty(PROPERTY_PROXY, null);
 
         log.info("Codescoop client construct : host=" + host + ", client=" + client
                 + ", credentials=" + credentials + ", proxy=" + proxy);
 
+        if (isNotBlank(host) && isNotBlank(user) && isNotBlank(credentials)) {
+            while (client == null) {
```
imaykay: Would it be good to have some limit on retries? Like five times or something? Since I have no account and not even the necessary jar for building / trying, I cannot judge if it makes sense to try over and over again and what would be expected by sysadmins if the remote service is not reachable (which might probably be one of the reasons why the client creation fails).

libraries/lib-datahandler/src/main/thrift/codescoop.thrift
```diff
@@ -51,7 +51,7 @@ struct CodescoopComponentOrigin {
 }
 
 struct CodescoopSearchFilter {
-    1: required list<i32> type;
+    1: required list<string> type;
```
imaykay: I suppose nothing is stored of the thrift requests? So this change should not be a problem.

If older exchanged objects would be persisted somewhere, this might lead to problems if they should be exchanged again.

imaykay:
@maxhbr, @mcjaeger: what was the idea of the thrift service versions? Since we have a change in existing fields and not just an addition of a new one, is it still only increase of minor version or would this be an increase of the major version?

libraries/lib-datahandler/src/main/thrift/codescoop.thrift
```diff
@@ -98,7 +103,7 @@ struct CodescoopComponent {
     3: required string owner,
     4: required string name,
     5: required string purl,
-    6: required i32 type,
+    6: required string type,
```
maxhbr: I think thrift requests, that one should also change the number in the beginning to a number that is new, e.g.
```suggestion
    // 6: required i32 type, // type changned to string
    9: required string type,
```

maxhbr:
I think since thrift is only internal API and this does not involve a database migration, it is only `increases minor version`

maxhbr:
changed from `needs general test` to `needs special test` since the test needs a codescoop connection.

mcjaeger:
maybe there are too many changes since then, so it is an alternative to close. @sweetca, please reopen if necessary.

Issue #559, opened 2019-06-12
Project UI : Set to default text feature is not working correctly for License info header


nutanv1contr:
Steps to reproduce;
1.Open project in edit mode.
2.Go to administration tab and click on "Set to Default Text" button under License Info Header.

Actual Result:
1.If the text contains  double quote then the corresponding texts will not be displayed on the screen.
Expected result:
1.The license info header should be displayed as it is by adding "Set to Default Text".

akapti:
Tested the changes, and functionality is working as expected.
Double quoted text is not getting removed when "Set to Default Text" button is clicked.

PR #560, opened 2019-06-12
fix(Project UI): Fixed "Set To Default Text" feature for project license info header

nutanv1contr:
Issue: Under Project>>Administration, if user click on "Set To default text" of License info header then License info header text will display only till first double quote.

Reason: When html encounters the first double quote then it considers that as end of default text and display only till that point.
Fixed the issue by using the text area value in side java script which is  formatted as per html.

Closes : #559

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/administrationEdit.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/administrationEdit.jspf
@@ -187,7 +187,7 @@
         <td>
             <textarea name="<portlet:namespace/><%=Project._Fields.LICENSE_INFO_HEADER_TEXT%>" rows="5"
                       class="projectLicenseInfoHeaderText broader" id="licenseInfoHeaderText"
-                      data-defaulttext="${defaultLicenseInfoHeaderText}"><sw360:DisplayLicenseInfoHeader
+                      data-defaulttext="<c:out value='${defaultLicenseInfoHeaderText}'/>"><sw360:DisplayLicenseInfoHeader
                     project="${project}" defaultText="${defaultLicenseInfoHeaderText}"/></textarea>
             <input id="setToDefaultLicenseInfoHeaderTextButton" type="button" class="addButton pull-right"
                    value="Set To Default Text">
@@ -225,4 +225,3 @@
 
     });
 </script>
-


akapti:
@nutanv1contr 
Tested the PR, and it's working as expected.

frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/administrationEdit.jspf
```diff
@@ -216,7 +216,7 @@
         });
 
         function setDefaultTextAndResize(textarea) {
-            $(textarea).val($(textarea).data("defaulttext"));
+            $(textarea).val($(textarea).val());
```
imaykay: I have the feeling that this change is not doing what it's supposed to do. If I am correct then a buttenclick now just takes the current value of the textarea and sets this exact same value into the textarea. So the button should actually have no effect any more. Or am I misunderstanding something?

In general, there might be even more places where to check if the text is placed into the html in an escaped manner. The `org.eclipse.sw360.portal.tags.DisplayObligations.doTag()` is doing it correct for example. Maybe it would be best to search the codebase for `defaultObligationsText` NOT case sensitive and check all occurences. I also thought of passing the text from `org.eclipse.sw360.licenseinfo.LicenseInfoHandler.getDefaultObligationsText()` but that might not be correct because the service should not implicitly assume that one wants to place the text into an html attribute. So probably every usage of the text has to be checked.

nutanv1contr: Yes you are right about the implementation. I have modified the implementation to format the text to display special symbols in html. I had to pass the "defaultlicenseinfoheader" inside a <c:out/> jsp tag which takes care of the formatting. 
 defaultObligationsText  is also having the same issue right now. As the default-text is passed directly without any formatting.This can be discussed and fixed also.

mcjaeger: > This can be discussed and fixed also.
let's try to fix other occurrences too

smrutis1:
Tested this PR. It's working fine.

imaykay:
@nutanv1contr did you create another issue to look for similar problems or did you check other occurrences and came to the conclusion that they are all fine?

imaykay:
Since @smrutis1 tested this PR, I removed the label.

nutanv1contr:
> @nutanv1contr did you create another issue to look for similar problems or did you check other occurrences and came to the conclusion that they are all fine?

Yes , I have created a new issue to track the same. #582

Issue #569, opened 2019-06-19
REST API: Update project end point should be present


nutanv1contr:
-Currently there is no rest end point to update project.User with correct access rights should be able to update project from rest end point.
-Rest API documentation should be updated with the update project request and response details.

PR #570, opened 2019-06-19
feat(Rest-Project) : Added possibility to update project from rest endpoint.

nutanv1contr:
-Added update project rest end point.
-Updated the Rest API documentation for the same.

closes: #569

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -331,6 +331,16 @@ public class RestControllerHelper<T> {
         halResource.addEmbeddedResource("sw360:projects", halProject);
     }
 
+    public Project updateProject(Project projectToUpdate, Project requestBodyProject) {
+        for(Project._Fields field:Project._Fields.values()) {
+            Object fieldValue = requestBodyProject.getFieldValue(field);
+            if(fieldValue != null) {
+                projectToUpdate.setFieldValue(field, fieldValue);
+            }
+        }
+        return projectToUpdate;
+    }
+
     public Component updateComponent(Component componentToUpdate, Component requestBodyComponent) {
         for(Component._Fields field:Component._Fields.values()) {
             Object fieldValue = requestBodyComponent.getFieldValue(field);
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -340,6 +340,19 @@ public class ProjectController implements ResourceProcessor<RepositoryLinksResou
         }
     }
 
+    @PreAuthorize("hasAuthority('WRITE')")
+    @RequestMapping(value = PROJECTS_URL + "/{id}", method = RequestMethod.PATCH)
+    public ResponseEntity<Resource<Project>> patchProject(
+            @PathVariable("id") String id,
+            @RequestBody Project updateProject) throws TException {
+        User user = restControllerHelper.getSw360UserFromAuthentication();
+        Project sw360Project = projectService.getProjectForUserById(id, user);
+        sw360Project = this.restControllerHelper.updateProject(sw360Project, updateProject);
+        projectService.updateProject(sw360Project, user);
+        HalResource<Project> userHalResource = createHalProject(sw360Project, user);
+        return new ResponseEntity<>(userHalResource, HttpStatus.OK);
+    }
+
     @RequestMapping(value = PROJECTS_URL + "/{projectId}/attachments", method = RequestMethod.POST, consumes = {"multipart/mixed", "multipart/form-data"})
     public ResponseEntity<HalResource> addAttachmentToProject(@PathVariable("projectId") String projectId,
                                                               @RequestPart("file") MultipartFile file,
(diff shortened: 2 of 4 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -330,6 +330,16 @@ public void addEmbeddedProject(HalResource halResource, Project project) {
         halResource.addEmbeddedResource("sw360:projects", halProject);
     }
 
+    public Project updateProject(Project projectToUpdate, Project requestBodyProject) {
+        for(Project._Fields field:Project._Fields.values()) {
+            Object fieldValue = requestBodyProject.getFieldValue(field);
+            if(fieldValue != null) {
```
imaykay: I know that you implemented this logic the same way like it is for Release and Component. But do you have any idea why we wouldn't want to let the user reset fields (set them to null)?
Would it maybe be helpful to check with `isSet(Field)` - even though I cannot tell how this would work out with spring deserialization.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -340,6 +340,19 @@ public void downloadClearingReports(
...
 
+    @PreAuthorize("hasAuthority('WRITE')")
+    @RequestMapping(value = PROJECTS_URL + "/{id}", method = RequestMethod.PATCH)
+    public ResponseEntity<Resource<Project>> patchProject(
+            @PathVariable("id") String id,
+            @RequestBody Project updateProject) throws TException {
+        User user = restControllerHelper.getSw360UserFromAuthentication();
+        Project sw360Project = projectService.getProjectForUserById(id, user);
+        sw360Project = this.restControllerHelper.updateProject(sw360Project, updateProject);
+        projectService.updateProject(sw360Project, user);
```
imaykay: Again, I am pretty sure you just used the implementation for Release and Component as a blueprint. But the `projectService.updateProject()` returns a `RequestStatus` object which we might want to inspect to check if the update succeeded.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
```diff
@@ -495,6 +496,83 @@ public void should_document_create_project() throws Exception {
...
+                        fieldWithPath("version").description("The version of the new project"),
+                        fieldWithPath("visibility").description("The project visibility, possible values are: "
+                                + Arrays.asList(Visibility.values())),
+                        fieldWithPath("description").description("The project description"),
+                        fieldWithPath("projectType").description("The project type, possible values are: "
+                                + Arrays.asList(ProjectType.values())),
+                        fieldWithPath("securityResponsibles").description("An array of users responsible for security of the project.")),
+                responseFields(fieldWithPath("name").description("The name of the project"),
+                        fieldWithPath("version").description("The project version"),
+                        fieldWithPath("createdOn").description("The date the project was created"),
```
imaykay: I just stumbled upon that field because I think from the UI, the user is not able to modify the `createdOn` field. There might be others as well. Do we want to let users change it via REST?

imaykay: Ah, I just checked that `org.eclipse.sw360.datahandler.db.ProjectDatabaseHandler.copyImmutableFields(Project, Project)` will not let any service user change the fields. So the behavior might be correct, but the documentation is misleading.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
imaykay: Well, the `isSet(Field)` seems to do nothing else then a null check... too bad. So I have no other spontaneous idea how to actually enable users to reset a value.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
imaykay: Oh man, sorry for this monologue, but the documentation is alright. I just did not see in this diff view that there has been already the call of `responseFields()` before.
Just one question: What happens if one just passes another valid field of a project inside the request (one of the undocumented)? Will it be updated? Or is there some logic somewhere that restricts the patch request to the fields that you documented here?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
nutanv1contr: Hi @imaykay ,
There is a RequestStatus check happening inside projectService.updateProject(sw360Project, user) ,where it checks for the success status.

imaykay: Ah damn, I only checked the diff here in github and thought this would already be the thrift call.

The differentiation of the possible status is still not good in the service, but this should not be part of this ticket.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
nutanv1contr: The same update behavior is followed for project as we have for component and release. Any editable field can be provided inside request json to be updated. I gave few fields as part of sample request json.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
nutanv1contr: As you said, the logic is implemented the same was as we have for Release and Component. I agree this needs to be improved for all update operation . Currently if user wants to reset the field then he can pass "" string . if the field is not passed in the request json then that field will be ignored.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
nutanv1contr: I have updated the rest documentation accordingly.

smrutis1:
Tested few scenarios to update the project(e.g. updating the name, description, version, security responsible etc. along with updating the linked project and linked releases), its working fine and the documentation also looks good.

Issue #571, opened 2019-06-20
Add comment field for release 


akapti:
We found several cases in SW360 where the component-release and its internal attachments do not properly describe the component.  
Especially for modified components a comment field would be helpful to describe the modification in a few words.  
Eg: that GPL affected sample/test code has been removed.

Another case are components where we have certain variations, like commercial OTS cleared for a dedicated Business Line only.

This issue is about a "simple" comment in a 'release' to provide more background information for this specific release.  
Currently we only have the option to give a description (eg. purpose) inside a component. But there is no option to define what might be special for a 'release'.  
We have releases which have a 'Release Mainline State: Mainline" which are "factory default component, cleared according to policy".  
But there are also 'Release Mainline State: Specific" releases where it is impossible to tell what exactly 'specific' means without reading the whole clearing report.  
So for this purpose we wish for a comment field to shortly note what is so 'Specific' on this release (eg. having a variant without GPL-affected sample code, or a "high level license clearing only").  
From that perspective no information in this 'Linked Releases and Projects' user interface is required.

mcjaeger:
@akapti  maybe comments such as in the clearing requests could be helpful. Would the part be so modular so it could be added to a number of documents (not only clearing requests)?

mcjaeger:
I think it was discussed recently that it would be good, to have a king of conversation with mixed application-generated comments and user-entered comments

akapti:
We have support for user-entered comments at project to release relationship level (linked release). PR: https://github.com/eclipse/sw360/pull/947
So for `Release Mainline State: Specific` relation we can specify the specific details in comments.

PR #572, opened 2019-06-20
feat(Release): added new field 'comment' in release summary

akapti:
**Summary:**
Added new field "**comment**" (text-area) in release summary [Component -> Release Overview -> Summary]
Added rest API  (GET, POST and PATCH) support and documentation for the field.

**Steps to test:**
- Create new Component or Edit existing Component and then go to "Release Overview".
- Create a new Release or Open the existing release (is present).
- User should be able to see and edit the new field "`comment`" in "`Release Summary`" tab.
- User should be able to add/update the field value while creating/updating the release via rest end point.

closes : #571 

Signed-off-by: Abdul Kapti <abdul.mannankapti@siemens.com>

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/summaryRelease.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/summaryRelease.jspf
@@ -12,6 +12,7 @@
 <%@ taglib prefix="tags" tagdir="/WEB-INF/tags" %>
 
 <core_rt:if test="${release != null}">
+    <div class="up_Summary" id="up_Summary"><p><sw360:out value="${release.comment}" maxChar="500"/></p></div>
     <table class="table info_table" id="ReleaseDetailOverview">
         <thead>
         <tr>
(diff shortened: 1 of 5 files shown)


frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/summaryRelease.jspf
```diff
@@ -12,6 +12,7 @@
 <%@ taglib prefix="tags" tagdir="/WEB-INF/tags" %>
 
 <core_rt:if test="${release != null}">
+<div class="up_Summary" id="up_Summary"><p><sw360:out value="${release.comment}" maxChar="500"/></p></div>
```
maxhbr: Formatting:
```suggestion
    <div class="up_Summary" id="up_Summary"><p><sw360:out value="${release.comment}" maxChar="500"/></p></div>
```

And why is it limited to 500 chars ?

akapti: @maxhbr The 500 chars limit is present in Component description as well. (file: `/sw360-portlet/src/main/webapp/html/components/includes/components/summary.jspf`).
So to make the release description ( view mode) in sync with component, I added that char limit.
Let me know if this has to be removed.

mcjaeger:
discussed to have a more general structure for comments (which is then also suitable for all documents, such as projects, vendors, components etc.) in some general place and then reference the structure:

Screenshot 2019-09-04 at 10 21 27

mcjaeger:
proposed change is so different, makes sense to open another PR

PR #574, opened 2019-06-26
feat(attachment-usages): Take over the attachment usages from the original project.

### LINKED ISSUE #573: Attachment usage is not available for a project's copy
When a user prepares a new version of a project, he/she generally makes a copy of the previous project. But the cloned project does not take over the attachment usages of the original project. It would be useful if this information was also copied over.

smrutis1:
- Fetched the attachment usages of the original project.
- Persisted the same by associating the new project(cloned project) id to it.

closes #573 


smrutis1:
@imaykay Thanks. I have made the necessary changes to replace the new `projectId` in the `projectPath` of the cloned project.

nutanv1contr:
Tested the changes.Now saved attachment usages on the parent project are reflecting on the cloned project.

PR #581, opened 2019-07-05
fix(rest): Hiding unwanted fields in project listing response in REST

smrutis1:
* There were some additional fields exposed through project listing ( `api/projects`) endpoint.  

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.sw360.rest.resourceserver.core;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -182,13 +183,14 @@ class JacksonCustomizations {
             abstract public String getId();
         }
 
-	@JsonInclude(JsonInclude.Include.NON_NULL)
-	@JsonIgnoreProperties({
-		"enableSvm",
-		"enableVulnerabilitiesDisplay"
-	})
 	static abstract class EmbeddedProjectMixin extends ProjectMixin {
+            @Override
+            @JsonIgnore
+            abstract public boolean isEnableSvm();
 
+            @Override
+            @JsonIgnore
+            abstract public boolean isEnableVulnerabilitiesDisplay();
 	}
 
         @JsonInclude(JsonInclude.Include.NON_NULL)


nutanv1contr:
Tested the changes. Now functionality is  working as expected.
Scenarios tested from REST:
1.Get all projects.
2.Get specific project.
3.Create project.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
```diff
@@ -184,6 +184,89 @@ public Sw360Module() {
...
+                "setSecurityResponsibles",
+                "setOwnerGroup",
+                "setOwnerCountry",
+                "rolesSize",
+                "setRoles",
+                "setOwnerAccountingUnit",
+                "setLicenseInfoHeaderText",
+                "setProjectOwner",
+                "setEnableSvm",
+                "setEnableVulnerabilitiesDisplay",
```
nutanv1contr: Please correct the indentation.

nutanv1contr:
Please correct the indentation, other then that code looks good.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
smrutis1: corrected the indentation by removing the tabs.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
```diff
@@ -184,8 +184,91 @@ public Sw360Module() {
...
+                "enableSvm",
+                "enableVulnerabilitiesDisplay"
 	})
 	static abstract class EmbeddedProjectMixin extends ProjectMixin {
```
maxhbr: These should not be necessary, since `EmbeddedProjectMixin` extends `ProjectMixin` and thus also derives the Ignore list from `ProjectMixin`. See discussion in https://github.com/eclipse/sw360/pull/531

smrutis1: Unfortunately, It's not behaving this way right now. There might be some deployment issue while testing the PR #531 in local environment previously. But while testing it in higher environment, we found that it's not behaving correctly, as it should.

maxhbr: But then there has to be a way to reduce duplicated code or to enforce, that the list for `EmbeddedProjectMixin` is larger then `ProjectMixin`

smrutis1: I have reduced the duplicated code. Kindly re-review.

nutanv1contr:
Test results are also successful.

Issue #554, opened 2019-06-06
POC: store generic or user specific data in SW360 DB


maxhbr:
Currently we have `externalIds` as a somewhat generic data store for primary keys in other systems.

There are also other kinds of "data" which a user might want to store, or add semantics to it. These values should be
- visible and correctly rendered in the UI (URL clickable, ...)
- editable in the UI
- supported by REST
- validated by some simple logic (e.g. email addresses should contain an `@`)

This implies a basic kind of typing, e.g. `String`, `URL`, `Email`, `Hash`, `List_of_String`, ...

Once this is implemented, one could replace some of the currently hard coded  but mostly unused fields with the generic option.

Examples for that are:
- URLs to the corresponding issue tracker,
- ID to SoftwareHeritage which is clickable and directs to the corresponding page,
- some hash value, which allows only for alphanumeric chars.
- Maven coordinates consisting of three strings

maxhbr:
Actually requested data to be stored:

| Name | Type |
|---|---|
| tbd. | |

maxhbr:
Thrift is able to do [unions](https://thrift.apache.org/docs/idl#union).

maxhbr:
Requirements:
- one should be able to configure, which "extensions" apply to which type in the db (e.g. Project, ...)
- one should be able to edit the extensions in the frontend
- the extensions should be displayed in the fronted
- the extensions should be supported by REST
- every type of extension should have a pattern or example to be displayed in the UI
- one should not need to recompile `lib-datahandler` to update the configuration
- It might be good if
  - there could be some kind of validation
  - there could be support for non-string content types (e.g. list of string)

maxhbr:
**Current result / opinion:**

I would recommend to not implement such a generic data structure with displaying, editing, rest support and semantics. This is way to complex for the benefit it yields.

mcjaeger:
I think the PR has been merged ...

PR #583, opened 2019-07-09
Add additionalData maps to releases, components, projects and licenses 

maxhbr:
maps for licenses are currently supported by the backend and rest, but not by the frontend or moderation requests (already true for external IDs).

This is a **partial** fix to https://github.com/eclipse/sw360/issues/554. it allows to store generic data but without typing or configuration possibilities.

TODOs:

- [x] Editing of additional Data with components does not work

### DIFF
--- a/backend/src-common/src/test/java/org/eclipse/sw360/components/summary/ProjectSummaryTest.java
+++ b/backend/src-common/src/test/java/org/eclipse/sw360/components/summary/ProjectSummaryTest.java
@@ -42,6 +42,9 @@ public class ProjectSummaryTest {
                 case EXTERNAL_IDS:
                     project.externalIds = Collections.emptyMap();
                     break;
+                case ADDITIONAL_DATA:
+                    project.additionalData = Collections.emptyMap();
+                    break;
                 case ATTACHMENTS:
                     project.attachments = Collections.emptySet();
                     break;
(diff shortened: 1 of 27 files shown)


maxhbr:
> 2. there is no gui for additional data for licenses

There is also no gui for externalIds for licenses, and thus it is consistend.

lepokle:
Hmm, that's correct but maybe it would a good idea to add this GUI as well? Maybe it is very easy by just including the two jsps?

maxhbr:
> Hmm, that's correct but maybe it would a good idea to add this GUI as well? Maybe it is very easy by just including the two jsps?

Yes sure, but I think it is something for a separate issue?

maxhbr:
> 1. additional data should be part of component merge as well (like external ids)

should be fixed by https://github.com/eclipse/sw360/commit/666af9571c323f2e42e8a21b0d40cf6591f13cd8, which is incorporated in the commit here.

mcjaeger:
@maxhbr what is the plan? I am asking because of tagging 4.0.0 in order to continue with the liferay 7 migration.

maxhbr:
I am working on it

maxhbr:
There is a bug in the component merge, which is out of scope of this PR: https://github.com/eclipse/sw360/issues/596

Issue #592, opened 2019-07-22
Fix CVE Search Tests


mcjaeger:
Currently, the master crashes:
```
Running org.eclipse.sw360.cvesearch.entitytranslation.CveSearchDataTranslatorTest
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

PR #593, opened 2019-07-23
fix(cve-search): disable tests by assume statement and refactor

maxhbr:
this should fix https://github.com/eclipse/sw360/issues/592

the line disabling the tests is: https://github.com/eclipse/sw360/pull/593/files#diff-a160378f662a9d3ddb8a062609ffac05R36

current state: ~~waiting for travis.~~ travis is green.

### DIFF
--- a/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/CveSearchApiImpl.java
+++ b/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/CveSearchApiImpl.java
@@ -11,10 +11,11 @@
  */
 package org.eclipse.sw360.cvesearch.datasource;
 
-import org.eclipse.sw360.cvesearch.datasource.json.CveSearchJsonParser;
 import com.google.common.base.Strings;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import org.eclipse.sw360.cvesearch.datasource.json.ListCveSearchJsonParser;
+import org.eclipse.sw360.cvesearch.datasource.json.SingleCveSearchJsonParser;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.apache.log4j.Logger;
 
@@ -31,7 +32,7 @@ import java.util.function.Function;
 
 public class CveSearchApiImpl implements CveSearchApi {
 
-    Logger log = Logger.getLogger(CveSearchApiImpl.class);
+    private Logger log = Logger.getLogger(CveSearchApiImpl.class);
 
     private String host;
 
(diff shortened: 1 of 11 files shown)


backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/CveSearchApiImpl.java
```diff
@@ -49,38 +50,39 @@ public CveSearchApiImpl(String host) {
...
-    private Object getParsedContentFor(String query, Function<BufferedReader,Object> parser) throws IOException {
+    private <T> T getParsedContentFor(String query, Function<BufferedReader,T> parser) throws IOException {
         log.debug("Execute query: " + query);
-        InputStream is = new URL(query).openStream();
-
-        if (is != null) {
-            try (BufferedReader content = new BufferedReader(new InputStreamReader(is, Charset.forName("UTF-8")))) {
-                return parser.apply(content);
+        try(InputStream is = new URL(query).openStream()) {
+            if (is != null) {
```
lepokle: Is this still necessary? Wouldn't the try exit if stream cannot be opened?

backend/src/src-cvesearch/src/test/java/org/eclipse/sw360/cvesearch/TestWithCveSearchConnection.java
```diff
@@ -0,0 +1,42 @@
...
+
+public abstract class TestWithCveSearchConnection {
+    private String PUBLIC_CVE_SEARCH_SERVER = "https://cve.circl.lu";
+
+    protected CveSearchApi cveSearchApi;
+
+    @Before
+    public void setUpApi() {
+        Properties props = CommonUtils.loadProperties(TestWithCveSearchConnection.class, "/cvesearch.properties");
+        String host = props.getProperty(CveSearchHandler.CVESEARCH_HOST_PROPERTY, "https://cve.circl.lu");
```
lepokle: Use constant here for the default value?

backend/src/src-cvesearch/src/test/java/org/eclipse/sw360/cvesearch/TestWithCveSearchConnection.java
```diff
@@ -0,0 +1,42 @@
...
+
+        Assume.assumeTrue("The public CVE Search server is currently unreliable and tests are only executed agains other instances.", ! PUBLIC_CVE_SEARCH_SERVER.equals(host));
```
lepokle: Isn't it testing that you have configured another CVE server than the public one? Therefore the message sould be someting like "Another then the public server has been configured. ...."

maxhbr: if the default public server is used, `! PUBLIC_CVE_SEARCH_SERVER.equals(host)` is false and thus triggers the assume statement. This then says that the tests are skipped since the default public server is not reliable.

lepokle: Thanks for clarification! So the statement is: we never test against the public CVE server because it is too unstable.

I think the currently has mislead me because it sounds like: we've checked the public server right now and unfortunately it is **currently** unreliable. So for **this** time we skip the tests.

Maybe remove it. Then it is clear that we think that the public server is unreliable in general.

lepokle:
Code looks good. Nice refactoring!

PR #598, opened 2019-07-24
fix(user export): fix Nullpointer Exception on user export

lepokle:


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/admin/UserPortlet.java
```diff
@@ -205,7 +205,7 @@ public void backUpUsers(ResourceRequest request, ResourceResponse response) thro
             org.eclipse.sw360.datahandler.thrift.users.User sw360user = UserCacheHolder.getUserFromEmail(emailAddress);
             boolean wantsMailNotification =
                     sw360user.isSetWantsMailNotification() ? sw360user.wantsMailNotification : true;
-            String userGroup = sw360user.getUserGroup().toString();
+            String userGroup = sw360user.getUserGroup() != null ? sw360user.getUserGroup().toString() : null;
```
maxhbr: alternatively:
```suggestion
            String userGroup = Optional.ofNullable(sw360user.getUserGroup())
                .map(Object::toString)
                .orElse(null); 
```

PR #602, opened 2019-07-31
fix(chores): updated documentation including licenses file

mcjaeger:
* removed outdated license files
* updated slightly readme

according to https://reuse.software/practices/2.0/

Signed-off-by: Michael C. Jaeger <michael.c.jaeger@siemens.com>


maxhbr:
- I would like to preserve the unchanged LICENSE top level file
- instead of a file `LICENSES` I think it would be better to have a folder `LICENSES` with files `LICENSES/MIT.LICENSE` and `LICENSES/CC-0.LICENSE`

mcjaeger:
(from discussion)

there is a pro and con about the licenses folder which adds maybe a lot of files and also is more difficult to keep up to date (that a file)

I think it could be OK to  call this 'license' because many mechanisms seem to be working on that fact that there is such file.

mcjaeger:
> * I would like to preserve the unchanged LICENSE top level file
> * instead of a file `LICENSES` I think it would be better to have a folder `LICENSES` with files `LICENSES/MIT.LICENSE` and `LICENSES/CC-0.LICENSE`

I see a problem for maintaining it, because the generation such files is more overhead that just putting this into a single file.

PR #605, opened 2019-08-02
fix(Rest-API): Small fix around ProjectClearingState during create and update project

nutanv1contr:
1.While creating project from rest end point, "Project clearing state" should set to "open" by default if not set explicitly.
2. Only users with Admin rights should be allowed to edit a closed project.Small fix was needed from  for this feature to work from rest end point.

Steps to test:
*  Scenario 1:
   *  Create a project from rest end point without setting the "Project clearing state" field.
   *  Project clearing state value should set to "Open" by default.
* Scenario 2:
  *  Open a project and mark the "Project clearing state" as "closed".
  *  Now try to update the same project from rest end point. If the user does not have admin rights then the user should not be allowed to edit the project.
  *  Users with admin rights  should still be allowed to edit the closed project.
  *  Change the "Project clearing state" as "open" .Now all the users who has access to modify the projects should be allowed to edit the project from rest end point.


smrutis1:
Tested the above mentioned scenarios, its working fine.

maxhbr:
Please rebase

nutanv1contr:
I have done the rebase . Issue around update project from rest end point is already merged with master as part of another PR.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
```diff
@@ -269,6 +270,9 @@ private static void validateLicenseIdMatch(License license) throws SW360Exceptio
     public static void prepareProject(Project project) throws SW360Exception {
         assertNotEmpty(project.getName());
         project.setType(TYPE_PROJECT);
+        if(!project.isSetClearingState()) {
+           project.clearingState =  ProjectClearingState.OPEN;
```
maxhbr: is there a `project.setClearingState(ProjectClearingState.OPEN);` ?

maxhbr: Whitespace and formatting problems:
```suggestion
            project.clearingState = ProjectClearingState.OPEN;
```

nutanv1contr: Thanks for pointing it out. Made the necessary changes.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
```diff
@@ -269,6 +270,9 @@ private static void validateLicenseIdMatch(License license) throws SW360Exceptio
     public static void prepareProject(Project project) throws SW360Exception {
         assertNotEmpty(project.getName());
         project.setType(TYPE_PROJECT);
+        if(!project.isSetClearingState()) {
+           project.setClearingState(ProjectClearingState.OPEN);
```
maxhbr: We indent with multiples of 4 spaces:
```suggestion
            project.setClearingState(ProjectClearingState.OPEN);
```

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
```diff
@@ -269,6 +270,9 @@ private static void validateLicenseIdMatch(License license) throws SW360Exceptio
     public static void prepareProject(Project project) throws SW360Exception {
         assertNotEmpty(project.getName());
         project.setType(TYPE_PROJECT);
+        if(!project.isSetClearingState()) {
```
maxhbr: We usually have a space after `if`:
```suggestion
        if (!project.isSetClearingState()) {
```

nutanv1contr: Done.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
nutanv1contr: Sorry for that.Done with the changes.

Issue #556, opened 2019-06-11
Is this bug of the adding to component?


BinhTH-tsdv:
I proceed to add new component. I add name field for the component: " " (space only), similar: i add category field: " "(space only). The following result:  New component is created successfully.
![image](https://user-images.githubusercontent.com/39551943/59250930-c281bc80-8c52-11e9-9c3d-894c55d01bd9.png)

I think the SW360 should report: "Should display error message such as: This field is required".
Is this bug?

VERSION: 3.4.0- COMMIT: 3a8562f
**PLEASE HELP ME?**

PR #608, opened 2019-08-07
fix(Components): Fix naming component error (component's name contains only space characters)

trungpl:
Add checking naming component error and error message.
Issue #556

Signed-off-by: Pham Le Trung <trung.phamle@toshiba-tsdv.com>


nutanv1contr:
Hi @trungpl ,
Few observations : 
1.  I am able to update the component with name as space character. Please add the same validation for update component flow.
2.  Enable the feature for rest end point as well.

PR #610, opened 2019-08-13
Lepokle/#398/update to liferay 7 2

lepokle:
See for instructions to upgrade: https://github.com/eclipse/sw360/wiki/Deploy-Upgrade-to-Liferay-7.2

Notes for early adopters that have already deployed a recent version:
- Redeploy with this branch instead
- Re-import all lar files

Note: single commits should make tracking changes easier.

PR #617, opened 2019-08-20
chores(liferay): Feedback from Liferay 7 review

lepokle:
This PR contains changes to resolve the feedback from PR #610.

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/homepage/AbstractTasksPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/homepage/AbstractTasksPortlet.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright Siemens AG, 2019. Part of the SW360 Portal Project.
+ *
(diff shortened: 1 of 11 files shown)


libraries/importers/src/main/java/org/eclipse/sw360/importer/ComponentCSVRecord.java
```diff
@@ -717,7 +717,7 @@ public String getVendorName() {
         elements.add(SampleOptions.URL_OPTION);
         elements.add(SampleOptions.URL_OPTION);
         elements.add(SampleOptions.URL_OPTION);
-
+
```
imaykay: This is no review, just by scrolling over the changeset these whitespaces in blank lines were something that I stumbled across. They seem to be in some java files.

lepokle: It was a feedback from nutan that these files didn't have relevant changes therefore it would be better to revert them.
Since this is a new PR now, it might be better to remove them from this change as a revert now does not have any advantages.

nutanv1contr:
Reviewed the changes. Code looks good to me now.

PR #619, opened 2019-08-26
feat(auth): script to add the unsafe default client directly to DB

maxhbr:


scripts/addUnsafeDefaultClient.sh
```diff
@@ -0,0 +1,85 @@
...
+   "client_id": "trusted-sw360-client",
+   "client_secret": "sw360-secret",
+   "access_token_validity": 3600,
+   "refresh_token_validity": 3600,
+   "authorized_grant_types": [
+       "refresh_token",
+       "client_credentials",
+       "password"
+   ],
+   "autoapprove": [
```
imaykay: Will not be used by the server anyway, so remove

scripts/addUnsafeDefaultClient.sh
```diff
@@ -0,0 +1,85 @@
...
+       "true"
+   ],
+   "resource_ids": [
```
imaykay: Will not be used by the server anyway, so remove

scripts/addUnsafeDefaultClient.sh
```diff
@@ -0,0 +1,85 @@
...
+   "scoped": true,
+   "description": "my first test client",
+   "scope": [
+       "READ", "WRITE", "ADMIN"
+   ],
+   "client_id": "trusted-sw360-client",
+   "client_secret": "sw360-secret",
+   "access_token_validity": 3600,
+   "refresh_token_validity": 3600,
+   "authorized_grant_types": [
```
imaykay: Will not be used by the server anyway, so remove

scripts/addUnsafeDefaultClient.sh
```diff
@@ -0,0 +1,85 @@
...
+deleteDefaultApp() {
+    curl --silent \
+         --output /dev/null \
+         -X DELETE "${TRUSTED_CLIENT_URL}?rev=${oldRev}"
+}
+
+oldRev="$(getDefaultAppRev)"
+
+if [[ "$oldRev" == "null" ]]; then
+    if [[ "$1" != "-d" ]]; then
```
imaykay: Please document the option somewhere. I would be fine with a usage comment inside this script at the beginning. It does not necessarily needs to be a printed help message on some -h switch or on wrong usage.

scripts/addUnsafeDefaultClient.sh
imaykay: I did not get during the first review that you are writing directly to the couch and not through the rest interface of the auth server. So you were right, we need these props!

scripts/addUnsafeDefaultClient.sh
imaykay: I did not get during the first review that you are writing directly to the couch and not through the rest interface of the auth server. So you were right, we need these props!

scripts/addUnsafeDefaultClient.sh
imaykay: I did not get during the first review that you are writing directly to the couch and not through the rest interface of the auth server. So you were right, we need these props!

imaykay:
As discussed in slack with @maxhbr, next time a solution would be preferred that uses official interfaces (like the REST interface for client management in this case).

PR #620, opened 2019-08-27
fix(rest): fix broken logic in updateProject

maxhbr:


lepokle:
Code looks good. However we should create a branch to update 4.0.0-M1 and create the PR against that branch?

maxhbr:
i think we can create the branch to update later, and cherry pick this commit

lepokle:
Also fine for me. But Travis should be ok before merging.

Issue #627, opened 2019-09-03
Merge Components misses fields


mcjaeger:
During liferay 7 testing it became apparent that some fields are not covered (maybe anymore):

* ~~Component type~~
* ~~Component category~~
* Owner Country

lepokle:
@mcjaeger The component name is not merged as well. Is that intentional or is it a bug?

lepokle:
I think we shouldn't allow merging the name. Normally you can merge into a component which name you would to retain. In other cases you could rename the component afterwards.

Allowing to merge has the following drawback: We have a constraint that prevents duplicate component names.
When merging components, first the source component is merged into the destination and then saved. Afterwards the source component is deleted. If the component name would have been changed to the source component, an error would be thrown because the name would then be duplicate (source and destination). In addition it would be a bad idea to remove the source component first to prevent this error. You will lost the source component in case of errors.
Therefore an extended workflow would be needed to merge the name:
1. merge component except of the name
2. delete source
3. rename destination

I'm not sure if that is really worth it since components can be renamed after merge as well.

Issue #208, opened 2018-04-26
Merge Components may fail in last step because of database update issue


maierthomas:
<a href="https://github.com/mcjaeger"><img src="https://avatars1.githubusercontent.com/u/3418977?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mcjaeger](https://github.com/mcjaeger)**
_Monday Jan 22, 2018 at 09:59 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/723_

----

(Alex and me) When merging a component, the last write to the target may fail at last write of plain fields at:

https://github.com/sw360/sw360portal/blob/e90785fb89e584eb3094fbc002787ddf7bdc9570/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java#L476

because of an update conflict when the releases are merged, but the components dataset is not retrieved again.

In the stack trace below merging releases is at rev 20 resulting in rev 21, then plain fields write following does not see rev 20 as latest, but Java code call is at that level.

```
2018-01-22 12:01:45 ERROR DatabaseConnector:180 - Document cannot be updated Component(id:1cfc47ff...)
 org.ektorp.UpdateConflictException: document update conflict: id: 1cfc47ff3694425ac5e98f23a2355d90 rev: 20-0fb05174942c3e1a559b1dd3eba9a27d
        at org.ektorp.impl.StdCouchDbConnector$9.error(StdCouchDbConnector.java:378)
        at org.ektorp.impl.StdCouchDbConnector$9.error(StdCouchDbConnector.java:364)
        at org.ektorp.http.RestTemplate.handleResponse(RestTemplate.java:126)
        at org.ektorp.http.RestTemplate.put(RestTemplate.java:47)
        at org.ektorp.impl.StdCouchDbConnector.update(StdCouchDbConnector.java:363)
        at org.eclipse.sw360.datahandler.couchdb.DatabaseConnector.update(DatabaseConnector.java:177)
        at org.ektorp.support.CouchDbRepositorySupport.update(CouchDbRepositorySupport.java:181)
        at org.eclipse.sw360.datahandler.db.ComponentDatabaseHandler.updateComponentInternal(ComponentDatabaseHandler.java:417)
        at org.eclipse.sw360.datahandler.db.ComponentDatabaseHandler.updateComponentCompletely(ComponentDatabaseHandler.java:593)
```

lepokle:
Two steps have to be done:
[...]

PR #640, opened 2019-09-10
fix(merge): Some fields were not merged

lepokle:
The field 'owner country' was not merged during component merge.

Closes: #627


nutanv1contr:
Hi @lepokle ,
There is one issue related to "owner country" field merge.I am not sure if this needs to be fixed as part of this PR.
1. When I merge two components without release then "Owner Country" field is getting merged successfully but if I merge two components which are containing releases then in that case "Owner Country" is not updating and below exception is appearing in log.

 ```org.ektorp.UpdateConflictException: document update conflict```

lepokle:
Hi @nutanv1contr,
this is not a cause of the field. This bug is related to the merge of releases inside components and wil be fixed with #208.
Please test with components that do not have any releases.

nutanv1contr:
- Code looks good to me. 
- "owner country" field is getting merged successfully if there is no release present on the component.

Issue #643, opened 2019-09-10
Attachment Usage is not taken over for License ReadMe Generation


nutanv1contr:
**Steps to reproduce:**
1. Create a new project which is linked with release containing SPDX / Component License Information attachment.
2. Select  any of the SPDX / Component License Information  attachment from AttachmentUsage tab.
3. Go to  Download License Info tab and check for the license selection.

**Actual result:**
1. All the licenses for the selected attachment should be checked but its not coming as selected.

PR #644, opened 2019-09-10
fix(DownloadLicenseInfo): Corrected license selection based on attachment selection on attachmentusage

nutanv1contr:
*   When user save the attachment usage  then project path was not getting saved in DB because of which license selection for the selected attachment was not appearing on download license info tab.
*  Now saving the project path, also during attachment usage save.

Closes: #643

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
@@ -220,7 +220,7 @@ public class ProjectPortlet extends FossologyAwarePortlet {
                         .filter(usage -> allUsagesByProject.stream()
                                 .noneMatch(isUsageEquivalent(usage)))
                         .collect(Collectors.toList());
-
+                setProjectPathToAttachmentUsages(usagesToCreate,project);
                 if (!usagesToDelete.isEmpty()) {
                     attachmentClient.deleteAttachmentUsages(usagesToDelete);
                 }
@@ -237,6 +237,14 @@ public class ProjectPortlet extends FossologyAwarePortlet {
             response.setProperty(ResourceResponse.HTTP_STATUS_CODE, Integer.toString(HttpServletResponse.SC_INTERNAL_SERVER_ERROR));
         }
 
+    }
+
+    private void setProjectPathToAttachmentUsages(List<AttachmentUsage> usagesToCreate, Project project) {
+        usagesToCreate.forEach(usage -> {
+            LicenseInfoUsage licenseInfo= usage.getUsageData().getLicenseInfo();
+            if (licenseInfo != null && !licenseInfo.isSetProjectPath())
+                licenseInfo.setProjectPath(project.getId());
+        });
     }
 
     private void serveAttachmentUsagesRows(ResourceRequest request, ResourceResponse response) throws PortletException, IOException {


mcjaeger:
@nutanv1contr please sign off the commit

nutanv1contr:
Thank you @mcjaeger .

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
```diff
@@ -239,6 +239,13 @@ private void saveAttachmentUsages(ResourceRequest request, ResourceResponse resp
 
     }
 
+    private void setProjectPathToAttachmentUsages(List<AttachmentUsage> usagesToCreate, Project project) {
+        usagesToCreate.forEach(usage -> {
+            if (!usage.usageData.getLicenseInfo().isSetProjectPath())
```
kallesoranko: Should we maybe check that `usage.usageData.getLicenseInfo() != null` ?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
```diff
@@ -239,6 +239,13 @@ private void saveAttachmentUsages(ResourceRequest request, ResourceResponse resp
...
+                usage.usageData.getLicenseInfo().setProjectPath(project.getId());
```
smrutis1: May be we can use `usage.getUsageData()` here.

smrutis1:
Tested this PR. Everything seems to working, except if there is no selection on attachment usages, then its by default selecting the first license file.

nutanv1contr:
> Tested this PR. Everything seems to working, except if there is no selection on attachment usages, then its by default selecting the first license file.

Hi @smrutis1 ,
This seems to be a default behavior when no saved license selection found.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
nutanv1contr: Refactored the code as per review comment.

nutanv1contr:
Refactored the code a little more and pushed the changes.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
mcjaeger: @nutanv1contr I am not sure if that is relevant, but could you check if that makes sense? If it was incorporated, please comment to the question to make the situation clear.

mcjaeger:
I think there is an old functionality that does not select the first file, but the file which was uploaded by a user of the same group.

I think changing the default file selection to the first file would actually confuse users.

mcjaeger:
added a ticket https://github.com/eclipse/sw360/issues/660

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectPortlet.java
nutanv1contr: Sorry I missed add my comment . I have modified and refactored the changes as per review comment.

smrutis1:
Thanks @mcjaeger for sharing some insight on this and creating a ticket for the same. Tested it again after incorporation of review comments. It's working as mentioned in my last comment.

mcjaeger:
merging, because build essentially went fine:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 08:51 min
[INFO] Finished at: 2019-09-19T08:34:37Z
[INFO] Final Memory: 182M/727M
[INFO] ------------------------------------------------------------------------
*** Error in `/usr/lib/jvm/java-8-openjdk-amd64/bin/java': double free or corruption (fasttop): 0x00007f9d1e270e30 ***
/home/travis/.travis/functions: line 109:  9172 Aborted                 (core dumped) mvn --fail-at-end $MVN_ARGS
The command "mvn --fail-at-end $MVN_ARGS" exited with 134.
```

Issue #647, opened 2019-09-11
REST: Get component by id not working for specific cases


smrutis1:
Currently, fetch component by id i.e through `/components/{id}` endpoint is not working for the components having default vendor id as empty(`""`) value.

PR #648, opened 2019-09-11
fix(rest): Fixed get component API having default vendor id as empty

smrutis1:
- Added validation for default vendor id. 

Closes: #647


smrutis1:
Added the same validation in `ComponentPortlet.java` as well.

nutanv1contr:
- Tested and reviewed the PR. 
- Code looks good to me. 
- Now get specific component from rest end point is working correctly.

**Steps followed to reproduce:**
1. Create a project from rest end point without providing vendor 
id on the request json.
2. Make get  specific component request from rest end point. Getting the result successfully. 
3. Get all components from rest end point is also working correctly.
4. Components are also loading successfully from application UI.

Issue #649, opened 2019-09-11
Scroll bar not appearing in link releases to project popup


nutanv1contr:
**Steps to reproduce:**
1. Open  a release and clink on "Link to Project" button and search for project.
2. In the popup there is no scroll bar displayed to scroll down and select project.

PR #650, opened 2019-09-11
fix(Release-UI): Vertical scrollbar for link release to project popup

nutanv1contr:
Added scrollbar for link release to project popup. 

**Steps to test:**
1. Open a release and click on "Link to Project" button.
2. On the search project popup scroll bar should be present to scroll down and select project.

closes: #649


smrutis1:
Tested this PR. The scroll bar is coming properly and the code also looks good.

PR #654, opened 2019-09-13
fix(EditRelease UI): Removed duplicate field 'Licenses' from edit release

nutanv1contr:
There were two fields for "Licenses" in edit release page. I have removed one entry to fix the same.


smrutis1:
Tested this PR. Duplicate one has been removed.

PR #665, opened 2019-09-23
feature(table-filter): add filter box, fix print

### LINKED ISSUE #657: Filter Box in Data Tables: general roundup
For . the filter text box in the data tables appearances: we seem to have manually enable it for every occurence, because sometimes they make sense sometimes not.

Proposal is to have a general mechanism which can be enabled / disabled on each data table occurence.

lepokle:
This commits adds an option to the generic create-method for datatables as
well as fixes the print button on some pages.

* Filter Box
There is no one more parameter to the create-method of the bridges/datatables
module. Giving true as last parameter will add a filter box at the top right
for the datatable.
This option has been set on many pages now, except where explicit quick filters
already exist.

* Fix print button
When adding a print button, the additional option 'orthogonal' has been set.
With this option in place, all render methods are called with type set to
'print' when printing. This way renderer can omit html tags or links when
rendering for print view. Renderes has been adapted as necessary.

Closes #657

### DIFF
--- a/frontend/liferay-theme/src/main/webapp/css/sw360/components/_table.scss
+++ b/frontend/liferay-theme/src/main/webapp/css/sw360/components/_table.scss
@@ -11,7 +11,8 @@
 
 table.table {
 	&.label-value-table {
-		border-color: $light-color;
+		// for some reason, the color is not applied when switching pages without the '!important' modifier.
+		border-color: $light-color !important;
 	}
 
 	th {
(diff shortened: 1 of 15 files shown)


nutanv1contr:
Tested the PR successfully. 
Print button is working correctly. Below are places where search filter is introduced and tested.
1. Project clearing status.
2. Project ECC status.
3. Project vulnerability.
4. release overview on specific component.
5. Release vulnerability.
6. Search component during component merge.
7. Using projects on project summary.

Issue #670, opened 2019-09-25
REST: Create project is returning error response.


nutanv1contr:
1. When user makes a create project request then in response it is giving 500 : internal server error.
2. Project is getting saved successfully in database but create project rest request is always giving the error response. 
3.It should return the create project data as response with success status. 

Currently the response is coming as:

![image](https://user-images.githubusercontent.com/46559892/65600133-29866f80-dfbd-11e9-827b-e82b00c4ec4f.png)

kallesoranko:
This is something that I also came across when trying to use Antenna's SW360Updater module to create project and other entries to SW360 with LR7. I believe this is the same issue.
Stacktrace for this:
```
sw360_1                | 2019-09-23 11:16:26 ERROR SW360Assert:135 - Invalid empty email UserService.java: 784
sw360_1                |  SW360Exception(why:null)
sw360_1                |        at org.eclipse.sw360.datahandler.common.SW360Assert.fail(SW360Assert.java:129)
sw360_1                |        at org.eclipse.sw360.datahandler.common.SW360Assert.assertNotEmpty(SW360Assert.java:85)
sw360_1                |        at org.eclipse.sw360.users.UserHandler.getByEmail(UserHandler.java:50)
sw360_1                |        at org.eclipse.sw360.datahandler.thrift.users.UserService$Processor$getByEmail.getResult(UserService.java:784)
sw360_1                |        at org.eclipse.sw360.datahandler.thrift.users.UserService$Processor$getByEmail.getResult(UserService.java:764)
sw360_1                |        at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:38)
sw360_1                |        at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39)
sw360_1                |        at org.apache.thrift.server.TServlet.doPost(TServlet.java:83)
sw360_1                |        at org.eclipse.sw360.projects.Sw360ThriftServlet.doPost(Sw360ThriftServlet.java:42)
sw360_1                | 2019-09-23 11:16:26 ERROR ProcessFunction:47 - Internal error processing getByEmail
sw360_1                |  SW360Exception(why:Invalid empty email UserService.java: 784)
sw360_1                |        at org.eclipse.sw360.datahandler.common.SW360Assert.fail(SW360Assert.java:136)
sw360_1                |        at org.eclipse.sw360.datahandler.common.SW360Assert.fail(SW360Assert.java:130)
[...]

PR #671, opened 2019-09-25
fix(REST Project) : Fixed error response for create project from rest

nutanv1contr:
Currently create project request from rest end point is returning `500 internal server error` as error response. 
Data is getting saved successfully in backend however while formulating the HATEOAS data it was failing.  Added the necessary fix for the same. 

Closes : #670


kallesoranko:
Tested this PR successfully and issue described in https://github.com/eclipse/sw360/issues/670 is removed

Issue #674, opened 2019-09-26
Acknowledgement notices are missing in TEXT generated document


smrutis1:
Acknowledgements are missing in the text format of license disclosure document.

Issue #369, opened 2018-09-24
Acknowledgements notices in the docx generated document missing


mcjaeger:
The acknowledgement are there in the text and html (xml) output, but not in the word processor output document and should be there too in the similar fashion as with the two other documents.

PR #675, opened 2019-09-26
fix(licenseDisclosure): Added acknowledgements in TEXT and DOCX format

smrutis1:
- Acknowledgements have been added in the TEXT and DOCx format disclosure documents which was missing earlier.

closes: #369 
closes: #674


nutanv1contr:
Tested the PR. Acknowledgement is now coming for both text an docx format license disclosure document. 
Code also looks good to me.

Issue #676, opened 2019-09-27
Component type field should be mandatory


nutanv1contr:
1. User is able to create component without specifying component type field value.
2. Component type field should be mandatory. 
![image](https://user-images.githubusercontent.com/46559892/65766653-4780db00-e149-11e9-86ff-28355881fbf9.png)

Issue #584, opened 2019-07-16
Standard validation for rest end point fields


nutanv1contr:
A cleaner standard validation is needed for all the rest end point fields that require validation.

mcjaeger:
for references it is there

PR #677, opened 2019-09-27
fix(Componnet): ComponentType field should be mandatory

nutanv1contr:
1. Added a fix to make the component type field as mandatory. 
2. This fix is  both for SW360 UI and Rest end point. 
3. From rest end point user might not get the exact error response with field name,  if component type input is left  blank.  Rest end point field validation  will be taken care as part of issue fix for  
#584 .

closes: #676

### DIFF
--- a/frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/includes/components/editBasicInfo.jspf
+++ b/frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/includes/components/editBasicInfo.jspf
@@ -50,8 +50,8 @@
     <tr>
         <td>
             <div class="form-group">
-                <label for="comp_componenttype">Component Type</label>
-                <select class="form-control" id="comp_componenttype" name="<portlet:namespace/><%=Component._Fields.COMPONENT_TYPE%>">
+                <label class="mandatory" for="comp_componenttype">Component Type</label>
+                <select class="form-control" id="comp_componenttype" required name="<portlet:namespace/><%=Component._Fields.COMPONENT_TYPE%>">
                     <option value=""></option>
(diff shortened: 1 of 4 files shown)


frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/includes/components/editBasicInfo.jspf
```diff
@@ -50,8 +50,8 @@
     <tr>
         <td>
             <div class="form-group">
-                <label for="comp_componenttype">Component Type</label>
-                <select class="form-control" id="comp_componenttype" name="<portlet:namespace/><%=Component._Fields.COMPONENT_TYPE%>">
+                <label class="mandatory" for="comp_componenttype">Component Type</label>
+                <select class="form-control" id="comp_componenttype" required name="<portlet:namespace/><%=Component._Fields.COMPONENT_TYPE%>">
```
smrutis1: Please add a appropriate error message for mandatory field component type.

nutanv1contr: Done.

nutanv1contr:
> While creating component from the rest endpoint without component type, its giving an error with 500 response code as below, which might be little confusing for the user. but as mentioned we can fix that as part of issue #584
> 
> `{ "status": 500, "error": "Internal Server Error", "message": "Internal error processing addComponent", "timestamp": "2019-10-04T15:37:00.547Z" }`

I have added a validation message at rest end point for create component with valid status code(400).

smrutis1:
Tested again. It's working now with appropriate error message from rest and UI both.

smrutis1:
Tested it after resolving merge conflicts. Its working fine.

PR #685, opened 2019-10-01
fix(Rest-Component): Corrected components by type rest end point

nutanv1contr:
1. From rest end point when user was requesting `get all components`  based on component type, it was failing with 500 error.
2. Corrected the same by adding necessary checks.


smrutis1:
Tested successfully. It's filtering out components based on component type.

PR #690, opened 2019-10-04
feat(Authorization): Added support to read keystore from central configuration

nutanv1contr:
1. Added support to read jwt-keystore.jks file from etc/sw360 folder, if not found then it will search from default project resource path. 
2. Read keystore secret key from application.yml . 

closes: #691

### DIFF
--- a/rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360AuthorizationServerConfiguration.java
+++ b/rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360AuthorizationServerConfiguration.java
@@ -10,13 +10,19 @@
  */
 package org.eclipse.sw360.rest.authserver.security;
 
+import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.rest.authserver.client.service.Sw360ClientDetailsService;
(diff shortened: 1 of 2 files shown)


rest/authorization-server/src/main/java/org/eclipse/sw360/rest/authserver/security/Sw360AuthorizationServerConfiguration.java
```diff
@@ -95,8 +105,12 @@ public TokenStore tokenStore() {
 
     @Bean
     protected JwtAccessTokenConverter jwtAccessTokenConverter() {
+        Resource resource = new FileSystemResource(new File("/etc/sw360/jwt-keystore.jks"));
```
smrutis1: Please remove the hard-coded file path and read it from `CommonUtils.java`

nutanv1contr: Thanks @smrutis1 , I have incorporated the changes.

mcjaeger:
note that build runs fine, but does not exit cleanly, should be OK, restarted anyways.

smrutis1:
Tested the PR. 
1) Reading the keystore password from `application.yml`.
1) Able to read the keystore from central configuration path(`/etc/sw360`).

PR #701, opened 2019-10-16
feat(moderation): Moderation requests to all clearing admin irrespective of their BU

smrutis1:
* Previously the moderation requests were going to the clearing admins of the same group. These changes send the moderation request to all clearing admins.


smrutis1:
Resolved test cases.

akapti:
Tested the changes, and working as expected.
Created 2 moderation requests for 2 projects of 2 different groups.
and clearing admins from both the group can see and accept/reject the moderation request.
Also, Code looks good to me.

One issue I noticed is that (maybe not w.r.t this change), Accepting the moderation request is not redirecting the page to moderation list, it stays on the same page, and this might give a wrong perception to user that moderation request is not accepted.
we can create new issue for this and work on it.

PR #708, opened 2019-10-24
fix(component): component merge not working

JaideepPalit:
Component merge was not working due error in Json parsing.

Issue Fixed.


frontend/sw360-portlet/src/main/resources/META-INF/resources/js/modules/sw360Wizard.js
```diff
@@ -178,7 +178,12 @@ define('modules/sw360Wizard', [ 'jquery', 'modules/button' ], function($, button
                     button.finish($('.wizardNext', $wizardRoot));
                 }).done(function(data, textStatus, xhr) {
                     try {
-                        var dataJson = JSON.parse(data);
+                        var dataJson = data;
+                        if (typeof data === 'string' || data instanceof String)
```
nutanv1contr: only `typeof` check should do. Do we need to add `instanceof` check also ?

JaideepPalit: Removed instanceof check. typeof check will serve the purpose

PR #725, opened 2019-11-12
fix(UI/REST): Remove Trailing and leading whitespace for all fields in component, release and project

JaideepPalit:
Remove Trailing and leading whitespace for all fields in component ,release and project during add or update through UI or REST

Validation for fields which should not contain only white space while creating or updating through UI or REST
- Name field for Project
- Name and Category field for Component
- Name and Version field for Release

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -110,7 +110,25 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
     public static final List<EccInformation._Fields> ECC_FIELDS = Arrays.asList(EccInformation._Fields.ECC_STATUS, EccInformation._Fields.AL, EccInformation._Fields.ECCN, EccInformation._Fields.MATERIAL_INDEX_NUMBER, EccInformation._Fields.ECC_COMMENT);
 
     private final MailUtil mailUtil = new MailUtil();
-
+    private static final ImmutableList<Component._Fields> listOfStringFieldsInCompToTrim = ImmutableList.of(
+            Component._Fields.NAME, Component._Fields.DESCRIPTION, Component._Fields.COMPONENT_OWNER,
+            Component._Fields.OWNER_ACCOUNTING_UNIT, Component._Fields.OWNER_GROUP, Component._Fields.OWNER_COUNTRY,
+            Component._Fields.HOMEPAGE, Component._Fields.MAILINGLIST, Component._Fields.WIKI, Component._Fields.BLOG);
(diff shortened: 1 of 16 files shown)


JaideepPalit:
Done some Refactoring of code.

mcjaeger:
apart from the merge conflicts, @maxhbr, is this OK now?

frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/includes/components/editBasicInfo.jspf
```diff
@@ -22,7 +22,7 @@
         <td>
             <div class="form-group">
                 <label class="mandatory" for="comp_name">Name</label>
-                <input id="comp_name" name="<portlet:namespace/><%=Component._Fields.NAME%>" type="text" placeholder="Enter Name" required class="form-control"
+                <input id="comp_name" name="<portlet:namespace/><%=Component._Fields.NAME%>" type="text" placeholder="Enter Name" required pattern=".*\S.*" class="form-control"
```
maxhbr: If you want to avoid trailing white-space the correct patter would be `pattern="\S.*\S"` (but this also enforces a length of at least 2 chars

JaideepPalit: Actually this pattern is used to solve problem where user can add only white space characters for ComponentName like `           `.
` pattern=".*\S.*" ` will prevent user from doing that from UI . However user can add something like 
`     Component123   `  with leading and trailing white-spaces but in the backend it will trim the whitespaces and save in DB.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -419,6 +449,15 @@ private Vendor getVendor(String vendorId) {
     ///////////////////////////////
 
     public RequestStatus updateComponent(Component component, User user) throws SW360Exception {
+        removeLeadingTrailingWhitespace(component);
```
mcjaeger: I am not sure about this particular part, the trim whitespace option will be maybe difficult to extend if particular fields are referenced like this example here

frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/includes/components/editBasicInfo.jspf
mcjaeger: @maxhbr would only whitespace search be OK here?

maxhbr: Its OK

smrutis1:
Tested successfully from REST and UI.

Issue #280, opened 2018-05-29
Changes of the Project/Release Mainline State should only be allowed for the role Clearing Admin


tngraf:
Unfortunately is has happend that project team members added new components to a project and immediately set the clearing state - this must not be. Only people in charge like Clearing Admins should be allowed to change the Project/Release Mainline State of a release.

mcjaeger:
(would be good if the user role level is configurable for this)

tngraf:
Thanks for implementing this, but it seems that there is some misunderstanding in another feature of SW360: When someone copies a project, then the mainline state **must** also get copied.
Example: There is a project that has 230 release. I have approved all of them (= set the proper mainline state). When the project lead now copies the project because they want to release a new version of the project, it is required to copy also all mainline states. Otherwise I have to go through all releases again, check if a can approve them and set the mainline state accordingly. I have done this now for several projects with more than 400 releases in total ... this costs hours!
From my point of view it would be sufficient just to deny write access to these fields in the UI. But it should not stop SW360 from copying my decisions.

PR #727, opened 2019-11-14
fix(project): Keep project mainline state as it is while cloning/creating project

smrutis1:
This change will only applicable when copying the projects.

closes: #280 


JaideepPalit:
Tested . Working fine .

PR #731, opened 2019-11-20
fix(ui): Added missing tooltip

smrutis1:
* Fixed the tool tip by converting the title attribute to title tag and added the missing tool tip for clearing state and project active state in the project listing page.


smrutis1:
Thanks @JaideepPalit for pointing it out. have corrected the same.

PR #734, opened 2019-11-29
feat(ProjectMigration): script to migrate a project field to new value

JaideepPalit:
Migration script to migrate a project field to new value.
Following variables needs to be updated based on requirement
`fieldName` = "field_to_be_migrated"
`oldValue` = "old_value"
`newValue` = "new_value"

For updating ClearingTeam Field in Project - `clearing.teams` property in `sw360.properties` file should be updated accordingly

### DIFF
--- a/scripts/utilities/001_update_project_field_value.py
+++ b/scripts/utilities/001_update_project_field_value.py
@@ -1,0 +1,88 @@
+#!/usr/bin/python
+# -----------------------------------------------------------------------------
+# Copyright Siemens AG, 2019. Part of the SW360 Portal Project.
+#
+# All rights reserved.   This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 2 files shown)


scripts/migrations/015_migrate_project_field_value.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/python
```
maxhbr: This script needs to be mentioned and explained in the `README.md`

JaideepPalit: Thanks @maxhbr for suggestion.
Have mentioned and explained the script in README.md

maxhbr:
This script looks more like a utility instead of a migration script which compensates database schema changes. Maybe move it to a different folder?

JaideepPalit:
Moved script to a different folder.

PR #740, opened 2019-12-04
feat(REST): Add parameter to GET release by name

JaideepPalit:
Added request parameter to GET release by name

### DIFF
--- a/rest/resource-server/src/docs/asciidoc/releases.adoc
+++ b/rest/resource-server/src/docs/asciidoc/releases.adoc
@@ -49,6 +49,25 @@ include::{snippets}/should_document_get_releases_with_fields/http-response.adoc[
 include::{snippets}/should_document_get_releases_with_fields/links.adoc[]
 
 
+[[resources-releases-list-by-name]]
+==== Listing releases by name
+
+A `GET` request will list all of the service's releases by name. +
+Please set the request parameter `&name=<NAME>`.
+
+===== Response structure
+include::{snippets}/should_document_get_releases_by_name/response-fields.adoc[]
+
+===== Example request
+include::{snippets}/should_document_get_releases_by_name/curl-request.adoc[]
+
+===== Example response
+include::{snippets}/should_document_get_releases_by_name/http-response.adoc[]
+
+===== Links
+include::{snippets}/should_document_get_releases_by_name/links.adoc[]
+
+
 [[resources-releases-get-by-externalids]]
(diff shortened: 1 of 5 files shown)


rest/resource-server/src/docs/asciidoc/releases.adoc
```diff
@@ -49,6 +49,24 @@ include::{snippets}/should_document_get_releases_with_fields/http-response.adoc[
 include::{snippets}/should_document_get_releases_with_fields/links.adoc[]
 
 
+[[resources-releases-list-by-name]]
+==== Listing releases by name
+
+A `GET` request will list all of the service's releases by name.
```
smrutis1: It will improve the readability if we can add the parameter name in description as well.

mcjaeger:
how about returning a 204 which just says, response identically empty:

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

?

rest/resource-server/src/docs/asciidoc/releases.adoc
JaideepPalit: Changes made as suggested.

JaideepPalit:
Changes made as suggested.
Returning Status Code - 204 (No Content) , when response is empty.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/release/ReleaseController.java
```diff
@@ -89,15 +90,25 @@
...
         List<Resource> releaseResources = new ArrayList<>();
         for (Release sw360Release : sw360Releases) {
             Release embeddedRelease = restControllerHelper.convertToEmbeddedRelease(sw360Release, fields);
             Resource<Release> releaseResource = new Resource<>(embeddedRelease);
             releaseResources.add(releaseResource);
         }
-        Resources<Resource> resources = new Resources<>(releaseResources);
 
-        return new ResponseEntity<>(resources, HttpStatus.OK);
+        Resources<Resource> resources = null;
```
smrutis1: Is it possible to make it generic for all the resource like projects, components etc by moving it to `RestControllerHelper.java` and also it would be good to make the response status consistence for all other resource which is returning  response without content.

JaideepPalit: Changes made.

PR #752, opened 2020-01-07
fix(spreadsheet-export): fixed the sequence of values based on headers

### LINKED ISSUE #751: Spreadsheet Export Error
When exporting to an Excel spreadsheet from a Project's Clearing Status tab, the excel sheet column names don't match the project columns. For example Project Mainline State column shows Project Names or Project Mainline State is in the categories column.

smrutis1:
Closes #751


mcjaeger:
bug seems to be solved, however, user roles are not coming (contributors | moderators)

smrutis1:
@mcjaeger  @JaideepPalit Thanks for pointing this out. It should be working now.

PR #753, opened 2020-01-09
fix(ProjectUI): fixed 'Project is temporarily unavailable' issue due to obligation feature

akapti:
Signed-off-by: akapti <abdul.mannankapti@siemens.com>


akapti:
deployed and tested the changes in stage environment, and projects details/edit page is loading as expected.

Issue #762, opened 2020-01-24
:-( missing bracket in viewjsp of component portlet


mcjaeger:
Unfortunately, sw360 7 is faulty:

https://github.com/eclipse/sw360/commit/df54014729a05e977095c187f65ce7f0d22f433f#r36950304

-> one tag character is missing, components list does not render.

maxhbr:
PR is created: https://github.com/eclipse/sw360/pull/763

PR #763, opened 2020-01-24
fix(ui): fix issue #762, introduced in df5401472

maxhbr:
see also:
- https://github.com/eclipse/sw360/commit/df54014729a05e977095c187f65ce7f0d22f433f#commitcomment-36950304
- https://github.com/eclipse/sw360/issues/762

### DIFF
--- a/frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/view.jsp
+++ b/frontend/sw360-portlet/src/main/resources/META-INF/resources/html/components/view.jsp
@@ -63,7 +63,7 @@
     <portlet:param name="<%=PortalConstants.LICENSE_ID%>" value="<%=PortalConstants.FRIENDLY_URL_PLACEHOLDER_ID%>"/>
 </liferay-portlet:renderURL>
 
-div class="container" style="display: none;">
+<div class="container" style="display: none;">
 	<div class="row">
 		<div class="col-3 sidebar">
 			<div class="card-deck">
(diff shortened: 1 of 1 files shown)


mcjaeger:
I do not know why, but it is not loading. Components shows circling partial circle and the console shows this:

Screenshot 2020-01-26 at 22 35 31

(good news that vagrant virtual box works in general, however, checkout of your fork did not work ...)

mcjaeger:
Adding to that: it is not the bug to have zero components - I have forged a request to add a component and created one ... Also with one entry (which is visible in home for example) the components list view is not loading.

mcjaeger:
Hm, I think these two lines need to be in and should not have been deleted. I am very sorry, I did not test this thoroughly ...

https://github.com/eclipse/sw360/commit/df54014729a05e977095c187f65ce7f0d22f433f#r36971909

maxhbr:
The PR is updated, please retest.

mcjaeger:
still not working:

Screenshot 2020-01-27 at 10 14 59

maxhbr:
i have also added the following snippet back in
```html
<script>
    var renderCallback = function () {
    };
    var dataGetter = function(field) {
    };
</script>
```

mcjaeger:
me too (locally), but right now there seems to be test failure again, I am on it and will report if done.

mcjaeger:
looks good, but let me test with a clean checkout

PR #770, opened 2020-02-04
feat(REST): Patch Releases to Project

JaideepPalit:
Added a new request parameter `&patch=<true|false>` in-order to patch releases to existing releases in a Project.


mcjaeger:
please note that the PR description is outdated as the patch is expressed by using the http method instead of the parameter-based approach

smrutis1:
Tested this PR. Linking releases to projects working without overriding the existing one with the patch request. However I think it would be good if the rest users can patch the release relations and mainline state as well from this endpoint while linking it to the project.

Issue #826, opened 2020-04-09
Error while merge of component, release, vendor


JaideepPalit:
While merging Component, Release or Vendor,
Error message pops out , indicating JS error.
Even though table load , Next button is disabled and Abort button is enabled.
Therefore not able to proceed to next step after selecting the Source Component.

![mergeCompUpload](https://user-images.githubusercontent.com/56534710/78870395-acf04580-7a63-11ea-8319-53a295d17087.png)

mcjaeger:
was solved here: https://github.com/eclipse/sw360/pull/827

PR #827, opened 2020-04-09
fix(mergeUI): Provided fix for error message on merge component, release, vendor - choose source page

JaideepPalit:
> Merge of component, release, vendor - `choose source page` , gives JS error which blocks user to proceed to next step
> * Which issue is this pull request belonging to and how is it solving it? (*#826*)

### DIFF
--- a/frontend/sw360-portlet/src/main/resources/META-INF/resources/html/admin/vendors/mergeVendor.jsp
+++ b/frontend/sw360-portlet/src/main/resources/META-INF/resources/html/admin/vendors/mergeVendor.jsp
@@ -166,7 +166,10 @@
                     url: "<liferay-ui:message key="datatables.lang" />"
                 },
             });
-            datatables.enableCheckboxForSelection(table, 0);
+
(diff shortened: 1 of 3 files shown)


frontend/sw360-portlet/src/main/resources/META-INF/resources/html/admin/vendors/mergeVendor.jsp
```diff
@@ -166,7 +166,10 @@
                     url: "<liferay-ui:message key="datatables.lang" />"
                 },
             });
-            datatables.enableCheckboxForSelection(table, 0);
+
+            $("#vendorSourcesTable").on( 'init.dt', function (e, settings ) {
```
smrutis1, line 170: May be the additional parameters `e` and `settings` can be removed as it is not being used currently.

JaideepPalit: Removed additional parameters.

Issue #865, opened 2020-05-27
Feature request: support multi-values of same external id key.


albert-hjy:
### Summary
We need the externalIds supports multi-values for the same external id key.
>e.g. For a release (Name: "Spring Boot", version: "2.1.14")
![image](https://user-images.githubusercontent.com/61676785/83028522-22d85c80-a064-11ea-9c2b-5094d9e732d5.png)

### Use cases
1. Store CPE.
I know that there is a CPE ID field on the release page, but there are cross-platform components that may have multiple CPE IDs, so we are trying to store CPE IDs in externalIds as following.
> Example [component: 7-zip](https://nvd.nist.gov/products/cpe/search/results?keyword=7-zip&status=FINAL&orderBy=CPEURI&namingFormat=2.3)
![image](https://user-images.githubusercontent.com/61676785/83029228-0557c280-a065-11ea-8c92-14a4d9b238d9.png)

2. Use externalIds to group components/releases.
Reference to the example that been used in **Summary**. We also add externalIds of the sub-modules under "Spring Boot" to group sub-modules to a parent component "Spring Boot".

3. This artifact of a component might be moved to another one. We store both the old artifact and the new artifact.
> Here is an [example](https://mvnrepository.com/artifact/org.springframework/spring).
> The artifact of Spring Framework was moved to Spring Core.

### Proposal 

**Option 1** 
- Change the externalIds data structure from `key:value` pair to `key:array` values.
>e.g.
>```
>"org.apache.maven":[
>        "org.springframework.boot:spring-boot-starter-data-redis:2.1.14",
>        "org.springframework.boot:spring-boot-configuration-processor:2.1.14"
>        ]
>```
- No change to UI
- datahandler needs to be changed to support the new data structure.

**Option 2** 
- Don't change the data structure, when there is only one value, store it as `"key":"value"`, when there is more than one value, store them as `"key":"['value1', 'value2']"`, the value is converted from `array['value1', 'value2']`
>e.g.
>```
>"org.apache.maven":"[
>        'org.springframework.boot:spring-boot-starter-data-redis:2.1.14',
>        'org.springframework.boot:spring-boot-configuration-processor:2.1.14'
>        ]"
>```
- No change to UI

I prefer the second option, I found that there are many processes are relying on the `key:value` data structure.

I can implement this if the feature sounds reasonable.🙂

albert-hjy:
Hi @mcjaeger , any suggestions on this feature?🙂

PR #890, opened 2020-06-19
feat(ExternalIds):Support multi-externalId Values of same externalId key in Project, Component, and Release page

albert-hjy:

> * Which issue is this pull request belonging to and how is it solving it? (https://github.com/eclipse/sw360/issues/865)
> * No dependencies updated

Issue: 

**Project**
> 1. UI: Edit Project
image
> 2. UI: Display Project
image
> 3. API: Search by external id
image

**Component**
> 1. UI: Edit Component
image

> 2. UI: Display Component
image

> 3. API: Search by external id
image

**Release**
> 1. UI: Edit Release
image

> 2. UI: Display Release
image

> 3. API: Search by external id
image
image

> No additional test

### Checklist
Must:


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortletUtils.java
```diff
@@ -369,6 +371,27 @@ public static VerificationState getVerificationState(VulnerabilityDTO vul){
...
+        return customMap.entrySet().stream()
+                .collect(Collectors.toMap(Map.Entry::getKey,
+                        e -> {
+                            try {
+                                if(!isNotEmpty(e.getValue()) && e.getValue().size() == 1) {
+                                    return e.getValue().stream()
+                                            .findFirst()
+                                            .orElse("");
+                                }
+                                ObjectMapper mapper = new ObjectMapper();
```
oheger-bosch, line 385: Creating an object mapper is a rather expensive operation; so this should not be done for each element in the map. The object creation should at least be moved to the beginning of the method.

As the ObjectMapper class is thread-safe, in this case it would also be an option to have a static ObjectMapper field in PortletUtils that is always reused.

albert-hjy: Thanks a lot, @oheger-bosch 
I have refactored the code.

oheger-bosch: @albert-hjy, thanks for the update. This looks better now performance-wise. I am, however, a bit worried that the lazy initialization approach is not thread-safe. As the utils class is called during request execution, there will be multiple threads involved, and we should take care that this is safe.

The easiest solution is certainly to get rid off the lazy initialization, make the ObjectMapper field *final*, and create the mapper directly.

If you prefer lazy initialization, then I would suggest using the [Lazy initialization holder class idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom) for static fields.

Please also refer to J. Bloch, "Effective Java", 3rd edition, item 83 "Use lazy initialization judiciously"

albert-hjy: @oheger-bosch , thanks a lot. Agree that it's unnecessary to use lazy initialization here. I have changed ObjectMapper field _final_.

albert-hjy:
Thank you @JaideepPalit , 
I have

- added `EscapeXml` to `DisplayMapOfExternalIdsString.java`
- added migration scripts for updating CouchDB Project, Component, and Release `byExternalId` view
- documented the two different types of value `String` and `Array of String` to Project, Component, and Release APIs
- fixed `update existing project or Component or Release without any change in External IDs. ExternalId value will be converted from String to Array of String`

scripts/migrations/020_update_byExternalIds_release_view.py
```diff
@@ -0,0 +1,97 @@
...
+# dedicated framework for automatic migration will be written in the
+# future. When that happens, this script should be refactored to conform
+# to the framework's prerequisites to be run by the framework. For
+# example, server address and db name should be parameterized, the code
+# reorganized into a single class or function, etc.
+# -----------------------------------------------------------------------------
+
+import time
+import couchdb
+import jsonß
```
JaideepPalit, line 21: Getting error while executing this script

```
File "020_update_byExternalIds_release_view.py", line 21
SyntaxError: Non-ASCII character '\xc3' in file 020_update_byExternalIds_release_view.py on line 21, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

albert-hjy:
Thank you @JaideepPalit 
- There are some error occurring while running migration script scripts/migrations/020_update_byExternalIds_release_view.py due to presence of foreign character
> This was a typo when I change the CouchDB from our remote testing server to localhost. Fixed it.
- Kindly provide same read/write/execute access for newly added migration script
> Added the same permission to new migration scripts.
image

mcjaeger:
hi, just tested it on updating the PR over an installation which had current master :
* install ok
* multiple values for same id works and stored and displayed properly

However, executing the migration scripts brings (from migration log):
```
{
    "old_value": "function(doc) {  if (doc.type == 'component') {    for (var externalId in doc.externalIds) {      try {            var values = JSON.parse(doc.externalIds[externalId]);            for (
var idx in values) {              emit( [externalId, values[idx]], doc._id);            }      } catch(error) {          emit( [externalId, doc.externalIds[externalId]], doc._id);      }    }  }}", 
    "field_to_update": "views->byExternalIds->map", 
    "docId": "_design/Component", 
    "new_value": "function(doc) {  if (doc.type == 'component') {    for (var externalId in doc.externalIds) {      try {            var values = JSON.parse(doc.externalIds[externalId]);            for (
var idx in values) {              emit( [externalId, values[idx]], doc._id);            }      } catch(error) {          emit( [externalId, doc.externalIds[externalId]], doc._id);      }    }  }}"
}
```

-> I am not sure if that is serious, because the records had external ids before and everything seems to work ...

mcjaeger:
wow, even search works for newly added multiple values from the same key ...

mcjaeger:
@JaideepPalit I think it works. if it looks OK to you, could you pls. approve and merge?

PR #899, opened 2020-06-23
Oheger bosch/feat/attachments delete

### LINKED ISSUE #897: Support deleting attachments via REST API
While the Thrift API supports deleting attachments, this functionality is not yet exposed via the REST API. It would be good to have endpoints analogously to the attachment upload and download endpoints, e.g. 
`DELETE .../releases/{releaseId}/attachments/{attachmentIds}`

As was discussed during the weekly SW360dev tech call, a pre-requisite for the deletion of an attachment is that it must not be used by a project.

oheger-bosch:
Fixes #897  

This PR adds the ability to delete attachments from components and releases.

The functionality is implemented by the controllers for components and releases with a helper function for checking whether an attachment can be deleted in the Sw360AttachmentService. (According to discussions in the SW360 tech call, attachments must not be deleted when they are used by a project; this is checked by evaluating the attachment usage relations.)

To test the new method of the attachment service, I added a unit test for the service layer. In order to allow injecting mock thrift services, a new service provider interface was introduced, which is basically a factory for thrift services. The factory can be injected by Spring. (As thrift client services are not thread-safe, new instances must be created for each request; so it is not possible to inject the services themselves.)

The new REST endpoints have been added to the documentation.

### How To Test?
Test cases for the new functionality have been added for the controllers and the attachment service.

### DIFF
--- a/rest/resource-server/src/docs/asciidoc/components.adoc
+++ b/rest/resource-server/src/docs/asciidoc/components.adoc
@@ -193,6 +193,28 @@ include::{snippets}/should_document_get_component_attachment/curl-request.adoc[]
 ===== Example response
 include::{snippets}/should_document_get_component_attachment/http-response.adoc[]
 
+[[resources-component-attachment-delete]]
+==== Delete attachments
+
+With a `DELETE` request one or multiple attachments can be deleted from a component.
+To delete multiple attachments at once, just specify a comma-separated list of
+attachment IDs.
+
+Note that attachments can only be deleted if they are not used by a project.
+Requests that cannot delete any of the attachments specified fail with response
+status 500.
+
+===== Example request
+include::{snippets}/should_document_delete_component_attachment/curl-request.adoc[]
+
+===== Example response
+include::{snippets}/should_document_delete_component_attachment/http-response.adoc[]
+
+The response contains the updated component. Here the set of attachments is especially
+relevant. If multiple attachments were to be deleted, but the deletion failed for
+some of them, those can still be found in the set. So callers should inspect
+the set of attachments in the result to find out whether all delete operations
+have been successful.
 
 [[resources-components-create]]
 ==== Creating a component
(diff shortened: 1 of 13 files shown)


mcjaeger:
travis is actually OK 

```
[INFO] ------------------------------------------------------------------------
5204[INFO] BUILD SUCCESS
5205[INFO] ------------------------------------------------------------------------
5206[INFO] Total time: 08:26 min
5207[INFO] Finished at: 2020-06-23T09:30:18Z
5208[INFO] Final Memory: 197M/747M
5209[INFO] ------------------------------------------------------------------------
```

oheger-bosch:
@smrutis1 What is meant by 'accepted attachment'? Is this an attachment used by a project?

Regarding error handling I was not sure. As the endpoint returns an updated entity, it was not possible to use a multi-status response.  So my approach was to return a success response if some of the attachments could be deleted, and based on the updated entity the caller could find out which deletes were successful and which are not. Only if all delete operations fail, an error status is returned.

But we can discuss alternatives you find more appropriate.

smrutis1:
> @smrutis1 What is meant by 'accepted attachment'? Is this an attachment used by a project?
   
   @oheger-bosch, Sorry for the confusion. while uploading the attachment in UI there is an option to choose the approval status as shown below. 
image

> Regarding error handling I was not sure. As the endpoint returns an updated entity, it was not possible to use a multi-status response. So my approach was to return a success response if some of the attachments could be deleted, and based on the updated entity the caller could find out which deletes were successful and which are not. Only if all delete operations fail, an error status is returned.

Sounds okay to me, as the same is well documented.

oheger-bosch:
@smrutis1, thanks for the clarification. So this field corresponds to the checkStatus property of the attachment. I will add a check that requests to delete attachments with the accepted state are rejected as well.

oheger-bosch:
Sw360AttachmentService.filterAttachmentsToRemove() now evaluates the checked status of attachments.

Issue #909, opened 2020-06-29
User can assign himself to a project to get extended rights


AlexFTrapp:
Hi,

I just came accross a possible security-risk in sw360. A User can assing himself write access without approval of the project owner or moderators.
Steps to reproduce
- Use Account that has a normal "user"-Role
- Change a foreign Project that is open to the group of the user, but the user has no write-access to it
- create moderation request: **add user himself as a role that is not yet filled. e.g. lead architect**
- Save Moderation restest to get extended rights for this project although the moderation request is not yet approved
- From now on the user can edit the original document

I don't know whether this is a feature or a bug. But considering there is such a thing as "moderation requests", I don't think the above is a desired behaviour.

Greets Alex

mcjaeger:
I think the PR was merged, so current master fixes this bug.

AlexFTrapp:
Merged fix works fine for me and my scenario.

PR #910, opened 2020-06-29
fix(ProjectModeration): Fixed isWriteActionAllowedOnProject check for project update

JaideepPalit:
> Fixed isWriteActionAllowedOnProject check for project update
> * Fixed incorrect value for Visibility in Edit Project view which has existing moderation request

> * Which issue is this pull request belonging to and how is it solving it? (*#909*)

> - To test - `incorrect value for Visibility in Edit Project view which has existing moderation request`
    Create a project with visibility as `Everyone` . Create moderation request without making any changes to visiblity.
    On Editing the project. The Project with moderation request changes should load , and visibility should be correct `Everyone`.

### Checklist
Must:


mcjaeger:
I am not sure how the visibility change in the sources are allright. Here is the test:

1. user@sw360.org creates a project with visibility `everyone`
2.  user@sw360.org logs out
3.  user2@sw360.org logs in
4.  user2@sw360.org edit the new project and changes moderators: adding own user
5. submitting the change
6. log out ( user2@sw360.org can check to write the project before, it is safe -> good)
7. user@sw360.org logs in again
8. user@sw360.org accepts the moderation request
ß. project visibility is change to `group and moderators`

JaideepPalit:
Thanks @mcjaeger for the test.
I have provided the fix for the same.

PR #939, opened 2020-08-04
chore(script): Added a script to uninstall the current thrift version.

oheger-bosch:
The script checks whether thrift is currently installed on the local
machine. If so, the thrift version is determined, and this version is
uninstalled.

This is useful when the thrift version is updated. Then the old
version needs to be uninstalled first before the new one can be
installed.

The install-thrift.sh script has been extended to print a message
referring to the new script if an incorrect version is detected.

### DIFF
--- a/scripts/install-thrift.sh
+++ b/scripts/install-thrift.sh
@@ -18,23 +18,116 @@
 # -----------------------------------------------------------------------------
 
 set -e
-
-VERSION=0.13.0
+CURRENT_THRIFT_VERSION=0.13.0
 
 BASEDIR="/tmp"
-BUILDDIR="${BASEDIR}/thrift-$VERSION"
+CLEANUP=true
+UNINSTALL=false
 
 has() { type "$1" &> /dev/null; }
 
-if has "thrift"; then
-    if thrift --version | grep -q "$VERSION"; then
-        echo "thrift is already installed at $(which thrift)"
-        exit 0
-    else
-        echo "thrift is already installed but does not have the correct version: $VERSION"
-        exit 1
-    fi
-fi
+processThrift() {
+  set -x
+  VERSION=$3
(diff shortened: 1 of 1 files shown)


oheger-bosch:
Yes, I am going to change this. As has already been discussed in the SW360 tech talk, a command line option will be added to the install script which triggers the un-install.

mcjaeger:
observation, after the uninstall (on u1604) some folders seems to remain:

```
mjaeger@ip-10-142-7-237:~/install-thrift$ thrift --version
-bash: /usr/local/bin/thrift: No such file or directory
mjaeger@ip-10-142-7-237:~/install-thrift$ sudo find / -name "thrift"
/usr/local/share/doc/thrift
/usr/local/share/doc/thrift/java/org/apache/thrift
/home/siemagrant/.m2/repository/org/apache/thrift

```

scripts/install-thrift.sh
```diff
@@ -18,23 +18,116 @@
...
+                      --without-python --without-erlang --without-perl --without-c_sharp --without-d --without-php \
+                      --without-go --without-lua --without-nodejs --without-cl
+      fi
+      if [ "$1" == true ]; then
+        make
+      fi
+  fi
+
+  echo "-[shell provisioning] Executing make $2 on thrift"
+  $SUDO_CMD make $2
```
mcjaeger, line 86: actually this seems to be a good way of doing the uninstall here (passing it as parameter

mcjaeger:
but result for installing it seems to be OK
```
mjaeger@ip-10-142-7-237:~/install-thrift$ thrift --version
Thrift version 0.13.0
```

mcjaeger (approved):
actually ledaving man pages deos not seem to a big issue ... works both uninstalling old version and adding new versions ...

PR #1009, opened 2020-10-08
fix(obligations):cover null pointer case if file with obligations is …

mcjaeger:
…missing

Signed-off-by: Michael C. Jaeger <michael.c.jaeger@siemens.com>

This small change covers a null pointer exception occuring if a license description file does not contain obligations while building the project view. otherwise a map built from a stream is not built up properly.

Issue: described here. 

### Suggest Reviewer

@JaideepPalit as the line was his proposal...

### How To Test?

tested on productive instance and this data set.


JaideepPalit (approved):
Code looks good

Issue #1031, opened 2020-11-03
Filter Option "Project Clearing State Open" for main project list view


JaideepPalit:
Add a filter setting for "Project clearing state", maybe a pop-up menu for the two states.

JaideepPalit:
Closing since PR is merged

PR #1032, opened 2020-11-03
feat(ProjectListUI): Added clearing state filter in Project List

JaideepPalit:
> Added clearing state filter in Project List
> * Which issue is this pull request belonging to and how is it solving it? (*#1031*)

FilterClearingState

### Checklist
Must:

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectSearchHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectSearchHandler.java
@@ -64,6 +64,9 @@ public class ProjectSearchHandler {
                     "    if(doc.state !== undefined && doc.state != null && doc.state.length >0) {  "+
                     "      ret.add(doc.state, {\"field\": \"state\"} );" +
                     "    }" +
+                    "    if(doc.clearingState) {  "+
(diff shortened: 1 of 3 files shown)


backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectSearchHandler.java
```diff
@@ -64,6 +64,9 @@
                     "    if(doc.state !== undefined && doc.state != null && doc.state.length >0) {  "+
                     "      ret.add(doc.state, {\"field\": \"state\"} );" +
                     "    }" +
+                    "    if(doc.clearingState !== undefined && doc.clearingState != null && doc.clearingState.length >0) {  "+
```
smrutis1, line 67: May be It can simply be written as `if(doc.clearingState)`

smrutis1 (changes requested):
Observed an issue while testing,

Searching with some entry in the clearing state filter field works fine. However after hitting the search button post resetting the filed value to blank/no-content, then there is no result. Ideally it should show the list of projects.

JaideepPalit:
Thanks for review and test.
Made changes and fixed the issue

smrutis1 (approved):

smrutis1:
Tested successfully

PR #1062, opened 2020-12-04
feat(sw360Docker): Single container Docker for SW360

JaideepPalit:
> Single container Docker for SW360.

### Checklist
Must:


blaumeiser-at-bosch (commented):
Works on my machine, but I would propose to summarize the sequential run commands, since every command creates a new image layer which is quite a waste of resources. That is only a hint, I can live with the state as it is though

mcjaeger:
it worked for me as well ...

Issue #1078, opened 2020-12-16
Create/Update Release with name same as component name


JaideepPalit:
It is possible to create Release with different name than its component via rest.

JaideepPalit:
Closing since PR is merged

PR #1079, opened 2020-12-16
fix(rest): Create/Update Release with name same as component name

JaideepPalit:
> Create/Update Release with name same as component name
> * Which issue is this pull request belonging to and how is it solving it? (*#1078*)

### Checklist
Must:


smrutis1 (changes requested):
Two remarks:
* Error message should contain also the version string along with the name.
* As there is no significant of passing `name` while creating release from REST anymore. Could it be removed from the REST docs as well.

JaideepPalit:
Thanks for review. 

> * Error message should contain also the version string along with the name.
      - fixed

> * As there is no significant of passing `name` while creating release from REST anymore. Could it be removed from the REST docs as well.
      - removed

smrutis1:
One minor comment:

In documentation, created on filed is appearing where as it's actually not coming in the response. Could it be removed?

image

image

JaideepPalit:
> In documentation, created on filed is appearing where as it's actually not coming in the response. Could it be removed?  
    - removed

smrutis1 (approved):
Code looks fine to me. Tested it successfully as well.

PR #1091, opened 2021-01-12
chore(eclipse): Publish SW360 artifacts in Eclipse infrastructure

blaumeiser-at-bosch:
Add the necessary infrastructure for publishing SW360 artifacts in the Eclipse Maven Repository.

Contains:

- Dockerfile for a build container published to Docker Hub, that is used in the Eclipse environment for the build. The container has to be published manually, a script for doing that and a readme for the container is provided additionally
- Jenkinsfile that is used by the EF Jenkins infrastructure

Issue: #981

Issue is not completely solved, since this pr only adds snapshot releases, real releases and milestones will be added with an additional change.

### Suggest Reviewer
n/a

### How To Test?
Test is to run the SW360 build on our Jenkins server https://ci.eclipse.org/sw360/

### Checklist
Must:
- [x] All related issues are referenced in commit messages and in PR

PR #1112, opened 2021-02-10
chore(eclipse): Enable release and milestone builds

### LINKED ISSUE #981: Release SW360 in Maven
SW360 should have a release in Maven

Todo:
- [x] make a weekly release in the eclipse maven
    - [x] create jenkins job with jenkinsfile
- [x] clear process

blaumeiser-at-bosch:
fixes #981 

Insert mechanism to run either a commit count or a release build in the Eclipse CI Jenkinsfile. A boolean parameter allows to state that the build is a release build, the default case is a commit count build. The artifacts are deployed to the Eclipse Nexus. Builds are all release builds, so the build break during deployment, if a version is created for a second time. This is the standard behavior for a release build in the maven sense.

Issue: #981  

### Suggest Reviewer
@mcjaeger 

### How To Test?
Run the build and check the artifacts being deployed one Eclipse Nexus

### Checklist
Must:

### DIFF
--- a/jenkins.eclipse/Jenkinsfile
+++ b/jenkins.eclipse/Jenkinsfile
@@ -61,18 +61,38 @@ spec:
         }
     }
 
+    environment {
+        COUNT = sh(
+                script: "printf \$(git rev-list `git rev-list --tags --no-walk --max-count=1`..HEAD --count)",
+                returnStdout: true
+        )
+    }
+
+    parameters {
+        booleanParam(
+            name: 'RELEASE',
+            defaultValue: false,
+            description: 'If true, it is a milestone build with a m1 version suffix, or, if false, an intermediate build with a version suffix based on the commit count')
+    }
     stages {
         stage('Deploy SW360') {
             steps {
                 container('sw360buildenv') {
+                    script {
+                        if (params.RELEASE) {
+                            VERSION_SUFFIX = "0-m1"
+                        } else {
+                            VERSION_SUFFIX = "0-" + COUNT
+                        }
+                    }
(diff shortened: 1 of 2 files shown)


blaumeiser-at-bosch:
@mcjaeger : As soon as PR https://github.com/eclipse/sw360/pull/1091 is merged, I will set this PR ready. It builds either a version as discussed on Wednesday (based on the commit count) or a release version in the form x.y.z-m1.

I also updated the minor version in the ci profile. We have to change the procedure here with the deploy builds, because all versions build after a tag need to be higher than the tagged version, currently it seems that you only update the version prior to the release, this means that all intermediate versions are technically newer then the tag, but semantically older due to the version compare mechanism.

When this PR is through, I will install the final build that builds on a weekly basis and also allows to build the release. The current setup requires to run the build after the tagging, but it will be a small change to allow a build to define a certain tag for building.

blaumeiser-at-bosch:
Additionally, after this PR is merged, I will remove all existing builds from the Nexus release repository. I will also have the snapshot repository removed, since from our discussion, we only want to use the maven release mechanism to publish intermediate and milestone releases.

By the way, I have clarified with Eclipse Foundation PMO, the requirement concerning signing is only for stuff that can be signed, other stuff can be published without signing it. This is the case for the war files we generate, since Eclipse Infrastructure cannot sign war files in the current state.

Related issues in Eclipse Foundation:
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=570917
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=571028

PR #1138, opened 2021-03-15
fix(ReleaseUI): fix Spdx parser for License withh + sign

### LINKED ISSUE #1129: 'Show License Info' button doesn't work with spdx file export from FOSSology when there is a '+' in license name
RHEL 8, SW360 12.0
The spdx file exported by fossology will replace '+' with '%2B'.
The 'Show License Info' in Components - Releases - Clearing Details couldn't work fine with it, even the main license's name doesn't have a '+'.
The problem will make other spdx attachments which don't have '%2B' in it broken by chance. Unless user delete all spdx files with '%2B' the problem won't be solved.

akapti:

> Which issue is this pull request belonging to and how is it solving it?
* #1129 

> Did you add or update any new dependencies that are required for your change?
* No

Issue: closes #1129 

* `Show License Info` button in `Clearing Details` tab under `Release Details Page` should display the license with `+` sign.
* Should display `Other Licenses` as well. (from `spdx:licenseInfoInFile` tag)

* No

### Checklist
Must:

### DIFF
--- a/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParserTools.java
+++ b/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParserTools.java
@@ -18,6 +18,11 @@ import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentContent;
 import org.eclipse.sw360.datahandler.thrift.licenseinfo.*;
 import org.apache.jena.ext.xerces.util.XMLChar;
 
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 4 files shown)


backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParserTools.java
```diff
@@ -117,6 +122,20 @@ private static String getLocalName(String label) {
...
             // label contain invalid NCName
             if (!XMLChar.isNCName(label.charAt(i))) {
+                if (Util.splitNamespaceXML(label) == label.length()) {
+                    try {
+                        label = URLDecoder.decode(label, StandardCharsets.UTF_8.name());
+                        URL url = new URL(label);
+                        if (null != url.getRef()) {
+                            return url.getRef();
+                        } else {
+                            String path = url.getPath();
```
ravi110336, line 132: We can do a null check for the path variable as  well similar we did for the Ref variable, just in case to avoid any null pointer.

ravi110336 (commented):
Tested and working fine one small review comment. Please check.
image

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParserTools.java
akapti: For a valid URL, path cannot be null (can be empty string), So we don't need Null check here unlike ref which can be null.

ravi110336: Thanks for the clarification.

PR #1150, opened 2021-03-17
fix(build): Fix issue with overwriting of patchlevel variable in pom.xml

### LINKED ISSUE #1144: Bug: The mechanism used in the ci profile in the main pom.xml does not work with the maven flatten module
### Description

The ci profile in the main pom.xml replaces the *patchlevel* environment variable with a minor version and the content of a *count* environment variable. This results in the variable *patchlevel* been defined twice. The variable itself is then used to define the *revision* environment variable which is used as version of the pom.xml file.

The problem with this mechanism is that redefining the same environment variable works well with the maven build, so the version is set properly. But the maven flatten plugin, when creating the flattened pom, does not see the reset value but used the default resulting in a flattened pom with a wrong version.

During mvn deploy, the pom with the wrong version is used for uploading it, e.g., to a Nexus repository, resulting in a corrupt repository which does not allow to download the dependency in a maven/gradle based build.

#### How to reproduce

Run a maven build with -Pci enabled and -Dcount=<something>. Observe that the version number of the created jar files reflect the generated version with <something> as postfix. Check the .flattened-pom.xml in the main folder of a maven component and observe that the version given in the file ends with SNAPSHOT as defined in the <properties> section of the main pom.xml

### Versions

* Last commit id on master: https://github.com/eclipse/sw360/commit/ce69b3bd7537d4601d4e504f83fa42332adcf65b
* Operating System (lsb_release -a): The build container for the Eclipse Infrastructure

blaumeiser-at-bosch:
Issue is that the maven flatten module is not able to handle the
double definition of variable patchlevel in the pom correctly.
The generated pom files have the wrong version.
Solution: Put the minor version in the revision variable and
reduce the standard patchlevel content to the idea from the "ci"
profile. Remove the "ci" profile and adapt the travis and jenkins
build files.

Removed the version handling from travis build because it was obsolete.
The whole mechanism was only used to define the name of the build, but not
the version of the mvn call. Since the variable did not make sense, I removed 
the whole mechanism.

Signed-off-by: Lars Geyer-Blaumeiser <lars.geyer-blaumeiser@bosch.io>

Fixes #1144 

### Suggest Reviewer
@mcjaeger 

### How To Test?
Check Travis and Eclipse build log for the right version to be build

### Checklist
Must:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ before_install:
   - ./scripts/install-thrift.sh --no-cleanup
 
 env:
-  GIT_COMMIT_COUNT="$(git rev-list  `git rev-list --tags --no-walk --max-count=1`..HEAD --count)"
   MVN_ARGS="package"
 install:
   mvn dependency:resolve || true
@@ -45,9 +44,9 @@ script:
 
 matrix:
   include:
-    - name: mvn package -Pci -Dcount=$GIT_COMMIT_COUNT
+    - name: mvn package
       env: MVN_ARGS="package"
-    - name: mvn validate -Pci -Dcount=$GIT_COMMIT_COUNT
+    - name: mvn validate
       env: MVN_ARGS="validate"
       before_install:
     - name: mvn dependency:analyze
--- a/jenkins.eclipse/Jenkinsfile
+++ b/jenkins.eclipse/Jenkinsfile
@@ -80,9 +80,9 @@ spec:
                 container('sw360buildenv') {
                     script {
                         if (params.RELEASE) {
-                            VERSION_SUFFIX = "1.0-m1"
+                            VERSION_SUFFIX = "0-m1"
                         } else {
-                            VERSION_SUFFIX = "1.0-" + COUNT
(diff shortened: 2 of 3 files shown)


blaumeiser-at-bosch:
@mcjaeger: This fixes the issue, we discussed last Wednesday. One thing, in the travis build the usage of the commit count mechanism was pure fake, the build always builds version x.1.0-SNAPSHOT because it detects the commit count but only uses the created variable to parameterize the name of the build step. Even that does not work.

It does not use the information in the maven call of the corresponding build step. I removed the mechanism since I don't think that the version built by Travis has any reason. The build is only there to prove that a PR does not add any regression.

So from my side, the PR is done and can be reviewed, it has no impact on the travis build, I only cleaned up the file as described above. The rest is not visible in the standard usage and the patchlevel variable is now defined in a way that it replaces the count variable formerly used in the ci profile which I have removed. It never worked properly and the functionality has been made obsolete.

blaumeiser-at-bosch:
@mcjaeger : Rebased the PR, we should be able to merge it now.

Issue #1003, opened 2020-10-01
updateRelease REST API return 500 on moderation request


mcjaeger:
There is the update method here:

https://github.com/eclipse/sw360/blob/07f1c080c6b6863bce440cb77f4d02eb3a65cdb1/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/release/Sw360ReleaseService.java#L151

turned out this is returning a 500 error even a moderation request could be created. Should be a different return info or maybe 500, but telling the user that a moderation request is there.

mcjaeger:
(from sw360telco): 
1. recommendation is to return 202 http code (https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)
1.a 202 over 200 would have the advantage to let the client understand by http code that a moderation request has been created
2. response shall contain json about the moderation request (maybe HAL style)
3. since this will be the next thing: offer api to access and approve moderation requests

PR #1160, opened 2021-03-25
fix(rest): Fix status code when moderation request is created as a part of an API call

JaideepPalit:
> Fix status code when moderation request is created as a part of an API call
>  * Affected APIs in ProjectController - Update Project, Add/Patch Releases to a project, Patch vulnerabilities of a project, Add Attachment To Project
>  * Affected APIs in ReleaseController - Update Releases, Delete Attachments from releases
>  * Affected APIs in ComponentController - Delete Attachments from component, Delete Components, Update Components, Add Attachment To Component

> * Which issue is this pull request belonging to and how is it solving it? (*#1003*)

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
@@ -51,6 +51,9 @@ import org.springframework.util.MultiValueMap;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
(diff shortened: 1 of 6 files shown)


ravi110336:
For update components API, the response status is perfectly fine with status code 202, but the response element is showing the updated element value but when we go and validate it in sw360 it didn't updated for user token which is correct instead it should create moderation request.

I think in these scenarios for status 202, we should have one generic response message like "Moderation Request is created."

image

JaideepPalit:
Thanks for review and test. 
Added proper message in response body when moderation request is created.

ravi110336:
Thanks Jaideep. Tested and working fine now, we can merge this.

image
image

ravi110336 (approved):
Code looks good, kindly merge.

PR #1220, opened 2021-05-18
fix(ui/rest): Issue fetching releases by external ids from REST and null value in external id breaks the release view

smrutis1:

Issue: 

500 Error fetching releases by external id with value from rest. If the release contains any external id with value null, then it breaks the release view.

### How To Test?
* Need to test the endpoint `/api/releases/searchByExternalIds?externalId=XXX`
* Put null as external Id value in release and verify if the release should open without any error.

### Checklist
Must:
Signed-off-by: Smruti Prakash Sahoo <smruti.sahoo@siemens.com>


JaideepPalit (approved):
Code looks fine to me. Tested successfully.
Note -
> Put null as external Id value in release and verify if the release should open without any error.
- `null` here means NULL_STRING

PR #1236, opened 2021-06-01
fix(rest): Auto-set release clearing state

smrutis1:

Issue: 

Release clearing state should be automatically set based on attachments and it's type in release, which is working fine in UI, however from REST it can be set manually.

This change is to ignore clearing state value while creating or updating release via REST, which is handled automatically in backend([here](https://github.com/eclipse/sw360/blob/8f7caf8aa02a0b7a28dcb23863eb52116466da4b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java#L923) and [here](https://github.com/eclipse/sw360/blob/8f7caf8aa02a0b7a28dcb23863eb52116466da4b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java#L422))

- Need to verify creating and updating release by setting `clearingState` field in request payload. It should ignore the clearing state value set in the REST request. 

### Checklist
Must:

Signed-off-by: Smruti Prakash Sahoo <smruti.sahoo@siemens.com>

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -393,9 +393,9 @@ public class RestControllerHelper<T> {
     }
 
     public Release updateRelease(Release releaseToUpdate, Release requestBodyRelease) {
-        for(Release._Fields field:Release._Fields.values()) {
+        for (Release._Fields field : Release._Fields.values()) {
             Object fieldValue = requestBodyRelease.getFieldValue(field);
-            if(fieldValue != null) {
+            if (fieldValue != null) {
                 releaseToUpdate.setFieldValue(field, fieldValue);
             }
         }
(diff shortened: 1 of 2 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -393,9 +393,12 @@ public Component updateComponent(Component componentToUpdate, Component requestB
     }
 
     public Release updateRelease(Release releaseToUpdate, Release requestBodyRelease) {
-        for(Release._Fields field:Release._Fields.values()) {
+        for (Release._Fields field : Release._Fields.values()) {
+            if (Release._Fields.CLEARING_STATE.equals(field)) {
```
akapti, line 397: `Release` document have 30+ fields, So rather than adding an `if` condition that's being checked for each field in for loop, I would suggest to reset the `clearingState` field in controller class.. something like this: `updateRelease.setClearingState(sw360Release.getClearingState());`

akapti (changes requested):
Tested the PR, and it's working as expected.

suggested code changes as part of code review.

smrutis1:
@akapti , Thanks for having a look. Have incorporated the changes

akapti (approved):
code looks good.

PR #1473, opened 2022-02-25
fix(modReq): Fixed moderation request for release with version overwrite

### LINKED ISSUE #1472: Component Moderation Request overwrites the data which should not be changed
(from sw360 user)

### Description

In my case I had a moderation request for a component release They wanted to add two file attachements for the component release.

I checked the data and I have seen that the version of the component release was wrong. The component release version was "0.5.1" but the URL, the source and all other stuff had "5.1"

So I have corrected the component releae version to "5.1" before I accepted the moderation request.

1 minute later I have accepted the moderation request to add the file attachements.

What happend is that the release version was overwritten with the old version "0.5.1"

conclusion: A SW360 Moderation request should only change the expected change and not all the old data set of a component.

#### How to reproduce

I hope you can easy reproduce it when you create a moderation request (e.g. description change) and you change the version for a component release afterwards and then you accept the moderation request.

I would expect the same behaviour.[](https://github.com/eclipse/sw360/issues/new?assignees=&labels=bug&template=bug-report.md#how-to-reproduce)Description
In my case I had a moderation request for a component release
They wanted to add two file attachements for the component release.
I checked the data and I have seen that the version of the component release was wrong.
The component release version was "0.5.1" but the URL, the source and all other stuff had "5.1"
So I have corrected the component releae version to "5.1" before I accepted the moderation request.
1 minute later I have accepted the moderation request to add the file attachements.
What happend is that the release version was overwritten with the old version "0.5.1"
conclusion:
A SW360 Moderation request should only change the expected change and not all the old data set of a component.

How to reproduce
I hope you can easy reproduce it when you create a moderation request (e.g. description change) and you change the version for a component release afterwards and then you accept the moderation request.
I would expect the same behaviour.

JaideepPalit:
> Fixed moderation request for release with version overwrite.
> * Which issue is this pull request belonging to and how is it solving it? (*#1472*)

Issue: closes #1472 

### Checklist
Must:

### DIFF
--- a/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ReleaseModerationRequestGenerator.java
+++ b/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ReleaseModerationRequestGenerator.java
@@ -24,6 +24,8 @@ import static com.google.common.base.Strings.isNullOrEmpty;
  */
 public class ReleaseModerationRequestGenerator extends ModerationRequestGenerator<Release._Fields, Release> {
 
+    private static final String DUMMY_VALUE = "Dummy_Value";
+
     @Override
     public ModerationRequest setAdditionsAndDeletions(ModerationRequest request, Release updateRelease, Release actualRelease){
         updateDocument = updateRelease;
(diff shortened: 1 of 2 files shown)


ag4ums:
Working as expected

akapti (approved):
Code Looks good to me. Approved the changes.
Tested by @ag4ums

PR #1474, opened 2022-03-03
fix(merge): Optimized code to check for write permission of release and components before starting to merge

JaideepPalit:
> Optimized code to check for write permission of release and components before starting to merge.

> As per changelogs, It is observed that in some cases, after merge of component . 
> The corresponding releases are not updated with new component Id.
> Thereby creating orphaned Release.

> Maybe Merge is failing in the middle due to permission issue of Release
> As a solution, check has been added for WRITE permission of all releases before actual merge is started.

Issue: 

### Checklist
Must:


akapti (approved):
Code looks good to me.
Tested the changes, and working as expected.

PR #1492, opened 2022-03-29
typo in the docker run command

alpianon:
Just a typo in the documentation, the `docker run` command option `--name` doesn't have the double hyphen

### DIFF
--- a/README_DOCKER.md
+++ b/README_DOCKER.md
@@ -107,7 +107,7 @@ docker exec -it sw360_postgresdb_1 createdb -U liferay -W fossology
 docker run \
     --network sw360net \
     -p 8081:80 \
-    -name fossology \
+    --name fossology \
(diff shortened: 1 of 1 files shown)


mcjaeger:
@alpianon please consider signing the Eclipse Contributor License Agreement and use the same e-mail as for your git commit. It is a bit strange for this single character commit, I agree, but in this way you re generally set.

alpianon:
> @alpianon many thanks! please consider signing the Eclipse Contributor License Agreement and use the same e-mail as for your git commit. It is a bit strange for this single character commit, I agree, but in this way you re generally set. more info here: https://wiki.eclipse.org/Development_Resources/Handling_Git_Contributions

I already signed it, but with another email, so I need to figure out how to add another email

alpianon:
@mcjaeger I had to temporarily change the email address in my EF account, re-sign the ECA, and relaunch the check, now it passed it

Issue #1589, opened 2022-07-12
User Management: Ability to filter for active users


rudra-superrr:
### Description

There should be a column under user's table in admin tab which will tell the status of the users.

PR #1590, opened 2022-07-12
feat(ui):filter active users

rudra-superrr:
Issue:  User Management : Ability to filter active users
#1589 

### How To Test?

- Go to admin tab.
- Then select user, you will see user table with an additional column (Active status).
- This column tells whether user is active or inactive.

users


smrutis1 (approved):
Code looks good

ag4ums (approved):
working as expected

PR #1601, opened 2022-07-27
upd(test): Update hamcrest and get rid of deprecated entries

heliocastro:
Signed-off-by: Helio Chissini de Castro <heliocastro@gmail.com>


smrutis1:
May be we should consider to update rest of the test classes as well.?

ag4ums:
Closing this as there will be new PR for the same

PR #1621, opened 2022-08-24
fix(deps): Fixed wrong dependency download

heliocastro:
Guava dependency was pointing out to wrong directory and not providing
any error on output, so creating an invalid file that failis only on
unpacking level.
If this kept used, in future maven should do the download automatically
instead of a script.

This provided possibility to alter Dockerfile deps download position,
and avoid recreate the build layer all the time.

In addition, a dockerignore file is added to avoid copy deploy / target
/ buiid massive outputs if local build is used to build docker too.

Signed-off-by: Helio Chissini de Castro <heliocastro@gmail.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+# Ignore base files on copy
+deploy?
+*/*/build
+*/*/target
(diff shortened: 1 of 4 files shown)


KoukiHama (approved):

KoukiHama:
I was just now trying to fix it, too. Thanks for fixing it.

PR #1633, opened 2022-09-07
Java 17 preparation

heliocastro:
Prepare SW360 to compile with Java 17, including necessary changes and dependency upgrades


heliocastro:
This is deprecated. Support for Java 17 is already present on main, but can't be used due Liferay issues.

Issue #1722, opened 2022-11-15
Make Authorities field mandatory while generating Token from Preferences


afsahsyeda:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description
The Authorities field containing checkbox(s) for Read and Write access is not mandatory which lets the user generate token without checking access. This leads to an empty array in the Authorities section of the generated token.

#### How to reproduce
1) Go To Preferences
2) Click on Generate Token without checking the Read Or Write Access checkbox.

#### Screenshots

![image](https://user-images.githubusercontent.com/115608771/201900539-f66dff88-e770-4d45-9b6f-ffda5a8b2ca1.png)

PR #1723, opened 2022-11-15
fix(PreferencesUI):Read Access has to be checked before Generating token

afsahsyeda:
Signed-off-by: afsahsyeda <afsah.syeda@siemens-healhtineers.com>

Changes:
1) Read Access will be checked by default while generating token.
2) It is mandatory for Read access to be checked.
3) An error message saying 'Read Access is required' will be displayed if Read access is unchecked by user.

Issue: #1722 

### How To Test?
Click on 'Generate Token' with the Read Access box unchecked.

image


heliocastro (approved):
We should have a way to notify that a new translation string was added, for translation purposes

akapti:
tested the changes. working as expected.

Issue #1726, opened 2022-11-22
Apply sorting (from newest to oldest) on versions in drop down menu when inspecting a specific component.


rudra-superrr:
### Description
My usual workflow to link releases to a project is by searching for a specific component (e.g. Angular) and choose from the drop down list a certain version.
This list is currently unsorted and this makes it kind of hard to find the version my project is using. Please see attached image

#### Steps followed and expected result
Describe the steps followed by you and your expected results after following the steps.

#### Screenshots
![image](https://user-images.githubusercontent.com/58290634/203297992-09b0498e-baeb-4b74-b8cb-86140f184571.png)

PR #1727, opened 2022-11-22
Feat(UI): apply sorting in release versions when inspecting component

rudra-superrr:
Issue:  #1726 

### How To Test?
Go to component tab and click the release dropdown button when inspecting a release of that component.
The list of release versions should be in descending order (newest to oldest).

image


nikkuma7:
Test successful.

akapti (commented):
code looks good to me.
one issue while testing I found is,
component with minor version are not being sorted as expected.
for example: version `3.12` will be below version `3.3`, `3.4` and so on..

sample screenshot below:
image

Issue #1747, opened 2022-11-30
Moderation requests: search/filter does not work


eldrin30:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description
Moderation requests: search/filter does not work

#### How to reproduce

1. Go to Requests tab
2. Filter the moderation request by searching for department 'SI' or 'ADV'
3. The search results an empty list even though there are many requests

PR #1749, opened 2022-11-30
fix(moderation_request): Added a check that if documentId is null then ignore

eldrin30:
Signed-off-by: Eldrin <eldrin.sanctis@siemens.com>

For moderation request filtering, added a check that if documentId is null then ignore that document

> * Which issue is this pull request belonging to and how is it solving it? (#1747 )

Issue: Moderation requests: search/filter does not work

### How To Test?

1. Go to Requests tab
2. Filter the moderation request by searching for department 'SI' or 'ADV'
3. The search results an empty list even though there are many requests


akapti (approved):
Looks good.

ag4ums (approved):
working as expected

Issue #1770, opened 2022-12-15
Some issues with PR 1447


tienlee:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

We tried to use PR #1447 and found some issues.

#### How to reproduce
**Issue 1**: Admin user creates a project A. Normal user can edit this project, create a new vendor A on project page. A moderation request will be created. If the admin user rejects this moderation request, vendor A still exists.

I don't know if this behavior is a bug or not.

**Issue 2**: When a user edits the project, in summary tab
+ Click on "Vendor" field, a "Search Vendor" dialog appears
+ Click on Search button
+ Choose a vendor.
+ Click on "Close" button on dialog
+ Click on "Vendor" field again
+ Choose a vendor
+ Click on "Select Vendor" button.

Result: "Vendor" field is blank.
I think this is a minor bug. Please consider fixing it.

**Issue 3**: Vendor will be removed after updating a project via API without changing vendor fields. I will fix this issue.
Precondition:
- There is a project (project name: P1, a vendor A was linked to P1

Step 1: Update project P1 via API http://{server_ip:port}/resource/api/projects/{IdProject}
{
    "name": "P2",
    "description": "This is a sample project",
}
Step 2. Check vendor information via API/GUI
Result: Vendor A was removed.

### Versions

* Last commit id on master:
* Operating System (lsb_release -a):

### Logs

Any logs (if any) generated in

#### SW360 logs

Logs generated under /var/log/sw360/sw360.log

#### Tomcat logs

Logs generated under /var/log/tomcat/error.log

PR #1771, opened 2022-12-16
Fix(REST): Fix bug update project without vendor information will remove existing vendor information

hoangnt2:
Fix bug update project without vendor information will remove existing vendor information

Fix issue 3 of #1770 

### How To Test?

Precondition:

There is a project (project name: P1, a vendor A was linked to P1
Step 1: Update project P1 via API http://{server_ip:port}/resource/api/projects/{IdProject}
{
"name": "P2",
"description": "This is a sample project",
}
Step 2. Check vendor information via API/GUI
Expected output: Vendor A won't be removed.


ag4ums (approved):
Changes looks good working as expected

Issue #1776, opened 2022-12-19
Bug get all components and create component REST API.


hoangnt2:
### Description
1. When get all components by REST API, visibility of component is not correct.
2. Create a new component by REST API (without businessUnit), do not set component group automatically.

##### Bug No1:  When get all components by REST API, visibility of component are not correct.
1. Send GET request to http://{server_ip:port}/resource/api/components.
- Current: Visibility of each component is not correct (always is EVERYONE)
- Expect: Show visibility of each component correctly.

##### Bug No2: Create a new component by REST API (without businessUnit), do not set component group automatically.
1. Send POST request to http://{server_ip:port}/resource/api/components to create a new component (without businessUnit).
- Current: BusinessUnit (Group) do not be set automatically.
- Expect: BusinessUnit (Group) will be set automatically (Admin group).

### Versions

* Last commit id on master: 0837f7a
* Operating System (lsb_release -a):

### Logs

Any logs (if any) generated in

#### SW360 logs

Logs generated under /var/log/sw360/sw360.log

#### Tomcat logs

Logs generated under /var/log/tomcat/error.log

PR #1777, opened 2022-12-19
Fix (REST): Fix bug do not set businessUnit automatically when create component. Get all components always return visibility EVERYONE.

hoangnt2:
Issue: #1776 

### How To Test?
Refer to: #1776 

### Checklist
Must:


KoukiHama:
when someone try to  do test, they need to change this parameter in sw360.properties

## This property is used to enable the component visibility restriction feature.
#component.visibility.restriction.enabled=true

nikkuma7:
Test successful.

ag4ums (approved):
Code looks good

PR #1813, opened 2023-01-19
fix(UI) : Remove null from User's lastname.

rudra-superrr:
Signed-off-by: rudra-superrr <rudra.chopra@siemens.com>

Description: Remove null from user's lastname and replacing it with dot(.) operator.

Issue: #1814 

### How To Test?
After running the script, goto users under admin tab and check if any user having lastname as null.

### DIFF
--- a/scripts/migrations/055_remove_null_from_lastname.py
+++ b/scripts/migrations/055_remove_null_from_lastname.py
@@ -1,0 +1,91 @@
+#!/usr/bin/python
+# -----------------------------------------------------------------------------
+# Copyright Siemens AG, 2023. Part of the SW360 Portal Project.
+#
+# This program and the accompanying materials are made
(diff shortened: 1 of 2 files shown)


heliocastro:
@rudra-superrr Why a dot ? Why not simply keep the field empty ?
If you are doing this because the UI representation comes NULL for an empty field, the change should be different.
We should NEVER touch the database

ag4ums:
> @rudra-superrr Why a dot ? Why not simply keep the field empty ? If you are doing this because the UI representation comes NULL for an empty field, the change should be different. We should NEVER touch the database

Db values should not be updated like this , seems we have a separate issue, In initial days some of the last-name were not captured properly in our db but later this was fixed in sw360, so the last-name with (null) need to updated by comparing the same with our corporate directory.

PR #1817, opened 2023-01-30
fix(db): Deactivate email notification of user not belonging to a domain

rudra-superrr:
Signed-off-by: rudra-superrr <rudra.chopra@siemens.com>

Issue: #1831 

### How To Test?
Run the script to test the changes in db.
User's wantMainNotification field will change to false if that user does not belong to that specific domain.
Note: In script, add the domains that should belong to a particular instance.

### DIFF
--- a/scripts/migrations/055_deactivateEmailNotification.py
+++ b/scripts/migrations/055_deactivateEmailNotification.py
@@ -1,0 +1,110 @@
+#!/usr/bin/python
+# -----------------------------------------------------------------------------
+# Copyright Siemens AG, 2023. Part of the SW360 Portal Project.
+#
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# This is a manual database migration script. It is assumed that a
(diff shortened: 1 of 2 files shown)


scripts/migrations/055_deactivateEmailNotification.py
```diff
@@ -0,0 +1,110 @@
...
+couch = couchdb.Server(COUCHSERVER)
+userDb = couch[USERS_DB]
+
+#Users
+NOTIFICATION = 'wantsMailNotification'
+EMAIL = 'email'
+
+####################################################################################################
+#Add the Domains that should belong to a particular instance and seperate them by using '|' operator
+DOMAINS = "@siemens.com|@siemens-energy.com|@siemens-healthineers.com|@gmail.com|sw360.org"
```
ag4ums, line 45: Please remove the siemens reference domain names may be only "@<domain>.com|@<domain>.org",  
also the comment can be more explicit like:
"#all email from the domains not included in list (separated with '|' ) will be disabled for email notification"

ag4ums (changes requested):

scripts/migrations/055_deactivateEmailNotification.py
rudra-superrr: done the changes.

ag4ums (approved):
Changes looks good

PR #1839, opened 2023-02-22
feat(ui): POC of sw360 frontend with React framework

hoangnt2:
This PR for discussing

This is the POC of Sw360 frontend with ReactJS
Implemented functions:
- Login page


ag4ums:
@hoangnt2, node version: v16.19.0 used here is going to be EOL soon (11-09-2023)
Can we consider latest LTS version (v18) (https://nodejs.dev/en/download/releases/).

hoangnt2:
@ag4ums, I'm using ubuntu 18.04 and it doesn't support node v18 so I installed node 16 for POC. But I think we should use lastest version (v18).

heliocastro:
@hoangnt2 I can present a better solution on the community meeting. Let's discuss there

heliocastro:
@ag4ums We should target Node 20, that will become next LTS, and dev cycle starts on April.
So, my proposal is start with Node 19 right now.

tienlee:
> I can present a better solution on the community meeting. Let's discuss there
Good news, let's share your idea.

> We should target Node 20 that will become next LTS, and dev cycle starts on April.
I agree with you about this target version.
> So, my proposal is start with Node 19 right now.
It's OK. Although the EOL of Node 19 is 2023-06-01, it is a PoC and can move to Node 20 soon.

KoukiHama:
https://github.com/eclipse-sw360/sw360-frontend
Here is the repo for new GUI framework. We can close this PR.

Issue #1874, opened 2023-03-20
REST API: enpoint api/projects does not return all projects


rudra-superrr:
### Description

REST API endpoint api/projects with the parameters allDetails=true and group=SI,
i.e. in total api/projects?allDetails=true&group=SI.

Unfortunately this does not return all projects matching the search spec.
The following projects are missing:

SiPass integrated - 2.76 SP1
SiPass integrated - 2.76 SP2
SiPass integrated - 2.85
Siveillance Vantage - Value Stream
Siveillance Identity - 1.2
Siveillance Identity - 1.3
Siveillance Identity - 1.4
Siveillance Identity - R1
Siveillance Identity - R2
Siveillance Identity - R3
Siveillance Identity - R4

hoangnt2:
@rudra-superrr, maybe search result is limited. You can try to modify **lucenesearch.limit** property in **/etc/sw360/couchdb.properties** file.

rudra-superrr:
Hi @hoangnt2 ,
Actually It's BU of the linked project that causing the problem.
Still thanks for the help.

PR #1875, opened 2023-03-20
Fix(rest): endpoint for projects does not return all projects with allDetails=true

rudra-superrr:
Issue: #1874 

Description:
REST API endpoint api/projects with the parameters allDetails=true and group=DEPARTMENT,
i.e. in total api/projects?allDetails=true&group=DEPARTMENT.
Unfortunately this does not return all projects matching the search spec as the projects that are not shown in response structure are the ones having linked projects of different group/BU.

How to test:
Test this endpoint and it will return all projects.


nikkuma7:
Test successful.

ag4ums (approved):
Changes looks good

Issue #1886, opened 2023-03-29
Attachments tab is not loading when there are many attachments in any document.


akapti:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

- Attachments tab is not loading in Project / Component / Release when there are more than 8 to 10 attachments.

#### How to reproduce

1. Create a Project / Component / Release
2. Add more than 8 to 10 attachments to the Project / Component / Release, and update.
3. Edit it again and you should see the Ajax request error.

#### Screenshots

![image](https://user-images.githubusercontent.com/50870174/228441932-b881e050-3022-4f24-864c-8c7fcdd26fdf.png)

PR #1887, opened 2023-03-29
fix(UI): Attachments tab not loading in UI

akapti:

- Attachment info will be sent in Ajax GET request only if there is duplicate error while updating the document.

Issue: #1886 

### Suggest Reviewer
@afsahsyeda 

### How To Test?
See the issue #1886  description

### Checklist
Must:

### UPDATE

The other way to fix this issue is to configure the `/tomcat/conf/server.xml` and `Liferay/portal-ext.properties` file to increase the default GET request size.

Changes in `Liferay/portal-ext.properties`:

- `invoker.filter.uri.max.length=8000` -> Refer: https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/portal.properties#L6852

Changes in  `/tomcat/conf/server.xml`:

- `maxHttpRequestHeaderSize="16384"` -> from default 8Kb to 16Kb or more (as per your need) Refer: https://tomcat.apache.org/tomcat-9.0-doc/config/http.html

- Ideally 8Kb is more than enough for `GET` request, changing this value might not be required, and changing only in `.portal-ext.properties` should be sufficient

```
    <Connector port="8080" protocol="HTTP/1.1"
               connectionTimeout="20000" maxHttpRequestHeaderSize="16384"
               redirectPort="8443" URIEncoding="UTF-8" />
```

### PS:
In case `tomcat` is running as proxy behind `apache2` server, then following property needs to be added in `apache2.conf` file:
`LimitRequestLine 16384`


smrutis1:
Tested it in stage. Working fine.

smrutis1 (approved):
Code looks good.

Issue #1808, opened 2023-01-17
Recent docker setup: use of "slim war" conceptually broken/risky


ljaenicke:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

The "create-slim-war-files.sh" script copies all jar files being part of the wars to a shared repository. As the versions of the packages are not synchronized, several versions of the same package may end up in the directory. As tomcat only loads one version of a package and ignores the others, this leads to more or less obvious failures.

#### How to reproduce

The duplication of commons-io in versions 1 and two leads to failure of the fossolgy component due to
  java.lang.NoSuchMethodError: long org.apache.commons.io.IOUtils.copyLarge
with copyLarge not being part of commons-io 1. (Result: on startup the fossology backend fails and the first login is stalled until timeout after 10 minutes)
(This is how I finally debugged the issue.)

Also most of the functions might fail with an "Unable to find user", I however do not know which confllict might be causing this behavior. In any case couchdb does not see traffic during operation with the "sim-war" setup.

List of packages, some duplicates may be the cause of errors, some may be non-critical.

accessors-smart-2.4.8.jar
activation-1.1.1.jar
activation-1.1.jar
annotations-1.3.2.jar
annotations-15.0.jar
apache-mime4j-core-0.7.2.jar
apache-mime4j-dom-0.7.2.jar
asm-9.1.jar
asm-debug-all-4.1.jar
aspectjrt-1.8.0.jar
bcmail-jdk15-1.45.jar
bcpkix-jdk15on-1.64.jar
bcprov-jdk15-1.45.jar
bcprov-jdk15on-1.64.jar
boilerpipe-1.1.0.jar
btf-1.3.jar
build-configuration-17.0.0-SNAPSHOT.jar
byte-buddy-1.12.13.jar
checker-qual-3.12.0.jar
cloudant-client-2.19.1.jar
cloudant-http-2.19.1.jar
collection-0.7.jar
commonIO-17.0.0-SNAPSHOT.jar
commons-beanutils-1.7.0.jar
commons-beanutils-1.9.4.jar
commons-beanutils-core-1.8.0.jar
commons-cli-1.4.jar
[...]

Issue #1810, opened 2023-01-17
While loading liferay "unable to load filter com.liferay.portal.kernel.servlet.filters.invoker.InvokerFilter" is logged


ljaenicke:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

While loading liferay (on startup), the error message
"unable to load filter com.liferay.portal.kernel.servlet.filters.invoker.InvokerFilter" is shown.

#### How to reproduce

Perform a startup (I used the Docker environment) and start working with SW360 and review the output.

* Last commit id on master: fe845bce389b7cd85101084285422535c7f06326
* Operating System (lsb_release -a):

PR #1920, opened 2023-04-19
Version normalization

heliocastro:
Cleanup versions due report in issue #1808

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -199,12 +199,13 @@ RUN --mount=type=bind,target=/build/sw360,rw \
 WORKDIR /sw360_tomcat_webapps/
 
 COPY scripts/create-slim-war-files.sh /bin/slim.sh
-COPY --from=sw360clucene /couchdb-lucene.war /sw360_tomcat_webapps
+
 RUN bash /bin/slim.sh
 
 FROM scratch AS sw360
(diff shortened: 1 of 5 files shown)


ljaenicke:
I have tested the branch. It looks much better now... but I still am seeing issues.
To reproduce: build everything from scratch (all containers pruned, volumes removed, ...). The server now starts up fine. I can set up the initial system following https://www.eclipse.org/sw360/docs/deployment/legacy/deploy-liferay7.4/.
The first step now is to load the SPDX license information (Still using the setup@sw360.org account, Admin->License->Import SPDX Information). This operation fails:
Screenshot 2023-04-20 084826
In the log, the information
  ERROR [UserCacheHolder:76] Unable to fetch user...
can be found.
[transcript.gz](https://github.com/eclipse-sw360/sw360/files/11283409/transcript.gz)

ljaenicke:
Note: I when testing yesterday with the slim-war mechanism disabled, I was able to successfully import the SPDX information, so I do not think this might be related to Issue #1810.

heliocastro:
@KoukiHama We still have all the old and new spdx libraries as dependencies. Can you check what are valid ?

heliocastro:
@ljaenicke Btw, this is the experimental test branch, will not be committed this way. I commited a single ont for remove couchdb_lucene from slim_jars now

Issue #1937, opened 2023-04-25
Data Quality and reduction of storage


rudra-superrr:
### Description

Releases are partly very old, have no up to date clearing results . Such releases and also components which are not used in any project should be deleted.

This script will delete the Releases that are not linked to any project and does not have any clearing results and will also delete the components that does not have any releases.

PR #1938, opened 2023-04-25
fix(db): Data quality and reduction of storage

rudra-superrr:
Issue: #1937 

Description: this migration script will delete the Releases that are not linked to any project and does not have any clearing results and will also delete the components that does not have any releases.

Note: before running the script, enter the date in bufferDate variable.
This date will collect the releases and components that were created before or on this date.


ag4ums (approved):
changes looks good, working as expected

Issue #1659, opened 2022-10-03
REST API /api/releases/searchByExternalIds is not working


smrutis1:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

Prerequisites:  
We have for example the component release @angular/cdk (13.3.2)  
With following externalID: `SHS.AutoIdentification = pkg:npm/%40angular/cdk@13.3.2`

Action:  
Execute following REST Call (constructed as described here: https://sw360.org/resource/docs/api-guide.html#resources-releases-get-by-externalids) !  
`https://sw360.org/resource/api/releases/searchByExternalIds?SHS.AutoIdentification=pkg:npm/%40angular/cdk@13.3.2`

Result:  
Response is empty

Expected:  
Response will include the release @angular/cdk (13.3.2)

Remark:  
Tested this also for @ngx-translate/core (14.0.0)
With externalID: `package-url = pkg:npm/%40ngx-translate/core@14.0.0`

PR #2022, opened 2023-06-30
fix(Rest): Allowing search for externalIds without encoding

eldrin30:

Modified code such that while searching for releases using externalIds via Rest endpoint, encoding is skipped
> * Which issue is this pull request belonging to and how is it solving it? (#1659)

### Suggest Reviewer
@ag4ums 

### How To Test?
Search for releases using externalId such that in key-value pair there are values that usually encodes to special characters(eg: %40)


ag4ums (approved):
Changes looks good, working as expected

Issue #1809, opened 2023-01-17
OSADL Obligation import: only 20 obligations imported or shown


ljaenicke:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

After importing OSADL obligations an error message is shown, that not all obligations could be imported "Error happend during license obligation importing. Some license obligations may not be imported.

In the log files, some licenses are not found (404), they are also not listed on the OSADL pages. https://www.osadl.org/Access-to-raw-data.oss-compliance-raw-data-access.0.html

In some other cases, errors are logged:
^[[33mcouchdb_1     |^[[0m [notice] 2023-01-16T18:24:58.706031Z nonode@nohost <0.25499.0> b214ba0062 couchdb:5984 172.27.0.4 admin GET /sw360db/_design/ObligationNode/_view/byobligationnodetype?include_docs=true&startkey=%22OR%22&endkey=%22OR%EF%BF%B0%22 200 ok 19 
^[[36msw360         |^[[0m 2023-01-16 18:24:58 ERROR LicenseDatabaseHandler:1055 - Can not add nodes from json object: {"val":["Obligation","YOU MUST","Provide","Delayed source code delivery Of Modified work"],"children":[{"val":["ATTRIBUTE","Machine-readable"],"children":[]},{"val":["ATTRIBUTE","Via Internet"],"children":[]},{"val":["ATTRIBUTE","No profit"],"children":[]},{"val":["ATTRIBUTE","Duration As long as distributed"],"children":[]}]}^M 
^[[36msw360         |^[[0m 2023-01-16 18:24:58 ERROR LicenseDatabaseHandler:1055 - Can not add nodes from json object: {"val":["Obligation","YOU MUST","Reference","Source code"],"children":[{"val":["ATTRIBUTE","Below Copyright notice"],"children":[]}]}

In any case, the list of obligations shown afterwards only contains 20 entries (from A-B), which may be that only 20 obligations have been imported or a problem with the display.

### Versions

* Last commit id on master: fe845bce389b7cd85101084285422535c7f06326
* Operating System (lsb_release -a): (Docker setup)

### Logs
Appended
[...]

PR #2061, opened 2023-07-24
Add missing obligation rules

heliocastro:
Some entries from current OSADL obligations format were missing:
* **YOU MUST**
* **YOU MUST NOT**

Reference: Issue #1809


KoukiHama (approved):
Codes are ok.
and alphabetical order is good to maintain codes.

PR #2099, opened 2023-08-23
fix(GUI): Fix edit dependency network GUI broken

hoangnt2:
### Precondition
 - Add following option to **/etc/sw360/sw360.properties** file **enable.flexible.project.release.relationship=true**
### Issue

#2100 

- When user click **Add release** button to add a release to dependency network, the GUI will be broken as bellow
MicrosoftTeams-image

### How To Test?
 Steps:
  - User click edit project
  - User select a release and click
  
 Expected:
  - The GUI will work normally
image


KoukiHama:
Codes are OK and Works fine.

Issue #2195, opened 2023-11-16
fix (rest) : Rest API for search component by Lucene search


keerthi-bl:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

Currently in link release api, there is no search feature while finding releases. It needs to be added using lucene search. For the reference, the link to the API documentation of `link project` is mentioned below.

https://stage.sw360.siemens.com/resource/docs/api-guide.html#resources-projects-list-by-lucene-search

#### How to reproduce

NA

#### Screenshots

NA

### Versions

* Last commit id on master:
* Operating System (lsb_release -a):

### Logs

Any logs (if any) generated in

#### SW360 logs

Logs generated under /var/log/sw360/sw360.log

#### Tomcat logs

Logs generated under /var/log/tomcat/error.log

keerthi-bl:
PR : #2201

PR #2201, opened 2023-11-17
fix (rest) : Rest API for component search by Lucene search

keerthi-bl:

Issue: #2195 

Sample URL and response as follows in the screenshot. 

URL : http://localhost:8080/resource/api/components?name=comp&page=0&page_entries=20&luceneSearch=true&sort=name,desc

image

No

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
@@ -23,6 +23,7 @@ import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.common.SW360Constants;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
+import org.eclipse.sw360.datahandler.couchdb.lucene.LuceneAwareDatabaseConnector;
 import org.eclipse.sw360.datahandler.resourcelists.PaginationParameterException;
 import org.eclipse.sw360.datahandler.resourcelists.PaginationResult;
 import org.eclipse.sw360.datahandler.resourcelists.ResourceClassNotFoundException;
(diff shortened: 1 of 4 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -141,6 +143,7 @@ public ResponseEntity<CollectionModel<EntityModel<Component>>> getComponents(
             @RequestParam(value = "fields", required = false) List<String> fields,
             @Parameter(description = "Flag to get components with all details.")
             @RequestParam(value = "allDetails", required = false) boolean allDetails,
+            @RequestParam(value = "luceneSearch", required = false) boolean luceneSearch,
```
hoangnt2, line 146: Add @Parameter same like other params

hoangnt2:
Another thing, this snippet is unnecessary in lucene search case.

image

It also causes the bug that don't have enough entries on a page when we search by component type.
For example: I want a page with only 10 items. But when I search by Component Type some pages don't have enough entries.

I think the filter should be done before we create pagination result.
I am appreciate if you can help me to fix it.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
keerthi-bl: done

hoangnt2: @keerthi-bl could you take a look on my [comment](https://github.com/eclipse-sw360/sw360/pull/2201#issuecomment-1816034322)?

keerthi-bl: @hoangnt2 , I have addressed that and changed the snippet accordingly, please have a look.

hoangnt2:
Change look good to me

PR #2218, opened 2023-11-27
fix(importCDX): Update failed component creation error message

### LINKED ISSUE #2215: Incorrect error being displayed during component creation.
### Description
During the manual creation of an component, if we provide the (VCS) that is already associated with another component,  the system won't let you create the new one. It doesn't matter if the component name is different. Instead, it shows an error saying that ( **a component with the same name already exists** ).
MicrosoftTeams-image (3)

sameed20:
Updated the previous error message to: **A component with the same name or VCS already exists.**

Issue:  closes #2215


ag4ums (approved):
LGTM

Issue #2220, opened 2023-11-28
Duplicate releases are getting created on sw360 during import of an SBOM


sameed20:
### Description
While importing a SBOM, a package gets created based on the purl and gets linked to a components based on its VCS. If a VCS already exists on SW360, it should check for the release based on the version tag. However, it is checking based on the name "manufacturer.component_name" and if not found, creating a new Release of different name under the same component.

![zrender](https://github.com/eclipse-sw360/sw360/assets/141239852/5e0c7e54-cd8c-4f6c-8b88-acd1af868372)

PR #2221, opened 2023-11-29
fix(importCDX): Resolve duplicate release creation during SBOM import

sameed20:
Now during the SBOM import, If we found a component already exists' s in SW360, then we are going to use the existing component name. After this we will search for the release in sw360 that we want to create using the SBOM, and if we found the release we are going to use the same release.

Closes: #2220

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
@@ -572,6 +572,8 @@ public class CycloneDxBOMImporter {
 
                 if (CommonUtils.isNotNullEmptyOrWhitespace(compAddSummary.getId())) {
                     comp.setId(compAddSummary.getId());
+                    String existingCompName = getComponetNameById(comp.getId(), user);
+                    comp.setName(existingCompName);
                 } else {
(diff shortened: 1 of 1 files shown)


nikkuma7:
I can able to upload the cycloneDX and component create along with release successfully. But vcs name is missing in created component.

sameed20:
Hi @nikkuma7 for a component to get created with this pattern vendor_name.repo_name, the component should contain a VCS field for that component.

nikkuma7:
Hi @sameed20  I have created two release with same name only the version is changed but you have above mentioned component should create in specific pattern ( vendor_name.repo_name) and component contains VCS name is not coming in pattern.

Below release SS :
image

Below Component SS:
image

sbom file ss:
image

afsahsyeda:
@sameed20 please handle the use-case where the version is the same but the letter v is missing.
image

sameed20:
Hi @afsahsyeda the release version v7.23.1 and 7.23.1 in the above provided screenshot is different if we go the babel github repo and look for the release version 7.23.1 then there exists nothing of such version but when we look for a release version v7.23.1 then we will end up finding the right release. 
Now, even if we apply the check on character v in the release version, then user can also create a release with such version e.g. a_7.23.1 or b_7.23.1 because we accept a string in the version field of the release.

PR #2241, opened 2023-12-14
fix(script): Fix migration script not working with python3

hoangnt2:
### Issue:
Migration script **060_migrate_project_dependency_network.py** does not work with Python3


KoukiHama:
Works fine and codes  are good

Issue #2287, opened 2024-01-24
Create a new Clearing Request state "Sanity Check" to perform sanity check before accepting a project.


akshitjoshii:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description

Clearing Request for a project should have a state **Sanity Check** for the clearing team to make sure that the CR for a project is eligible to be accepted. 

When the user changes the CR state from `New` to `Sanity Check` a csv containing all the linked release info with the name of `attachment type` = `SRC/SRS` is created. The clearing team shall perform the sanity check by verifying the details before accepting a `Clearing Request`.

PR #2319, opened 2024-02-19
feat(CR): Create a new Clearing Request state Sanity Check to perform sanity check before accepting a project.

akshitjoshii:
Clearing Request for a project should have a state **Sanity Check** for the clearing team to make sure that the CR for a project is eligible to be accepted. Clearing admin can change the CR state from `New` to `Sanity Check`. They can navigate to the **License Clearing** tab in the project portlet and select **Export Project with linked releases** option.
Screenshot from 2024-02-19 12-03-43

This will download an excel file containing all previous info with 2 additional columns:
**Source Code Downloadurl** - contains the Source Code Downloadurl for each linked releases 
**SRC/SRS Attachments** - contains the filenames of SRC and SRS type attachments for each linked release.
Clearing Team can go through this data to make sure CR is eligible to be accepted or not.
MicrosoftTeams-image (1)

Closes: #2287

### DIFF
--- a/libraries/datahandler/src/main/thrift/components.thrift
+++ b/libraries/datahandler/src/main/thrift/components.thrift
@@ -234,6 +234,10 @@ struct Release {
     // information from external data sources
     9: optional  map<string, string> externalIds,
     300: optional map<string, string> additionalData,
+    
+    // Urls for the project
+    70: optional string sourceCodeDownloadurl, // URL for download page for this release source code
+    71: optional string binaryDownloadurl, // URL for download page for this release binaries
 
     // Additional informations
     10: optional set<Attachment> attachments,
@@ -270,9 +274,6 @@ struct Release {
     55: optional EccInformation eccInformation,
     56: optional set<string> softwarePlatforms,
 
-    // Urls for the project
-    70: optional string sourceCodeDownloadurl, // URL for download page for this release source code
-    71: optional string binaryDownloadurl, // URL for download page for this release binaries
 
     80: optional map<string, ReleaseRelationship> releaseIdToRelationship,    //id, comment
     81: optional set<string> packageIds,
--- a/libraries/datahandler/src/main/thrift/sw360.thrift
+++ b/libraries/datahandler/src/main/thrift/sw360.thrift
@@ -71,8 +71,9 @@ enum ClearingRequestState {
     IN_QUEUE = 3,
     IN_PROGRESS = 4,
     CLOSED = 5,
-    AWAITING_RESPONSE = 6
-    ON_HOLD = 7
+    AWAITING_RESPONSE = 6,
+    ON_HOLD = 7,
+    SANITY_CHECK = 8
 }
 
 enum ClearingRequestPriority {
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
@@ -12,12 +12,15 @@ package org.eclipse.sw360.exporter.helper;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.thrift.TException;
+import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
 import org.eclipse.sw360.datahandler.common.ThriftEnumUtils;
 import org.eclipse.sw360.datahandler.common.WrappedException.WrappedSW360Exception;
 import org.eclipse.sw360.datahandler.thrift.ReleaseRelationship;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
 import org.eclipse.sw360.datahandler.thrift.ThriftUtils;
+import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
+import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentType;
 import org.eclipse.sw360.datahandler.thrift.components.*;
 import org.eclipse.sw360.datahandler.thrift.users.RequestedAction;
 import org.eclipse.sw360.datahandler.thrift.users.User;
@@ -26,6 +29,7 @@ import org.eclipse.sw360.exporter.ReleaseExporter;
 import org.eclipse.sw360.exporter.utils.SubTable;
 
 import java.util.*;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyList;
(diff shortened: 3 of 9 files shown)


libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -211,15 +212,30 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
...
                 row.add(fieldValueAsString(fieldValue));
         }
-    }
+    }    
+    
+    private void addSourceAttachmentsInformationToRow(Release release, List<String> row) throws SW360Exception{
+        Set<Attachment> attachments = release.getAttachments() != null ? release.getAttachments() : Collections.emptySet();
 
+        String result = attachments.stream()
+                .filter(attachment -> attachment.getAttachmentType().name().equals("SOURCE") || attachment.getAttachmentType().name().equals("SOURCE_SELF"))
```
akapti, line 232: I suggest to change the filter as below.
Define a Predicate:
```java
final Predicate<Attachment> isSourceAttachment = attachment -> AttachmentType.SOURCE.equals(attachment.getAttachmentType()) || AttachmentType.SOURCE_SELF.equals(attachment.getAttachmentType());
```
Use the predefined Predicate in the filter.
```java
.filter(isSourceAttachment)
```

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -211,15 +212,30 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
                 addReleaseIdToRelationShipToRow(release.getReleaseIdToRelationship(), row);
                 break;
             case ATTACHMENTS:
-                String size = Integer.toString(release.isSetAttachments() ? release.getAttachments().size() : 0);
-                row.add(size);
+                if(isForProjectExport) {
+                    addSourceAttachmentsInformationToRow(release, row);
+                    }else {
```
akapti, line 217: Please format the code.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -211,15 +212,30 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
...
                 break;
             default:
                 Object fieldValue = release.getFieldValue(field);
                 row.add(fieldValueAsString(fieldValue));
         }
-    }
+    }    
+    
+    private void addSourceAttachmentsInformationToRow(Release release, List<String> row) throws SW360Exception{
+        Set<Attachment> attachments = release.getAttachments() != null ? release.getAttachments() : Collections.emptySet();
```
akapti, line 229: Consider using `CommonUtils.nullToEmptySet`

libraries/datahandler/src/main/thrift/sw360.thrift
```diff
@@ -66,13 +66,14 @@ enum ModerationState {
...
-    AWAITING_RESPONSE = 6
-    ON_HOLD = 7
+    SANITY_CHECK = 1,
+    ACCEPTED = 2,
+    REJECTED = 3,
+    IN_QUEUE = 4,
+    IN_PROGRESS = 5,
+    CLOSED = 6,
+    AWAITING_RESPONSE = 7,
+    ON_HOLD = 8
```
akapti, line 76: I suggest not to change the **`enum`** values, and keep the new entry as `SANITY_CHECK = 8`.
It might create issue if these values are mapped / inserted to the database instead of keys. if it's mapped, then you might need to write the migration script or else this change should be good.

libraries/datahandler/src/main/thrift/components.thrift
```diff
@@ -234,6 +234,10 @@ struct Release {
     // information from external data sources
     9: optional  map<string, string> externalIds,
     300: optional map<string, string> additionalData,
+    
+    // Urls for the project
+    70: optional string sourceCodeDownloadurl, // URL for download page for this release source code
+    71: optional string binaryDownloadurl, // URL for download page for this release binaries
```
akapti, line 240: why not use proper camel casing?
`sourceCodeDownloadUrl` and `binaryDownloadUrl`

akapti (changes requested):
I have reviewed the code.
suggested some code changes.

nikkuma7:
Test successful.

image
image

libraries/datahandler/src/main/thrift/components.thrift
akshitjoshii: I have just shifted these variables in the .thrift file to fix the **sourceCodeDownloadUrl** column location in the excel. On modifying them to suggested camel case, it affects the generated code that relies on those variable names. For ex. SOURCE_CODE_DOWNLOADURL is after building changes to SOURCE_CODE_DOWNLOAD_URL and hence it gives not found variable compilation error. Then changes have to be made in all places where this has been used.
Screenshot from 2024-02-29 14-18-30

akshitjoshii:
Addressed review comments.

libraries/datahandler/src/main/thrift/components.thrift
akapti: Thanks for explaining

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -211,15 +214,30 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
...
                 break;
             default:
                 Object fieldValue = release.getFieldValue(field);
                 row.add(fieldValueAsString(fieldValue));
         }
+    }    
+    
+    private void addSourceAttachmentsInformationToRow(Release release, List<String> row) throws SW360Exception{
+        Set<Attachment> attachments = CommonUtils.nullToEmptySet(release.getAttachments());
+        final Predicate<Attachment> isSourceAttachment = attachment -> attachment.getAttachmentType().name().equals("SOURCE") || attachment.getAttachmentType().name().equals("SOURCE_SELF");
```
akapti, line 232: It is best practice to keep the constant / enum at first whenever you are comparing it with a variable.
Because there is a possibility of `getAttachmentType` returning null, and above code will throw `NullPointerException` in such cases.

Please refer: https://mkyong.com/java-best-practices/constant-value-should-always-come-first-in-comparison/

I suggest to change it something like this: `AttachmentType.SOURCE.equals(attachment.getAttachmentType())`

akapti (changes requested):
Requested one small change to avoid possible NPE.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/ReleaseHelper.java
```diff
@@ -211,15 +214,30 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
...
+        String result = attachments.stream()
+                .filter(isSourceAttachment)
+                .map(Attachment::getFilename)
+                .collect(Collectors.joining(", "));
+
+        row.add(result);
     }
-
+
```
afsahsyeda, line 240: Remove this line.

Issue #2409, opened 2024-04-15
Moderation request comment can not contain characters outside of the Latin1 range


hoangnt2:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Description
Moderation request comment can not contain characters outside of the Latin1 range

#### How to reproduce

1. Login with with User1, then create project PR1 with visibility EVERYONE

2. Login with User2, visit project page then click delete PR1. The delete dialog will show as below
![image](https://github.com/eclipse-sw360/sw360/assets/94659621/bc660114-33ae-44d5-8b42-69af2c6d2edb)

3. Enter this comment  "てすと", then click **Delete project**. There no moderation request are created and delete button will be disabled with spinner
![image](https://github.com/eclipse-sw360/sw360/assets/94659621/3acbeedd-a16c-42b1-9ea3-a5ad2c744d75)

**This error also happen with release and component pages**

### Expected:
Dialog will show create moderation request message as below 
![image](https://github.com/eclipse-sw360/sw360/assets/94659621/720e819b-c447-42c6-a292-d296e238d756)

### Versions

* Last commit id on master:
* Operating System (lsb_release -a):

PR #2410, opened 2024-04-15
fix(ModerationRequest): Fix bug moderation comment not allow 2-byte characters

hoangnt2:
Issue: #2409 

### How To Test?
Refer to  #2409


GMishx:
The changes are related to frontend which is now removed. Can we close this PR @hoangnt2 in that case? Or some changes are required in backend as well?

hoangnt2:
@GMishx I think we can close this PR

Issue #2411, opened 2024-04-15
Improve Error Messages in REST API for CycloneDX import


sameed20:
When utilizing the REST API to import a Software Bill of Materials (SBOM) for an existing project, encountering a scenario where the SBOM contains different name and version information than the project which is present on SW360 than it results in an unhelpful "500 Internal Server Error" response.

PR #2414, opened 2024-04-17
fix(REST): Improve error message handling for CycloneDX sbom import using REST API

sameed20:
Now, if we try to import the SBOM on a already existing project in SW360 and if this SBOM contains different name and version information than the project which is present on SW360, then were handling this case properly in showing the relevant message to the user.
pr_creation

Closes: #2411


rudra-superrr:
Looks good to me.

image

image

PR #2444, opened 2024-05-08
fix(SRCUploadService): Source upload should continue  for release without halting if any invalid vcs url found

FarooqAftab:
---
> * Resolves https://github.com/eclipse-sw360/sw360/issues/2421 
---


sameed20:
Hi @FarooqAftab In the current scenario if we have one comp with this invalid VCS https:/github.com/angular/angular.js. Then it's not blocking the download of SRC code for other comps which have the correct VCS field. Instead of this it's throwing this error message in the logs for the incorrect VCS and successfully uploading the SRC code for other comps.
fpr

Now, on switching to this PR, the functionality is still the same like it's uploading the SRC code for the correct VCS and failing for the invalid VCS. The only change that I can see as of now is in the error logs.
fpr2

GMishx:
@FarooqAftab please check the comments from @sameed20

PR #2446, opened 2024-05-09
fix(api): throw 409 if last moderator

GMishx:
1. Throw `HttpClientErrorException` instead of `InvalidParameterException` to get correct status code instead of 500.
2. Also, change the status code from 400 to 409 incase the moderator is last moderator.

### How To Test?
1. Create a moderation request.
2. Use REST API to unassign moderators from the MR.
3. Upon removal of last moderator, endpoint should return status code 409 instead of 500.

### Checklist
Must:


smrutis1 (approved):
Code looks good.

rudra-superrr:
Testing successful.
image

PR #2483, opened 2024-06-10
feat(ImportSBOM):Change naming convention of imported components

afsahsyeda:
### Description

Components are being created from the vcs field in a CDX SBOM using the naming convention "manf.name". This has been changed to create the name with only the component name extracted from the vcs, omitting the manufacturer part unless and until there is a name class, which in that case the name would be created as manf/name. 

#### How to reproduce

Import a CDX SBOM

image


GMishx:
@afsahsyeda can you please rebase the PR with main branch to get the latest REST API test cases?

afsahsyeda:
@akshitjoshii Please test this PR.

akshitjoshii:
Test is successful. Working as expected.

pr-test1

GMishx (approved):
Changes looks good.

Issue #2497, opened 2024-07-01
Need modified attachment info in response to the moderation request rest API


keerthi-bl:
### Description

Let's suppose an attachment of a project is modified and MR is created. And the rest api is called to get the details of that MR using the endpoint: `localhost:8080/resource/api/moderationrequest/[id]`

The rest api response doesn't contain any modified attachment info. But ideally, it should contain the modified attachments.

 
Below data should be available in the API end point.

![image](https://github.com/eclipse-sw360/sw360/assets/121924406/f16d8a3d-8404-416d-a6f0-e6bb8a738104)

PR #2405, opened 2024-04-13
fix(rest): Added a commit message while create a moderation request.

nikkuma7:
[//]: # (Copyright Bosch.IO GmbH 2020)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

> Please provide a summary of your changes here.
> * Which issue is this pull request belonging to and how is it solving it? (*Refer to issue here*)
> * Did you add or update any new dependencies that are required for your change?

Issue: Closes : #2404 

### Checklist
Must:
- [ ] All related issues are referenced in commit messages and in PR

PR #2498, opened 2024-07-01
fix(rest) : modified attachment info in response to the moderation request rest api

keerthi-bl:
Closes : #2497 

### Description

#### How to reproduce

image


rudra-superrr:
Resolve the merge conflicts

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
```diff
@@ -116,26 +124,36 @@ public ResponseEntity<CollectionModel<ModerationRequest>> getModerationRequests(
         return getModerationResponseEntity(pageable, request, allDetails, modRequestsWithPageData);
     }
 
-    @Operation(
-            summary = "Get a single moderation request.",
-            description = "Get a single moderation request by id.",
-            tags = {"Moderation Requests"}
-    )
+    @Operation(summary = "Get a single moderation request.", description = "Get a single moderation request by id.", tags = {
+            "Moderation Requests" })
```
GMishx, line 128: Please revert irrelevant changes.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
```diff
@@ -116,26 +124,36 @@ public ResponseEntity<CollectionModel<ModerationRequest>> getModerationRequests(
...
     @RequestMapping(value = MODERATION_REQUEST_URL + "/{id}", method = RequestMethod.GET)
-    public ResponseEntity<HalResource<ModerationRequest>> getModerationRequestById(
-            @Parameter(description = "The id of the moderation request to be retrieved.")
-            @PathVariable String id
-    ) throws TException {
+    public ResponseEntity getModerationRequestById(
```
GMishx, line 130: Any reason to remove the object type from `ResponseEntity`? It causes wrong Spring documents to be generated.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
keerthi-bl: reverted

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
keerthi-bl: Added proper Object type.

rudra-superrr:
Testing was successful.
image

rudra-superrr:
One thing that I have found is, when a user updates a project then MR is created but if the same user updates some other field of the same project then it overwrites the previous moderation request.

GMishx:
> One thing that I have found is, when a user updates a project then MR is created but if the same user updates some other field of the same project then it overwrites the previous moderation request.

Can you please open an issue for the same? We need to rework on the Moderation Request with REST API. With the Liferay UI, we had the liberty to check when an MR will be created and how to treat it. But with REST API, currently there is no way to check the same and we have to revert to this behavior.

Please check https://github.com/eclipse-sw360/sw360/pull/2405#issuecomment-2426291315 as well.

rudra-superrr:
> > One thing that I have found is, when a user updates a project then MR is created but if the same user updates some other field of the same project then it overwrites the previous moderation request.
> 
> Can you please open an issue for the same? We need to rework on the Moderation Request with REST API. With the Liferay UI, we had the liberty to check when an MR will be created and how to treat it. But with REST API, currently there is no way to check the same and we have to revert to this behavior.
> 
> Please check [#2405 (comment)](https://github.com/eclipse-sw360/sw360/pull/2405#issuecomment-2426291315) as well.

But this issue is happening because of the changes made in this PR.
Please have a look into this @keerthi-bl

rudra-superrr:
> > One thing that I have found is, when a user updates a project then MR is created but if the same user updates some other field of the same project then it overwrites the previous moderation request.
> 
> Are you sure about this finding @rudra-superrr ? This PR is modifying only GET endpoint. Ideally it should not effect any modification.

We both also discussed regarding the same point but while testing issue is coming with this branch.

GMishx:
Okay,. Keeping this branch on hold then.

GMishx:
@rudra-superrr I just tested and confirmed by @keerthi-bl as well, the issue exists with `main` branch not with this PR.

Created issue #2828 for the same.

GMishx (approved):
Changes looks good.

Issue #2503, opened 2024-07-08
fix(rest) : modified accept request to see the proposed changes in Project/Component/Release to moderation request rest Api.


nikkuma7:
### Description
Upon clicking the "Accept Request" button for a moderation request, the REST API response returns a 202 (Accepted) status. However, the database is not being updated with the proposed changes. For example, if a project page is edited and a moderation request is created, accepting the proposed changes does not update the project page with the new values; it still shows the old values.

Please describe your situation in few words here.

#### Steps followed and expected result

Describe the steps followed by you and your expected results after following the steps.

#### Screenshots ( if applicable )
![Uploading image.png…]()

Add screenshots to help explain your problem.

PR #2504, opened 2024-07-08
fix(rest) : modified accept request to see the proposed changes in Project/Component/Release to moderation request rest Api.

nikkuma7:
Closes: #2503 

### Description
Upon clicking the "Accept Request" button for a moderation request, the REST API response returns a 202 (Accepted) status. However, the database is not being updated with the proposed changes. For example, if a project page is edited and a moderation request is created, accepting the proposed changes does not update the project page with the new values; it still shows the old values.

image

#### How to reproduce

### Versions

* Docker version OR
* Last commit id on main

#### SW360 logs

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
@@ -28,13 +28,11 @@ import org.eclipse.sw360.datahandler.thrift.RequestStatus;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
 import org.eclipse.sw360.datahandler.thrift.ThriftClients;
 import org.eclipse.sw360.datahandler.thrift.components.ComponentService;
-import org.eclipse.sw360.datahandler.thrift.components.Release;
 import org.eclipse.sw360.datahandler.thrift.licenses.LicenseService;
 import org.eclipse.sw360.datahandler.thrift.moderation.ModerationRequest;
 import org.eclipse.sw360.datahandler.thrift.moderation.ModerationService;
 import org.eclipse.sw360.datahandler.thrift.projects.ProjectService;
 import org.eclipse.sw360.datahandler.thrift.spdx.documentcreationinformation.DocumentCreationInformationService;
-import org.eclipse.sw360.datahandler.thrift.spdx.documentcreationinformation.DocumentCreationInformationService.Iface;
 import org.eclipse.sw360.datahandler.thrift.spdx.spdxdocument.SPDXDocumentService;
 import org.eclipse.sw360.datahandler.thrift.spdx.spdxpackageinfo.PackageInformationService;
 import org.eclipse.sw360.datahandler.thrift.users.User;
@@ -58,7 +56,6 @@ import java.util.Map;
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class Sw360ModerationRequestService {
     private static final Logger log = LogManager.getLogger(Sw360ModerationRequestService.class);
-
 
     @Value("${sw360.thrift-server-url:http://localhost:8080}")
     private String thriftServerUrl;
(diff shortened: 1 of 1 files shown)


GMishx:
@nikkuma7 please rebase with latest main.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
```diff
@@ -193,10 +240,87 @@ public Map<PaginationData, List<ModerationRequest>> getRequestsByState(User sw36
...
+            } else {
+                sw360ProjectClient.updateProjectFromModerationRequest(request.getProjectAdditions(),
+                        request.getProjectDeletions(), reviewer);
+            }
+        }
+            break;
+        case RELEASE: {
+            if (request.isRequestDocumentDelete()) {
+                RequestStatus deleteStatus = sw360ComponentClient.deleteRelease(request.getDocumentId(), reviewer);
+                if (deleteStatus.equals(RequestStatus.SUCCESS)) {
```
GMishx, line 273: Should the endpoint respond with failure if the `deleteStatus` is not successful?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
```diff
@@ -193,10 +240,87 @@ public Map<PaginationData, List<ModerationRequest>> getRequestsByState(User sw36
...
+                packageInfoClient.deletePackageInformation(request.getDocumentId(), reviewer);
+            } else {
+                packageInfoClient.updatePackageInfomationFromModerationRequest(request.getPackageInfoAdditions(),
+                        request.getPackageInfoDeletions(), reviewer);
+            }
+        }
+            break;
+        }
         getThriftModerationClient().acceptRequest(request, moderatorComment, reviewer.getEmail());
         return ModerationState.APPROVED;
```
GMishx, line 317: The status here should be dependent on response of action performed like deletion or updation? Right now, the endpoint always accept the request and return APPROVED status.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
nikkuma7: Addressed.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
nikkuma7: comment Addressed.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
```diff
@@ -242,102 +245,93 @@ public Map<PaginationData, List<ModerationRequest>> getRequestsByState(User sw36
...
+            break;
+        case RELEASE: {
+            if (request.isRequestDocumentDelete()) {
+                actionStatus = sw360ComponentClient.deleteRelease(request.getDocumentId(), reviewer);
+                if (actionStatus.equals(RequestStatus.SUCCESS)) {
+                    SW360Utils.removeReleaseVulnerabilityRelation(request.getDocumentId(), userFromRequest);
                 } else {
-                    sw360ComponentClient.updateReleaseFromModerationRequest(request.getReleaseAdditions(),
-                            request.getReleaseDeletions(), reviewer);
+                    throw new TException("Failed to delete the release with ID: " + request.getDocumentId());
```
GMishx, line 283: All `actionStatus` are processed at the end of the function with `return ModerationState.REJECTED;`. Why a special `throw` here?

Also, would be better to use the `SW360Exception` instead of `TException` from thrift.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
```diff
@@ -242,102 +245,93 @@ public Map<PaginationData, List<ModerationRequest>> getRequestsByState(User sw36
      */
     public ModerationState acceptRequest(ModerationRequest request, String moderatorComment, @NotNull User reviewer)
             throws TException {
-
+        ComponentService.Iface sw360ComponentClient = getThriftComponentClient();
+        ProjectService.Iface sw360ProjectClient = getThriftProjectClient();
+        LicenseService.Iface sw360LicenseClient = getThriftLicenseClient();
+        SPDXDocumentService.Iface sw360SPDXClient = getThriftSPDXDocumentClient();
+        DocumentCreationInformationService.Iface documentCreationInfoClient = getThriftDocumentCreationInfo();
+        PackageInformationService.Iface packageInfoClient = getThriftPackageInfo();
```
GMishx, line 253: Any reason to move all of the client outside the individual `case`?

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/Sw360ModerationRequestService.java
GMishx: @nikkuma7 If there is no specific reason, can we move the clients back to switch/case as in current code?

PR #2513, opened 2024-07-14
Remodel docker to new base

heliocastro:
Docker is now cleaner and depends only on the official Tomcat image, simplifying the whole process.

Docker-compose will open space in the future to embrace possible Keycloack and sw360 images, depending on the setup chosen.


KoukiHama (approved):
Good

PR #2514, opened 2024-07-14
Remove entirely OSGI bundles

heliocastro:
We no longer depend on OSGI for the removal of Liferay.
This presents an opportunity to remove the half-baked solution adopted before and start cleanly.


hoangnt2:
Code look good and working as expected

PR #2515, opened 2024-07-15
Versioning fix and liferay build references removal

heliocastro:


rest/authorization-server/pom.xml
```diff
@@ -17,7 +17,7 @@
     <parent>
         <artifactId>rest</artifactId>
         <groupId>org.eclipse.sw360</groupId>
-        <version>${revision}</version>
+        <version>18.99.1</version>
```
hoangnt2, line 20: I wonder why you don't set this to ${sw360.version}?

heliocastro: @hoangnt2  Finally someone did the right question.
Welcome to the marvellous world of maven :-)

```bash
❯ git diff
diff --git a/rest/authorization-server/pom.xml b/rest/authorization-server/pom.xml
index 3b5b71ce..678b17bd 100644
--- a/rest/authorization-server/pom.xml
+++ b/rest/authorization-server/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <artifactId>rest</artifactId>
         <groupId>org.eclipse.sw360</groupId>
-        <version>18.99.1</version>
+        <version>${sw360.version}</version>
     </parent>
 
     <properties>
```

Let's build then...

```bash
helio@citadel ~/code/helio/sw360 on  heliocastro/versioning_noliferay ●
❯ sw360_build 
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for org.eclipse.sw360:authorization-server:${sw360.version}: The following artifacts could not be resolved: org.eclipse.sw360:rest:pom:${sw360.version} (absent): org.eclipse.sw360:rest:pom:${sw360.version} was not found in https://jfrog.devstack.vwgroup.com/artifactory/public/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of devstack has elapsed or updates are forced and 'parent.relativePath' points at wrong local POM @ line 17, column 13
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.springframework.security.oauth:spring-security-oauth2:jar -> version ${spring-security-oauth2.version} vs 2.5.2.RELEASE @ line 113, column 21
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project org.eclipse.sw360:authorization-server:${sw360.version} (/Users/DHXBWM7/code/helio/sw360/rest/authorization-server/pom.xml) has 1 error
[ERROR]     Non-resolvable parent POM for org.eclipse.sw360:authorization-server:${sw360.version}: The following artifacts could not be resolved: org.eclipse.sw360:rest:pom:${sw360.version} (absent): org.eclipse.sw360:rest:pom:${sw360.version} was not found in https://jfrog.devstack.vwgroup.com/artifactory/public/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of devstack has elapsed or updates are forced and 'parent.relativePath' points at wrong local POM @ line 17, column 13 -> [Help 2]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
[ERROR] [Help 2] http://cwiki.apache.org/confluence/display/MAVEN/UnresolvableModelException
```

Now the proper explanation. Variables in maven parser are not resolved in parsing time, but only after the entire chain of project files are parsed. So even if registered before reach the next project build file, Maven not know it and fail.

Why worked before ? We were using "revision", which is always set, but was not the version effectively of the project, so marked down to see "something" until effectively start to build where it will then replaced for the set one.

Worst, sometimes during builds could create a mismatch due to some library remained in the .m2 cache, and not cleaned during the new build as maven not knows if that was the previous build version, so forcing us to do a forced mvn clean.

I intentionally did that way to allow us to use this:
[Maven Versioning Plugin](https://www.mojohaus.org/versions/versions-maven-plugin/usage.html)

There's an interesting automation plugin for this, but i will not go there yet, until proper testing.

hoangnt2: @heliocastro Well I understood the situation. Let's do this after everything works.

hoangnt2:
Changes look good to me

Issue #2544, opened 2024-08-07
Making clearing report download from fossology configurable


rudra-superrr:
### Description

Right now when a user sends clearing report to fossology, the report gets generated and sent back sw360 as an attachment in release which sets the clearing state of release to Scan Available.
On making it configurable, fossology won't be sending back the generated report to sw360 and the clearing state of release will be Under Clearing.

PR #2545, opened 2024-08-07
fix(clearingState): making fossology report download configurable.

rudra-superrr:
Issue: #2544 

### How To Test?
To make it configurable, set the variable as true in sw360.properties file.
--> disable.clearing.fossology.report.download=true

After making it configurable, sw360 will now upload the clearing report to fossology for scanning purpose only i.e., it won't generate and download the report back to sw360 as an attachment in release.
As a result it will change the state from Scan Available to Under Clearing.

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/common/utils/BackendUtils.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/common/utils/BackendUtils.java
@@ -20,6 +20,7 @@ public class BackendUtils {
     public static final Boolean MAINLINE_STATE_ENABLED_FOR_USER;
     public static final Boolean IS_BULK_RELEASE_DELETING_ENABLED;
     public static final Boolean IS_FORCE_UPDATE_ENABLED;
+    public static final Boolean DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD;
 
     static {
         loadedProperties = CommonUtils.loadProperties(BackendUtils.class, PROPERTIES_FILE_PATH);
@@ -27,6 +28,7 @@ public class BackendUtils {
         IS_BULK_RELEASE_DELETING_ENABLED = Boolean.parseBoolean(System.getProperty("RunBulkReleaseDeletingTest", loadedProperties.getProperty("bulk.release.deleting.enabled", "false")));
         IS_FORCE_UPDATE_ENABLED = Boolean.parseBoolean(
                 System.getProperty("RunRestForceUpdateTest", loadedProperties.getProperty("rest.force.update.enabled", "false")));
+        DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD = Boolean.parseBoolean(loadedProperties.getProperty("disable.clearing.fossology.report.download", "false"));
     }
 
     protected BackendUtils() {
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -225,8 +225,6 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
         long isrCountAfter = evaluateClearingStateForScanAvailable(releaseAfter);
         if (isrCountAfter > 0) {
             releaseAfter.setClearingState(ClearingState.SCAN_AVAILABLE);
-        } else {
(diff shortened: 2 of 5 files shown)


backend/src/src-fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
```diff
@@ -162,12 +164,14 @@ public ExternalToolProcess process(String releaseId, User user, String uploadDes
             handleUploadStep(componentClient, release, user, fossologyProcess, sourceAttachment, uploadDescription);
         } else if (FossologyUtils.FOSSOLOGY_STEP_NAME_SCAN.equals(furthestStep.getStepName())) {
             handleScanStep(componentClient, release, user, fossologyProcess);
-        } else if (FossologyUtils.FOSSOLOGY_STEP_NAME_REPORT.equals(furthestStep.getStepName())) {
-            handleReportStep(componentClient, release, user, fossologyProcess);
         }
-
         updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
 
+        if(BackendUtils.DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD && FossologyUtils.FOSSOLOGY_STEP_NAME_REPORT.equals(furthestStep.getStepName())) {
```
GMishx, line 170: Should it be the **not** case for `BackendUtils.DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD` as this step should run only if the configuration is set to `false`?

backend/src/src-fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
```diff
@@ -393,8 +397,10 @@ private void handleScanStep(Iface componentClient, Release release, User user,
             break;
         case DONE:
             // start report
-            fossologyProcess.addToProcessSteps(createFossologyProcessStep(user, FossologyUtils.FOSSOLOGY_STEP_NAME_REPORT));
-            handleReportStep(componentClient, release, user, fossologyProcess);
+            if(BackendUtils.DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD) {
```
GMishx, line 400: Should be a negated check as process should be added if configuration is false.

backend/src-common/src/main/java/org/eclipse/sw360/common/utils/BackendUtils.java
```diff
@@ -20,13 +20,15 @@ public class BackendUtils {
...
     public static final Boolean IS_FORCE_UPDATE_ENABLED;
+    public static final Boolean DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD;
 
     static {
         loadedProperties = CommonUtils.loadProperties(BackendUtils.class, PROPERTIES_FILE_PATH);
         MAINLINE_STATE_ENABLED_FOR_USER = Boolean.parseBoolean(loadedProperties.getProperty("mainline.state.enabled.for.user", "false"));
         IS_BULK_RELEASE_DELETING_ENABLED = Boolean.parseBoolean(System.getProperty("RunBulkReleaseDeletingTest", loadedProperties.getProperty("bulk.release.deleting.enabled", "false")));
         IS_FORCE_UPDATE_ENABLED = Boolean.parseBoolean(
                 System.getProperty("RunRestForceUpdateTest", loadedProperties.getProperty("rest.force.update.enabled", "false")));
+        DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD = Boolean.parseBoolean(loadedProperties.getProperty("disable.clearing.fossology.report.download", "false"));
```
GMishx, line 31: Please add the property to `backend/src-common/src/main/resources/sw360.properties` file as well

GMishx (changes requested):

backend/src-common/src/main/java/org/eclipse/sw360/common/utils/BackendUtils.java
rudra-superrr: added the property

backend/src/src-fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
rudra-superrr: made the changes

GMishx (dismissed):
Changes looks good.

GMishx (approved):
Tested, working as expected.

Issue #2566, opened 2024-08-21
Issue with API Endpoint for Listing Linked Projects (Transitive)


sameed20:
### Description
When making a GET request to this endpoint with the transitive=true parameter, the server returns an Internal Server Error (500).
![linkedProject_error](https://github.com/user-attachments/assets/d8f7181e-5f8c-4a40-ba6b-621cd762b509)

PR #2568, opened 2024-08-22
fix(rest): Add null check for linkedProject field to prevent Internal Server Error on GET request to fetch the linked projects of a project

sameed20:
closes: #2566


rudra-superrr:
Looks good to me.

PR #2600, opened 2024-09-04
fix(rest): Add null check for linkedProject field

GMishx:
The commit from #2568 is in latest main but the changes are missing. Adding back the changes with cherry-pick.


sameed20:
@GMishx Approved.

Issue #2604, opened 2024-09-10
Endpoint to update a vendor


rudra-superrr:
### Description

Patch request for vendor.

PR #2606, opened 2024-09-10
feat(rest): endpoint to update a vendor.

rudra-superrr:
Issue: #2604 

Description: endpoint to update a vendor.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
@@ -16,6 +16,8 @@ import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import org.apache.thrift.TException;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
+import org.eclipse.sw360.datahandler.thrift.RequestStatus;
+import org.eclipse.sw360.datahandler.thrift.users.User;
 import org.eclipse.sw360.datahandler.thrift.vendors.Vendor;
 import org.eclipse.sw360.rest.resourceserver.core.HalResource;
 import org.eclipse.sw360.rest.resourceserver.core.RestControllerHelper;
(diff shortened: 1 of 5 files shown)


GMishx:
@deo002 can you please test this PR?

deo002:
image
Cannot update url

rudra-superrr:
Hi @deo002 , can you try now.

deo002:
> Hi @deo002 , can you try now.

Thanks for quick response.
image
When I put null value for shortname, it gives 200 with message that Vendor has been updated successfully. When I do get request for that vendor, it is not updated. Ideally, it shouldn't be updated, but an error should be thrown instead of 200.

deo002:
image
Issue still persists

rudra-superrr:
> image Issue still persists

If user updates a single field, like shortName then the other two fields will be taken as null by default. Therefore in the above ss it will update shortname only and response will be 200 even if it says fullName is null because by default also fullName will be taken as null and it wont update the fullName field of the vendor. On the other hand if you enter a single field as null in body then it will give 400 status code because all the three fields will be considered as null.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
```diff
@@ -129,6 +131,33 @@ public ResponseEntity<?> createVendor(
...
+    @PreAuthorize("hasAuthority('WRITE')")
+    @RequestMapping(value = VENDORS_URL + "/{id}", method = RequestMethod.PATCH)
+    public ResponseEntity<?> updateVendor(
+            @Parameter(description = "The id of the vendor")
+            @PathVariable("id") String id,
+            @Parameter(description = "The vendor to be updated.")
+            @RequestBody Vendor vendor
+    ) {
+        User sw360User = restControllerHelper.getSw360UserFromAuthentication();
+        if (vendor.getFullname() == null && vendor.getShortname() == null && vendor.getUrl() == null) {
```
GMishx, line 148: Shouldn't these conditions be `||` instead of `&&` to prevent reading and updating `null` values for some fields?

deo002:
Tested, working as expected!

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
rudra-superrr: If I use || condition then it will return bad request whenever user wants to update specific fields.

GMishx: Make sense

GMishx (approved):
Changes looks good.

PR #2609, opened 2024-09-11
Unify backend source code

heliocastro:
Reference: https://github.com/eclipse-sw360/sw360/discussions/2602


KoukiHama:
Thank you for big works. I will review it.

heliocastro:
@KoukiHama @GMishx Can we go ahead on this, we already have one conflict and will be only get worse ahead

GMishx:
@heliocastro The changes looks good to me and I just tested the branch and the `/api/components` endpoint is now throwing `500` error. Other endpoints are working flawlessly, I also tested them with the frontend project and Swagger.
Just for `http://localhost:8080/resource/api/components`, I get following error but nothing in logs.
```json
{
  "timestamp": "2024-09-20T05:54:24.790291950Z",
  "status": 500,
  "error": "Internal Server Error",
  "message": "HTTP Response code: 404"
}
```

Other than this break, I approve the PR.

Also, please understand we need to move carefully (definitely not mean slowly) for such major changes. More so, when the build time is down from 120 seconds to 45 seconds :wink:

I do understand your pain with constant conflict resolution, but please also understand this PR will cause merge conflict for 50 other PRs at the same time.

heliocastro:
> @heliocastro The changes looks good to me and I just tested the branch and the `/api/components` endpoint is now throwing `500` error. Other endpoints are working flawlessly, I also tested them with the frontend project and Swagger. Just for `http://localhost:8080/resource/api/components`, I get following error but nothing in logs.
> 
> ```json
> {
>   "timestamp": "2024-09-20T05:54:24.790291950Z",
>   "status": 500,
>   "error": "Internal Server Error",
>   "message": "HTTP Response code: 404"
> }
> ```
> 
> Other than this break, I approve the PR.
> 
> Also, please understand we need to move carefully (definitely not mean slowly) for such major changes. More so, when the build time is down from 120 seconds to 45 seconds 😉
> 
> I do understand your pain with constant conflict resolution, but please also understand this PR will cause merge conflict for 50 other PRs at the same time.

Yes, but will always have <number> PRs, so if we go in this mode, we will never do the changes. I just rebased again with CycloneDX, and this will go forever, so we need merge it, and fix the PR, but not wit the PR's because wealways come some more, and more

GMishx:
> Yes, but will always have PRs, so if we go in this mode, we will never do the changes. I just rebased again with CycloneDX, and this will go forever, so we need merge it, and fix the PR, but not wit the PR's because wealways come some more, and more

I agree to the point, that's why I approved it. However, we cannot merge a PR which breaks existing functionality, that too as basic as `/components` endpoint, not something obscured.

Once the components endpoint is fixed, I am good with merging this PR.

I guess @KoukiHama wanted to review it as well.

KoukiHama:
I'm currently in the process of reviewing. I've gone through about half of it and will finish reviewing everything by the weekend.

KoukiHama:
> I agree to the point, that's why I approved it. However, we cannot merge a PR which breaks existing functionality, that too as basic as /components endpoint, not something obscured.

> > Once the components endpoint is fixed, I am good with merging this PR.

I agree with this approach as well. While it's important to make progress, we shouldn't merge the PR until the basic functionality is fixed. Once the error is resolved, we can proceed with the merge.

libraries/importers/src/test/java/org/eclipse/sw360/importer/ComponentAndAttachmentAwareDBTest.java
KoukiHama, line 1: Could these comment-outs in the file be removed?

docker-compose.yml
KoukiHama, line 1: Could these comment-outs in the file be removed?

KoukiHama: L35-36

KoukiHama (commented):
Alomost all ok to me

heliocastro:
> @heliocastro The changes looks good to me and I just tested the branch and the `/api/components` endpoint is now throwing `500` error. Other endpoints are working flawlessly, I also tested them with the frontend project and Swagger. Just for `http://localhost:8080/resource/api/components`, I get following error but nothing in logs.
> 
> ```json
> {
>   "timestamp": "2024-09-20T05:54:24.790291950Z",
>   "status": 500,
>   "error": "Internal Server Error",
>   "message": "HTTP Response code: 404"
> }
> ```
> 
> Other than this break, I approve the PR.
> 
> Also, please understand we need to move carefully (definitely not mean slowly) for such major changes. More so, when the build time is down from 120 seconds to 45 seconds 😉
> 
> I do understand your pain with constant conflict resolution, but please also understand this PR will cause merge conflict for 50 other PRs at the same time.

@GMishx Fixed. Components portlet was not properly exported

GMishx:
Hi @heliocastro,

Thank you for your efforts in addressing the endpoint issue. Unfortunately, we haven't had the opportunity to review or test the changes yet. It appears that the PR was self-merged, which is generally not recommended for community projects.

Could we please discuss the topic of self-merging during today's community call to ensure we're all on the same page?

/cc @KoukiHama @ag4ums @arunazhakesan

KoukiHama:
FYI : https://eclipse.dev/sw360/docs/development/dev-dod-and-style/

Issue #2622, opened 2024-09-25
Enhance the built-in CDX importer of SW360 to sanitize VCS URLs of non-GitHub domains


sameed20:
### Description
These are the additional domain for which we want to extend the VCS sanitization feature.

gitlab.com
bitbucket.org
cs.opensource.google.com
go.googlesource.com
pypi.org

Findings:

1. **GITLAB**, **BITBUCKET**, **GOOGLE OPEN SOURCE** shares the common pattern for vcs which is similar to that of github vcs pattern only consisting of manufacturer followed by comp_name

* https://gitlab.com/manufacturer/comp_name
* https://bitbucket.org/manufacturer/comp_name
* https://cs.opensource.google/manufacturer/comp_name

2. **go.googlesource.com** has the common pattern as domain followed by comp name

* https://go.googlesource.com/comp_name

3. For **pypi.org** the observed pattern is

* https://pypi.org/project/comp_name

PR #2623, opened 2024-09-25
feat(importCDX): enhance CDX importer to sanitize VCS URLs for non-GitHub domains

sameed20:
Issue: #2622

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
@@ -113,6 +113,16 @@ public class CycloneDxBOMImporter {
     private final User user;
     private final AttachmentConnector attachmentConnector;
 
+    // Map of supported hosts and base URL formats
+    private static final Map<String, String> VCS_HOSTS = Map.of(
+            "github.com", "https://github.com/%s/%s",
+            "gitlab.com", "https://gitlab.com/%s/%s",
+            "bitbucket.org", "https://bitbucket.org/%s/%s",
+            "cs.opensource.google", "https://cs.opensource.google/%s/%s",
+            "go.googlesource.com", "https://go.googlesource.com/%s",
+            "pypi.org", "https://pypi.org/project/%s"
+    );
+
     public CycloneDxBOMImporter(ProjectDatabaseHandler projectDatabaseHandler, ComponentDatabaseHandler componentDatabaseHandler,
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
```diff
@@ -987,16 +987,115 @@ public String getComponetNameById(String id, User user) throws SW360Exception {
...
+        }else if(vcs.contains("gitlab.com")){
+            // GitLab repository URL Format: https://gitlab.com/supplier/name
+            vcs = "https://" + vcs.substring(vcs.indexOf("gitlab.com")).trim();
+            try{
+                URI uri = URI.create(vcs);
+                String[] urlParts = uri.getPath().split("/");
+                if (urlParts.length >= 3) {
+                    String firstSegment = urlParts[1];
+                    String secondSegment = urlParts[2].replaceAll("\\.git.*", "").replaceAll("#.*", "");
+                    vcs = "https://gitlab.com/" + firstSegment + "/" + secondSegment;
```
GMishx, line 1016: This will fail for repositories like https://gitlab.com/Orange-OpenSource/k8s-tz/applications/nexus
It will be changed to https://gitlab.com/Orange-OpenSource/k8s-tz which is technically wrong?

backend/common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
```diff
@@ -987,16 +987,115 @@ public String getComponetNameById(String id, User user) throws SW360Exception {
...
+                    String firstSegment = urlParts[1];
+                    String secondSegment = urlParts[2].replaceAll("\\.git.*", "").replaceAll("#.*", "");
+                    vcs = "https://bitbucket.org/" + firstSegment + "/" + secondSegment;
+                    return vcs;
+                } else {
+                    log.error("Invalid Bitbucket repository URL: " + vcs);
+                }
+            } catch (IllegalArgumentException e) {
+                log.error("Invalid URL format: {}", vcs, e);
+            }
```
GMishx, line 1040: There is lot of code repetition for github, gitlab and bitbucket. Can this be separated out as a function?

backend/common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java
GMishx: Note, there can be more nested groups in GitLab. I just choose the first example I saw in the wild.

GMishx (changes requested):

PR #2629, opened 2024-09-30
feat: Add default admin user if database is empty

heliocastro:
Attempt to create a default Admin user if SW360 was bootstraped.


backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
```diff
@@ -47,16 +52,54 @@ public class UserHandler implements UserService.Iface {
 
     private static final Logger log = LogManager.getLogger(UserHandler.class);
     private static final String EXTENSION = ".log";
+    private final PasswordEncoder passwordEncoder = new BCryptPasswordEncoder(
+        12
```
smrutis1, line 56: In other places the hashing rounds are set to default (10). May be we have to keep the same here for consistency or else we need to change in other places as well.

backend/users/pom.xml
```diff
@@ -33,5 +33,9 @@
             <artifactId>backend-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-core</artifactId>
```
smrutis1, line 38: May be we can use just the crypto from Spring
```
<dependency>
    <groupId>org.springframework.security</groupId>
     <artifactId>spring-security-crypto</artifactId>
</dependency>
```

backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
heliocastro: I simplified it and made it default as the rest of the places.

heliocastro:
@smrutis1 Can you review it again ?

smrutis1:
> @smrutis1 Can you review it again ?

@heliocastro Will complete it today. Looking into the other possibility instead of creating in `UserHandler` directly. May be somewhere from REST while bootstrapping the application.

backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
```diff
@@ -44,13 +50,40 @@ public class UserHandler implements UserService.Iface {
...
     private UserDatabaseHandler db;
     private ReadFileDepartmentConfig readFileDepartmentConfig;
 
     public UserHandler() throws IOException {
-        db = new UserDatabaseHandler(DatabaseSettings.getConfiguredClient(), DatabaseSettings.COUCH_DB_USERS);
+        db = new UserDatabaseHandler(DatabaseSettings.getConfiguredClient(),
+                DatabaseSettings.COUCH_DB_USERS);
         readFileDepartmentConfig = new ReadFileDepartmentConfig();
+
+        // Create admin user if not in database yet
```
smrutis1, line 63: Just a suggestion to add it inside `Sw360UserService.java` or `UserController.java` 
```
   @PostConstruct
    private void initializeDefaultAdminUser() {
        createAdminUserIfNotExists();
    }

    private void createAdminUserIfNotExists() {
        // Create admin user if not in database yet
        List<User> users = getAllUsers();
        if (users.isEmpty()) {
            Optional<String> COUCHDB_ADMIN_PASSWORD = Optional.ofNullable(
                    System.getenv("COUCHDB_ADMIN_PASSWORD") != null
                            ? System.getenv("COUCHDB_ADMIN_PASSWORD")
                            : "sw360fossie"
            );
            User admin = new User();
            admin.setEmail("setup@sw360.org");
            admin.setFullname("SW360 Admin");
            admin.setGivenname("SW360");
            admin.setLastname("Admin");
            admin.setDepartment("SW360");
            admin.setPassword(COUCHDB_ADMIN_PASSWORD.get());
            admin.setUserGroup(UserGroup.ADMIN);
            String encodedPassword = passwordEncoder.encode(
                    admin.getPassword()
            );
            admin.setPassword(encodedPassword);
            log.info("No users found. Creating default administrator user.");
            try {
                addUser(admin);
            } catch (TException e) {
                log.atError().withThrowable(e).log("Error creating admin user");
            }
        }
```

heliocastro: No dice. Thrift is set in a way that waits all services be available before enable communication, and we're trying to use it befor even been enabled.
Tests show that the startup of resource server becomes stuck.
So, the current imlpementation is the best until we in the future remove Thrift.

heliocastro:
@smrutis1 Ping ? Double-checked; the current solution is the best in a short time without starting to play over thrift. 
We can revise in the future when we start to remove thrift, but for now, it needs to be this one.

smrutis1:
> @smrutis1 Ping ? Double-checked; the current solution is the best in a short time without starting to play over thrift. We can revise in the future when we start to remove thrift, but for now, it needs to be this one.

Although its not recommended to perform any database operation inside constructor, But as mentioned it's ok for now, we can revise it later.

GMishx:
@smrutis1 can you then approve this PR and we can merge it?

smrutis1 (approved):
Good to go for now

PR #2631, opened 2024-10-01
feat(couchdb): Enable use of latest CouchDB with nouveau

heliocastro:


GMishx (approved):
Changes looks good. Working as expected.

PR #2633, opened 2024-10-08
fix(REST): Patch Release is causing the clearing state to be updated to NEW even if a Clearing is existing

afsahsyeda:

**Description:**
1. The new clearing state of a release was not being rendered after a successful PATCH. It was rereferring to the older clearing state.
2. If an empty array of attachments was passed in the request body, the clearing state of the release would get updated to NEW while an approved CLX/CRT was still present. 

**These issues are fixed by**

-  Fetching the updated release before creating a HalResource out of it.
-  Setting the new clearing state _after_ the approved attachments are retained.

### How To Test?
1. Pass an empty array of attachments and try to PATCH while the release has approved CRT/CLX attachments. The attachments would be retained and clearing state will still be "APPROVED". (would be NEW_CLEARING before the fix)
2. Remove the approval from both all the attachments and then pass an empty array of attachments and PATCH the release. In this case, all the attachments would be removed and clearing state would change to NEW_CLEARING. ( would still be APPROVED before the fix)
image

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -1152,12 +1152,13 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
                 copyFields(actual, release, immutableFields);
 
-                autosetReleaseClearingState(release, actual);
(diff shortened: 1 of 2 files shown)


afsahsyeda:
@sameed20 Please perform a general test on this.

sameed20:
Test Successful.

GMishx (approved):
Changes looks good.

PR #2642, opened 2024-10-16
feat(rest): Add size parameter to clearing request.

akshitjoshii:
Add a new field **size** to the clearing request which determines the size of the clearing request based on the number of **open releases** in the project for which the CR is created. 

Size based on the table below : 
1- 20 = **Very Small**
21- 50 = **Small**
50-75 = **Medium**
75-150 = **Large**
150 and above = **Very large**

The CR size should be dynamic. if release are uploaded after CR creation, it should reflect in the `CR Size`.

image

### DIFF
--- a/backend/moderation/src/main/java/org/eclipse/sw360/moderation/ModerationHandler.java
+++ b/backend/moderation/src/main/java/org/eclipse/sw360/moderation/ModerationHandler.java
@@ -19,6 +19,7 @@ import org.eclipse.sw360.datahandler.thrift.PaginationData;
 import org.eclipse.sw360.datahandler.thrift.RemoveModeratorRequestStatus;
 import org.eclipse.sw360.datahandler.thrift.RequestStatus;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
+import org.eclipse.sw360.datahandler.thrift.ClearingRequestSize;
 import org.eclipse.sw360.datahandler.thrift.components.Component;
 import org.eclipse.sw360.datahandler.thrift.components.Release;
 import org.eclipse.sw360.datahandler.thrift.licenses.License;
@@ -356,6 +357,14 @@ public class ModerationHandler implements ModerationService.Iface {
         assertUser(user);
 
         handler.updateClearingRequestForChangeInProjectBU(crId, businessUnit, user);
+    }
+
+    @Override
+    public void updateClearingRequestForChangeInClearingSize(String crId, ClearingRequestSize size) throws TException {
+        assertId(crId);
+        assertNotNull(size);
+
+        handler.updateClearingRequestForChangeInClearingSize(crId, size);
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


sameed20:
Testing this PR.

sameed20:
@akshitjoshii When increasing the number of releases in a project, the size of the CR should increase accordingly. However, I observed that when fetching a CR by ID from an endpoint, the size parameter in the response does not reflect the updated size value, even though the size field in the database updates as expected.

akshitjoshii:
@sameed20 , made the changes. Pls check

sameed20:
Test successful.

afsahsyeda (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -1582,4 +1590,18 @@ public void addEmbeddedReleaseDetails(HalResource<ClearingRequest> halClearingRe
...
     }
+
+    public ClearingRequest updateCRSize(ClearingRequest clearingRequest, Project project, User sw360User) throws TException {
+        int openReleaseCount = SW360Utils.getOpenReleaseCount(project.getReleaseClearingStateSummary());
+        ClearingRequestSize currentSize = SW360Utils.determineCRSize(openReleaseCount);
+        ClearingRequestSize initialSize = clearingRequest.getClearingSize();
+        if(initialSize == null) return clearingRequest;
+        if(!initialSize.equals(ClearingRequestSize.VERY_LARGE)) {
+            int limit = CLEARING_REQUEST_SIZE_MAP.get(initialSize);
+            if(openReleaseCount > limit){
```
GMishx, line 1601: @akshitjoshii Instead of putting limit map at 2 places (here and in `SW360Utils::determineCRSize()`), can we instead compare `currentSize` and `initialSize`? They will not be equal if there is an size update and thus the DB should be updated, as per my understanding.

This will help reduce number of places to be updated if the size limits are updated in future.

GMishx (approved):
Changes looks good.

GMishx:
@afsahsyeda can you please review the changes for auto merge of this branch?

afsahsyeda (dismissed):

backend/moderation/src/main/java/org/eclipse/sw360/moderation/ModerationHandler.java
```diff
@@ -13,12 +13,7 @@
 import org.apache.thrift.TException;
 import org.eclipse.sw360.datahandler.common.DatabaseSettings;
 import org.eclipse.sw360.datahandler.db.ModerationSearchHandler;
-import org.eclipse.sw360.datahandler.thrift.Comment;
-import org.eclipse.sw360.datahandler.thrift.ModerationState;
-import org.eclipse.sw360.datahandler.thrift.PaginationData;
-import org.eclipse.sw360.datahandler.thrift.RemoveModeratorRequestStatus;
-import org.eclipse.sw360.datahandler.thrift.RequestStatus;
-import org.eclipse.sw360.datahandler.thrift.SW360Exception;
+import org.eclipse.sw360.datahandler.thrift.*;
```
heliocastro, line 16: In the future, remember that best practices is no do global imports.

heliocastro (approved):
Approved but with some remarks with code style.

heliocastro:
@akshitjoshii Please resolve the merge conflict and if possible revert the glob on imports.

akshitjoshii:
@heliocastro, merge conflicts resolved and fixed global import.

GMishx (approved):

PR #2647, opened 2024-10-17
chore: Add maven validation on build

heliocastro:


GMishx:
Build should be fixed by #2648

PR #2649, opened 2024-10-17
chore(maven): segregate war and jar deploy dirs

GMishx:

Since now we do not have OSGi modules, in bare metal install, there is no need to put the jar files. Everything is packages as war and put in tomcat/webapps.

Set the default value of `jars.deploy.dir` to empty to prevent their copy but still allow Docker builds to work as it is passed as parameter to maven.

At the same time, add `listener.deploy.dir` (with default empty value) for copying keycloak related dependencies and jars to its provider directory.

### Suggest Reviewer
@heliocastro @smrutis1

### How To Test?
Install sw360 with `mvn clean install -DskipTests -Dbase.deploy.dir=/opt/apache-tomcat-11.0.0/ -Dlistener.deploy.dir=/opt/keycloak-24.x.x/providers -P deploy` and check if:
1. war files are copied to `/opt/apache-tomcat-11.0.0/webapps`
2. jar files are copied to `/opt/keycloak-24.x.x/providers`

### DIFF
--- a/keycloak/README.md
+++ b/keycloak/README.md
@@ -73,21 +73,22 @@ sudo ./kc.sh start  --log="console,file" --hostname-strict-backchannel=false --h
 
 ## Build the Backend:
 
-* Build the SW360 backend code using Maven, `mvn clean install -DskipTests`
-* Copy the generated WAR files to the webapps folder of Apache Tomcat, `cp $(find . -name "*.war")  /opt/apache-tomcat-10.x.x/webapps`
+* Build the SW360 backend code using Maven,
+  `mvn clean install -DskipTests -Dbase.deploy.dir=/opt/apache-tomcat-11.x.x/ -Dlistener.deploy.dir=/opt/keycloak-24.x.x/providers -P deploy`
 * Start the Apache Tomcat server.
 
 ## Keycloak Providers and Libraries:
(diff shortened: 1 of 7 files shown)


keycloak/README.md
```diff
@@ -73,21 +73,22 @@ sudo ./kc.sh start  --log="console,file" --hostname-strict-backchannel=false --h
 
 ## Build the Backend:
 
-* Build the SW360 backend code using Maven, `mvn clean install -DskipTests`
-* Copy the generated WAR files to the webapps folder of Apache Tomcat, `cp $(find . -name "*.war")  /opt/apache-tomcat-10.x.x/webapps`
+* Build the SW360 backend code using Maven,
+  `mvn clean install -DskipTests -Dbase.deploy.dir=/opt/apache-tomcat-10.x.x/ -Dlistener.deploy.dir=/opt/keycloak-24.x.x/providers -P deploy`
```
heliocastro, line 77: tomcat-11 :-)

heliocastro (changes requested):
All good, even besides we could think in baremetal use the shared. library process as well, to simplify the way we understand the deployment.
I added a comment on tell that 11 is default only :-)

GMishx:
Finally, got the correct GPG keys :-)

heliocastro (approved):
LGTM with signed commit :-)

PR #2687, opened 2024-11-05
Maven updates

heliocastro:
Remove pre-commit checkstyle in favour of maven solution
   
Checkstyle used in pre-commit rule depends on docker based setup, which
is not intended to enforce to al developers in their environments.
    
Using maven checkstyle solution not enforced is an intermediary solution
to start improve the source quality until we get the enforce status.


GMishx (approved):
Looks good.

Issue #2695, opened 2024-11-11
Attachments getting deleted while updating/adding component


rudra-superrr:
### Description

Attachments getting deleted while updating/adding component's external ids field through rest (making Patch request).

PR #2696, opened 2024-11-11
fix(rest): component attachment deletion while updating externalIds

rudra-superrr:
Issue: #2695 

Description: After these changes, attachment won't get deleted when component's externalIds field is updated through rest.


GMishx (approved):
Changes looks good. Tested, working as expected.

PR #2706, opened 2024-11-14
Adjust Maven dependency declarations

heliocastro:
- Move all versions from keycloack pom to superpom
- Adjust version or spring.security to be unique
- Replaced jaxb javax to jakarta
- Removed duplicate maven-sources entry

### DIFF
--- a/keycloak/user-storage-provider/pom.xml
+++ b/keycloak/user-storage-provider/pom.xml
@@ -2,8 +2,8 @@
 ~ SPDX-FileCopyrightText: © 2024 Siemens AG
 ~ SPDX-License-Identifier: EPL-2.0
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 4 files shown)


keycloak/user-storage-provider/pom.xml
```diff
@@ -24,49 +24,37 @@
...
         </dependency>
         <dependency>
             <groupId>org.springframework.security</groupId>
             <artifactId>spring-security-crypto</artifactId>
-            <version>${spring.security.crypto}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents.core5</groupId>
             <artifactId>httpcore5</artifactId>
-            <version>${httpcore5.version}</version>
```
GMishx, line 69: Won't it be a good idea to still limit the dependency to a specific version in the project's pom.xml instead of just relying on the super pom.xml?
I don't have much idea what Maven will give preference to..

heliocastro: It is limited by the superpom version.  That's why we should never add version explicitly on the poms.

Inside  `<dependencyManagement><depencies> `
```xml
   <dependency>
        <groupId>org.apache.httpcomponents</groupId>
        <artifactId>httpcore</artifactId>
        <version>${httpcore.version}</version>
      </dependency>
```

So, one single place that tell for every entry the same. 
The version ( and scopes when needed ) are rules in superpom. You declare the dependency only with group and artifact id

GMishx (approved):
LGTM

PR #2750, opened 2024-11-26
fix(deps): Deprecate old commmons-lang library

heliocastro:
Old apache-commons-lang was puled down by nouveau-handler, and misleading outdated code in different sources.
All code now is adjusted to commons-lang3 or commons-text where needed.


GMishx:
@heliocastro This PR was changed from `WIP` (thus not reviewed) to `ready` and self-merged without any review from anyone other than yourself. This does not follow the guidelines set for the SW360 project: https://eclipse.dev/sw360/docs/development/dev-dod-and-style/

We have discussed the same multiple times as a bad practice.

heliocastro:
Total my fault. I was in transit with plenty of tabs and was merging one
after another.
Sorry for that, I verified again to check if was harmful but is fine.
Promise will not happens again, and I will try put operation from frontend
and backend more separated.

PR #2751, opened 2024-11-26
Fix(User): Fix XSS vulnerability for revoke token endpoint

hoangnt2:
Fix XSS vulnerability for revoke token endpoint

CodeQL warning: https://github.com/eclipse-sw360/sw360/security/code-scanning/87

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/user/UserController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/user/UserController.java
@@ -23,7 +23,6 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.commons.lang.RandomStringUtils;
-import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.thrift.TException;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.common.SW360Constants;
@@ -368,7 +367,7 @@ public class UserController implements RepresentationModelProcessor<RepositoryLi
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
 
         if (!userService.isTokenNameExisted(sw360User, tokenName)) {
-            return new ResponseEntity<>("Token not found: " + StringEscapeUtils.escapeHtml(tokenName), HttpStatus.NOT_FOUND);
+            return new ResponseEntity<>("Token not found", HttpStatus.NOT_FOUND);
         }
 
         sw360User.getRestApiTokens().removeIf(t -> t.getName().equals(tokenName));


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/user/UserController.java
```diff
@@ -23,7 +23,6 @@
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.commons.lang.RandomStringUtils;
-import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.thrift.TException;
```
heliocastro, line 26: Change for org.apache.commons.text.StringEscapeUtils

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/user/UserController.java
```diff
@@ -368,7 +367,7 @@ public ResponseEntity<String> revokeUserRestApiToken(
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
 
         if (!userService.isTokenNameExisted(sw360User, tokenName)) {
-            return new ResponseEntity<>("Token not found: " + StringEscapeUtils.escapeHtml(tokenName), HttpStatus.NOT_FOUND);
+            return new ResponseEntity<>("Token not found", HttpStatus.NOT_FOUND);
```
heliocastro, line 370: Change for escapeHtml4

heliocastro (changes requested):
@hoangnt2 I started a branch deprecating commons-lang to commons-lang3.
This method you used moved to commons-text

hoangnt2:
> @hoangnt2 I started a branch deprecating commons-lang to commons-lang3. This method you used moved to commons-text

Escape HTML cannot resolve XSS problem, CodeQL still warns so I decided to remove **tokenName** from response. I also want to know, is there a way to run this check locally. I want to know if my fix can solve the problem before creating a PR

heliocastro:
> > @hoangnt2 I started a branch deprecating commons-lang to commons-lang3. This method you used moved to commons-text
> 
> @heliocastro Escape HTML cannot resolve XSS problem, CodeQL still warns so I decided to remove **tokenName** from response. I also want to know, is there a way to run this check locally. I want to know if my fix can solve the problem before creating a PR

Ok, for that part is fine, agreed. But remove the function unfortunately solve a small portion. We have this pervasive on the code. 
We need think in a sanitizer library

hoangnt2:
@heliocastro OK, I will try again with escapeHtml4. Hopefully, it can solve the problem.

GMishx:
Since #2750 is merged, this PR can be updated as well.

GMishx:
GitHub marked the issue fixed in https://github.com/eclipse-sw360/sw360/commit/325cf0ef5ef38f542edd16485a4186fd22671507

Thus this PR is no longer required.

PR #2709, opened 2024-11-15
fix(openapi)!: add health endpoint to openapi

GMishx:
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

> Please provide a summary of your changes here.

Add the missing docs for `/health` endpoint in OpenAPI docs. To make the changes work with Swagger UI, the `/health` endpoint had to be moved making it a breaking change.

**BREAKING CHANGE:** Move the health endpoint from `/resource/health` to `/resource/api/health` for OpenAPI documentations to be easy to use.

### How To Test?
Check OpenAPI docs for the health endpoint.

PR #2766, opened 2024-12-02
feat(config): unify properties files

### LINKED ISSUE #2765: Endpoint to expose all configs needed for the UI to be in-sync with backend (N)
### Description

Need an endpoint to expose configurations from `sw360.properties` the front-end to be in sync with the backend.

GMishx:

1. The PR unifies all the different `sw360.properties` resources into a single file under datahandler.
2. The load of these properties are unified `org.eclipse.sw360.datahandler.common.SW360Constants` where applicable.
3. Additional classes are created as needed under the `org.eclipse.sw360.datahandler.common` package extending the `SW360Constants` class.
4. New actuator for exposing the constants from these classes. The actuator deliberately exposes only the configs needed the the UI to work. The backend only configs are not exposed. This can be changed if needed.

Closes #2765

### Suggest Reviewer
@smrutis1

### How To Test?
Check the `/api/config` endpoint.

### Checklist
Must:

This PR needs changes from #2709 and should be merged after it.


heliocastro (approved):
Sounds good. I think need a second pair of eyes to review it.

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Constants.java
```diff
@@ -128,6 +145,78 @@ public class SW360Constants {
...
+    public static final Boolean REST_API_TOKEN_GENERATOR_ENABLE;
+    public static final Boolean REST_API_WRITE_ACCESS_TOKEN_ENABLE;
+    public static final Set<String> DOMAIN;
+    public static final String REST_REPORT_FILENAME_MAPPING;
+    public static final String REST_JWKS_ISSUER_URL;
+    public static final String REST_JWKS_ENDPOINT_URL;
+    public static final Boolean REST_IS_JWKS_VALIDATION_ENABLED;
+    public static final String REST_SERVER_PATH_URL;
+
+    // UI Portlet Constants
```
smrutis1, line 185: I was just wondering if UI constants are required anymore or it can be sanitized?

heliocastro:
@GMishx Question: Do we need a property files at all ?
Can we simply move everything to the database and create default entries and rest api to update.
This would simplify a lot.

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Constants.java
GMishx: I will recheck once which constants from this section are used by backend and which are not. The ones not used can be moved to the frontend project.

My initial thought was to have minimal changes the existing `sw360.properties` files. But yes, the best approach would be to separate the backend and frontend configs.

GMishx:
> @GMishx Question: Do we need a property files at all ? Can we simply move everything to the database and create default entries and rest api to update. This would simplify a lot.

That would be the ultimate goal. If anyone is interested in doing can work on it. Till we have a working release with the new front-end and back-end, it would be a low priority imo.

But the solution from this PR would be good for the front-end to at least have idea how the back-end is configured.

hoangnt2:
@heliocastro @GMishx, I am currently investigating and implementing a POC for storing some configurations in CouchDB.

My idea is to store configurations in the sw360config database in CouchDB (similar to the FOSSology configuration). This approach will allow users to modify them through the GUI. However, I don't intend to move all configurations from the properties file into the database, as some are related to migration scripts and the connections between back-end services.

I will create a PR for this in a few weeks.

GMishx:
Putting this PR on hold in favor of #2815

GMishx:
After #2815 I will update this PR to remove the properties from the file as they are not needed anymore.

GMishx:
Moved the properties to the DB and created new containers. Check #3191

PR #2815, opened 2024-12-12
Moving some configurations from properties to CouchDB 

hoangnt2:
## Idea:
- Move some configuration properties from sw360.properties file to CouchDB. These properties can be changed without breaking the application
- The configuration properties related to **back-end connection** (e.g: backend.url) or **migration scripts** (e.g: enable.flexible.project.release.relationship) will not be moved to CouchDB
- The configuration properties will be stored in **sw360config** database as below
image

- The configuration properties will be save and synchronize between memory (Java variable) and CouchDB to avoid query to much into CouchDB. When application is started the configuration properties will be loaded into memory and they also will be updated when there is any update request

## List of properties can be saved to CouchDB
```properties
spdx.document.enabled=false
component.visibility.restriction.enabled=false
licenseinfo.spdxparser.use-license-info-from-files=false
mainline.state.enabled.for.user=false
enable.attachment.store.to.file.system=false
attachment.delete.no.of.days=30
attachment.store.file.system.location=/opt/sw360tempattachments
auto.set.ecc.status=false
send.project.spreadsheet.export.to.mail.enabled=true
send.component.spreadsheet.export.to.mail.enabled=true
bulk.release.deleting.enabled=true
disable.clearing.fossology.report.download=true
rest.force.update.enabled=true
sbom.import.export.access.usergroup=USER
sw360.tool.name=SW360
sw360.tool.vendor=Eclipse Foundation
package.portlet.enabled=true
package.portlet.write.access.usergroup=USER
admin.private.project.access.enabled=true
```

## Remaining properties (Some of them are related to Liferay and no longer used in the current back-end.)
[sw360.properties.txt](https://github.com/user-attachments/files/19386091/sw360.properties.txt)

#### 1. Get all configuration properties (from both CouchDB and sw360.properties file)
- URL: http://localhost:8080/resource/api/configurations
- Method: GET
- Response:
```
{
    "spdx.document.enabled": "true",   // Moved properties in CouchDB
    "enable.flexible.project.release.relationship": "true" // Remaining properties in sw360.properties
}
```
#### 2. Update configuration properties in CouchDB (only moved properties are allowed to change)
- URL: http://localhost:8080/resource/api/configurations
- Method: PATCH
- Body:
```
{
    "spdx.document.enabled": "false"
}
```
- Response:
```
"Configurations are updated successfully"
```

- If configuration properties related to **back-end connection** or **migration scripts** are requested to update, return 400 error code with message:
```
"Invalid config: [${properties name} : ${value}]"
```
e.g:
```
"Invalid config: [enable.flexible.project.release.relationship : true]"
```

Demonstrate configuration changes from the GUI
change-configuration-from-GUI

Please take a look at this PR and feel free to add your comment. If you have any question please let me know


hoangnt2:
@heliocastro @GMishx, please add your comments (I haven't fixed the unit test yet so the CI fails)

GMishx:
:rocket: The idea looks great. In the same endpoint, we can list both, configs from properties file and from DB. This also allows quick update of config in running system, as current way requires restart of tomcat server to update the config constants.

Its a go ahead from me :+1:

@smrutis1 , what are your thoughts on this?

smrutis1:
> 🚀 The idea looks great. In the same endpoint, we can list both, configs from properties file and from DB. This also allows quick update of config in running system, as current way requires restart of tomcat server to update the config constants.
> 
> Its a go ahead from me 👍
> 
> @smrutis1 , what are your thoughts on this?

The approach is great. Just an thought: As we are planning to remove `thrift` in the recent future. Would it be possible here to do it without the thrift layer.

heliocastro:
iSomething that we will need, we need to have unique ids for each entry related to to the menu itself. 
To be morfe direct, i want to restore the menu json that we construct the interface retrieved from server.
Why that ?
Well, imagine that we have a user that i restricted to access only the licenses page, or an base administrator that can only see the users and the sanitize page ?

hoangnt2:
@heliocastro Yes, we can also retrieve the menu from the server. Do we need to allow the ADMIN user to modify the menu from the GUI? This will make the menu more flexible but could also introduce more problems.

hoangnt2:
@smrutis1 Do we have any alternative solutions to Thrift, or should we simply remove it? Currently, I don't know how to implement this without Thrift.

smrutis1:
> @smrutis1 Do we have any alternative solutions to Thrift, or should we simply remove it? Currently, I don't know how to implement this without Thrift.

I was planning to directly call the backend instead of passing through thrift layer. But lets take it as a separate task for analyzing it better and identify any potential pitfalls. We can proceed with this approach for now.

bibhuti230185:
@hoangnt2 Are you still engaged on fixing this build and test fails?

hoangnt2:
> @hoangnt2 Are you still engaged on fixing this build and test fails?
Hi @bibhuti230185, I am putting this task on hold to work on another task. I will fix this PR later.

hoangnt2:
@GMishx, I have add some configuration that I think it can be changed from GUI.
Here are the remaining properties (I referred to your unified sw360.properties file in https://github.com/eclipse-sw360/sw360/pull/2766 (Some of them are related to Liferay and no longer used in the current back-end)
[sw360.properties.txt](https://github.com/user-attachments/files/19386091/sw360.properties.txt)

Please feel free to add your comments

heliocastro:
Sounds good.
@hoangnt2 Maybe is a good idea merge this one and gradually do the other changes ?
Otherwise will become too big

hoangnt2:
> Sounds good. @hoangnt2 Maybe is a good idea merge this one and gradually do the other changes ? Otherwise will become too big

@heliocastro , I didn't test all the properties, so I'm worried this might cause some problems. By the way, I am also implementing the front-end for configuration changes and replacing some environment variables

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/configuration/SW360ConfigurationsController.java
```diff
@@ -0,0 +1,166 @@
...
+    @Override
+    public RepositoryLinksResource process(RepositoryLinksResource resource) {
+        resource.add(linkTo(SW360ConfigurationsController.class)
+                .slash("api" + SW360_CONFIG_URL).withRel("configurations"));
+        return resource;
+    }
+
+    @GetMapping(value = SW360_CONFIG_URL)
+    @Operation(summary  = "Get configurations",
+            description = "This method lists SW360 configurations")
```
GMishx, line 71: Please add `tags = {"Admin"}` or `tags = {"Config"}` to the `@Operation`.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/configuration/SW360ConfigurationsController.java
```diff
@@ -0,0 +1,166 @@
...
+            return ResponseEntity.ok(sw360ConfigurationsService.getSW360ConfigFromDb());
+        }
+
+        return ResponseEntity.ok(sw360ConfigurationsService.getSW360ConfigFromProperties());
+    }
+
+    @PreAuthorize("hasAuthority('ADMIN')")
+    @PatchMapping(value = SW360_CONFIG_URL, consumes = {MediaType.APPLICATION_JSON_VALUE})
+    @Operation(summary  = "Get configurations",
+            description = "This method lists SW360 configurations")
```
GMishx, line 110: Please add `tags = {"Admin"}` or `tags = {"Config"}` to the `@Operation`.

GMishx (changes requested):
Just 2 more changes to go :-)

GMishx (approved):
I see `@Tag` is used at controller level. Thus previous changes are no longer required.

The changes looks good. I am currently testing this PR.

GMishx (approved):
Tested, working as expected.

PR #2829, opened 2024-12-17
feat(version): generate OpenAPI doc version on fly

GMishx:

Use `git-commit-id-maven-plugin` plugin to read git properties and generate version for OpenAPI doc on the fly.

The version uses major and minor version from git and number of commits between HEAD and last tag as the patch version.

This helps in providing a reliable way of identifying updates in REST API until we have stable REST API to use custom API version.

### Suggest Reviewer
@heliocastro @hoangnt2 @smrutis1

### How To Test?
Open the OpenAPI docs and check the API version.

### DIFF
--- a/build-configuration/pom.xml
+++ b/build-configuration/pom.xml
@@ -26,6 +26,11 @@
         <resources>
             <resource>
                 <directory>resources</directory>
+                <filtering>true</filtering>
+                <includes>
+                    <include>**/*.properties</include>
+                    <include>**/*.xml</include>
+                </includes>
             </resource>
         </resources>
 
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,6 @@
     </repository>
   </distributionManagement>
   <properties>
-    <sw360.version>19.0.0</sw360.version>
     <rest.version>2.0.0</rest.version>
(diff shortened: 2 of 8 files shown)


heliocastro:
@GMishx Something special to test ? I want to add the new API page address as the one for footer in the frontend. What you think i should add ?

GMishx:
> @GMishx Something special to test ?

@heliocastro just need to test if the version is updated even if the repository does not contain the tag information (how the github checkout action works).

> I want to add the new API page address as the one for footer in the frontend. What you think i should add ?

It can be the link to the Swagger UI. The user/developer can fetch the OpenAPI docs from the same page.

However, this link can be changed in backend with `springdoc.swagger-ui.path` property in `application.yml`.

The default is http://localhost:8080/resource/swagger-ui/index.html

GMishx:
The XSS filter was causing issue with swagger-ui. For now I have ignored the headers causing the issue from the XSS filter and the Swagger-ui now load. However, XSS filter itself will need re-implementation.

amritkv (dismissed):
LGTM !

amritkv (approved):
LGTM !

PR #2938, opened 2025-03-02
issue#954

Shushant-Priyadarshi:
I added the hero section image from the website on top of the README.md to make it more inviting and attractive.

## Before:
Screenshot 2025-03-02 at 11 45 28 PM

## After:

Screenshot 2025-03-02 at 11 48 18 PM


heliocastro (approved):
It's good, but we do need update the main page though.

PR #2949, opened 2025-03-04
feat(rest): Added parameter for exporting linked package information along with release info.

### LINKED ISSUE #2948: Need an endpoint to export package info linked to a project along with release info.
Currently we have an endpoint for license clearing tab which exports releases info linked to a project but we also need the same for packages if package portlet is enabled.

akshitjoshii:
closes : #2948 

Endpoint url : http://localhost:8080/resource/api/reports

parameters : 
**withlinkedpackages** - `true`
**mimetype** -` xlsx`
**module** - `projects`
**Id** - `projectId`

Sample url : http://localhost:8080/resource/api/reports?withlinkedpackages=true&mimetype=xlsx&module=projects&projectId={projectId}

Sample response : Excel sheet.

image

### DIFF
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/PackageHelper.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/PackageHelper.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright Siemens Healthineers GmBH, 2025. Part of the SW360 Portal Project.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.exporter.helper;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.thrift.TException;
+import org.eclipse.sw360.datahandler.common.ThriftEnumUtils;
+import org.eclipse.sw360.datahandler.thrift.SW360Exception;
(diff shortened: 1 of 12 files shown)


sameed20:
Testing this PR

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/PackageHelper.java
```diff
@@ -0,0 +1,135 @@
...
+                row.add(document.getPurl());
+                break;
+            case PACKAGE_TYPE:
+                row.add(ThriftEnumUtils.enumToString(document.getPackageType()));
+                break;
+            case RELEASE:
+                try {
+                    if (document.getReleaseId() == null || document.getReleaseId().isEmpty()) {
+                        // Ensure row is filled for each header of the release
+                        for (int i = 0; i < 18; i++) {
```
GMishx, line 103: Why there is magic number (18) used here? What does it signify? How does it link to other parts of the code?

GMishx:
@sameed20 any update on this PR?

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/helper/PackageHelper.java
akshitjoshii: 18 is the number of columns that are contributed by the release and in the case where the package does not have a linked release we need to fill these columns with an empty string.

akshitjoshii:
@GMishx, this PR had some issues in the STAGE testing in the generation of the download URL for email. Will be taken up in the coming sprint.

GMishx:
@akshitjoshii any new updates on this PR?

GMishx:
Closing this PR as there is no activity. Please reopen if you have some updates.

Issue #236, opened 2018-04-26
Project sorting by versions is weird because of brackets around


maierthomas:
<a href="https://github.com/mcjaeger"><img src="https://avatars1.githubusercontent.com/u/3418977?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mcjaeger](https://github.com/mcjaeger)**
_Thursday Mar 15, 2018 at 09:57 GMT_
_Originally opened as https://github.com/sw360/sw360portal/issues/780_

----

Actually when doing the sorting of projects in some list view:
```
My Project (1.0)
My Project (1.1)
My Project (2.0)
```
is like not good when:
```
My Project (1.0)
My Project (1.1 SR1)
My Project (1.1)
My Project (2.0)
```
(desired would be:
```
My Project (1.0)
My Project (1.1)
My Project (1.1 SR1)
My Project (2.0)
```
)
Maybe omitting the bracket and adding a dash between name and version could solve it

mcjaeger:
still not there, but also not very much demanded

<img width="985" alt="Screenshot 2021-03-25 at 22 14 20" src="https://user-images.githubusercontent.com/3418977/112544612-8253e580-8db7-11eb-9916-4734d3bf5ed5.png">

lovechouhan:
I love to contribute in this ,please can you provide some guidance me for that.

MamathaGarudampalli:
Hi I'm interested in working on this issue. could you please confirm if its still open and if I can start working on it? Thanks!

himanshu07gupta:
Hi @sw360-bot  @mcjaeger @maierthomas , 
I'm interested in contributing to SW360 and would like to work on this issue.

I've set up the project locally and traced the root cause. The version sorting uses String.CASE_INSENSITIVE_ORDER in two places:

ResourceComparatorGenerator.java
ProjectDatabaseHandler.java
The problem is that lexicographic comparison treats version strings as plain text, so "15.0" < "2.0" because character '1' < '2'.

My proposed fix is to implement a NaturalVersionComparator that splits version strings into numeric and non-numeric segments using a regex pattern, then compares numeric segments as integers instead of strings. This correctly produces: 1.0 < 2.0 < 3.0 < 10.0 < 15.0. Also will handle edge case like which string is bigger at the end etc.

Could you assign this issue to me? Will raise a pr for this.

PR #2963, opened 2025-03-09
feat(projects): Improve project relationship iteration with VersionComparator

Ankush1oo8:

- Added VersionComparator in common/utils for accurate version sorting
- Updated iterateProjectRelationShips methods to use VersionComparator
- Improved sorting logic for project links based on name and version
- Ensured compatibility with existing project relationship handling

Issue: #236 

### How To Test?
1. Run the SW360 backend and verify project relationships are sorted correctly.
2. Ensure no functionality is broken by running tests.
3. Validate the sorting logic using different project versions.

### Checklist
Must:

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
@@ -1,0 +1,25 @@
+package org.eclipse.sw360.common.utils;
+
+import java.util.Comparator;
+
+public class VersionComparator implements Comparator<String> {
+    @Override
+    public int compare(String v1, String v2) {
+        if (v1 == null || v2 == null) {
+            return (v1 == null) ? ((v2 == null) ? 0 : -1) : 1;
+        }
+        String[] parts1 = v1.split("\\.");
+        String[] parts2 = v2.split("\\.");
+        
+        int length = Math.max(parts1.length, parts2.length);
+        for (int i = 0; i < length; i++) {
+            int num1 = (i < parts1.length) ? Integer.parseInt(parts1[i]) : 0;
+            int num2 = (i < parts2.length) ? Integer.parseInt(parts2[i]) : 0;
+            
+            if (num1 != num2) {
+                return Integer.compare(num1, num2);
+            }
+        }
+        return 0;
+    }
+}
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -90,7 +90,7 @@ import static org.eclipse.sw360.datahandler.common.WrappedException.wrapTExcepti
 import static org.eclipse.sw360.datahandler.permissions.PermissionUtils.makePermission;
 import org.eclipse.sw360.exporter.ProjectExporter;
 import java.nio.ByteBuffer;
-
+import org.eclipse.sw360.common.utils.VersionComparator;
 /**
  * Class for accessing the CouchDB database
  *
@@ -991,7 +991,7 @@ public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
             }
             projectLinkOptional.ifPresent(out::add);
         }
-        out.sort(Comparator.comparing(ProjectLink::getName).thenComparing(ProjectLink::getVersion));
+        out.sort(Comparator.comparing(ProjectLink::getName).thenComparing(ProjectLink::getVersion,new VersionComparator()));
         return out;
     }
 
@@ -2444,7 +2444,7 @@ public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
             }
             projectLinkOptional.ifPresent(out::add);
         }
-        out.sort(Comparator.comparing(ProjectLink::getName).thenComparing(ProjectLink::getVersion));
+        out.sort(Comparator.comparing(ProjectLink::getName).thenComparing(ProjectLink::getVersion, new VersionComparator()));
         return out;
     }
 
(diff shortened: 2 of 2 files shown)


backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
```diff
@@ -0,0 +1,25 @@
+package org.eclipse.sw360.common.utils;
```
GMishx, line 1: Please add license and copyright information for this file.

See: https://eclipse.dev/sw360/docs/development/dev-dod-and-style/#licensing-and-file-header

GMishx (changes requested):

backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
Ankush1oo8: done

backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright COPYRIGHT HOLDER, 2017.
+ * Copyright NEXT COPYRIGHT HOLDER, 2017.
```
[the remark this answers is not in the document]
Ankush1oo8: sorry Did the changes

backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
```diff
@@ -0,0 +1,37 @@
...
+        if (v1 == null || v2 == null) {
+            return (v1 == null) ? ((v2 == null) ? 0 : -1) : 1;
+        }
+        String[] parts1 = v1.split("\\.");
+        String[] parts2 = v2.split("\\.");
+        
+        int length = Math.max(parts1.length, parts2.length);
+        for (int i = 0; i < length; i++) {
+            int num1 = (i < parts1.length) ? Integer.parseInt(parts1[i]) : 0;
+            int num2 = (i < parts2.length) ? Integer.parseInt(parts2[i]) : 0;
```
GMishx, line 29: `parseInt` can throw `NumberFormatException` exception if the input is not a number which is possible for version. Example, version `3.16.43-2+deb8u2` is a valid Debian version but will fail here for `43-2+deb8u2`.

Please handle the exception appropriately.

GMishx (changes requested):
A small change is still required.

backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
Ankush1oo8: Ok

Ankush1oo8: done

Ankush1oo8:
@GMishx can you check now

backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
```diff
@@ -0,0 +1,52 @@
...
+            if (num1 != num2) {
+                return Integer.compare(num1, num2);
+            }
+        }
+        return 0;
+    }
+
+    private int handleNonNumeric(String part) {
+        // Assign a lower priority to non-numeric parts
+        return part.matches("\\d+") ? Integer.parseInt(part) : -1;
```
GMishx, line 50: `matches` here will just tell you the string contains integers or not. If the string contains integer (`43-2+deb8u2` for example), calling `parseInt` again inside exception handler for `NumberFormatException` will throw you another `NumberFormatException` don't you think?

GMishx (changes requested):

Ankush1oo8:
Why This Works

If part = "43-2+deb8u2", numericPrefix = "43", which can be parsed safely.

If part has no numeric prefix (e.g., "deb8u2"), it gets a default -1 for lower priority.

It ensures that versions like "3.16.43-2+deb8u2" are still sorted meaningfully.

This way, the comparator remains robust without causing additional NumberFormatException.

Ankush1oo8:
@GMishx can you check now

backend/common/src/main/java/org/eclipse/sw360/common/utils/VersionComparator.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private int handleNonNumeric(String part) {
+    // Extract leading numeric portion if available
+    String numericPrefix = part.replaceAll("^(\\d+).*", "$1");
+    
+    try {
+        return Integer.parseInt(numericPrefix);
+    } catch (NumberFormatException e) {
+        return -1; // Assign a lower priority for fully non-numeric parts
+    }
```
GMishx, line 56: I like where you are going with this. But it is not generic enough. What if project is using some other versioning scheme like `v1.2.3`?

Comparing `v1.2.3` and `v1.2.4` won't work with given regex.

GMishx (changes requested):

Ankush1oo8:
@GMishx should I also optimise for version with aplha or beta

GMishx:
> Why This Works
> 
> If part = "43-2+deb8u2", numericPrefix = "43", which can be parsed safely.
> 
> If part has no numeric prefix (e.g., "deb8u2"), it gets a default -1 for lower priority.
> 
> It ensures that versions like "3.16.43-2+deb8u2" are still sorted meaningfully.
> 
> 
> 

I agree but current solution works for cases like `1.2.2` < `1.2.3`

Will work for the Debian version strings.

But would say `v1.3.2` > `v2.1.1`

The solution should be generic enough.

Ankush1oo8:
@GMishx I have also added for
 Handles epoch versions (1:2.3.4 vs 2.3.4) correctly
 Ignores build metadata (1.2.3+build.4 == 1.2.3)
 Normalizes leading zeros (01.02.003 == 1.2.3)
 Supports configurable pre-release precedence (alpha < beta < rc)
 Optimized with caching for performance

GMishx:
@shi9qiu can you please help test / review this branch.

GMishx:
The is a more simpler implementation in #3791. Thus I would close this PR. Thanks for your valuable contribution @Ankush1oo8

Issue #2951, opened 2025-03-06
Quick Filter (UI) : License page


keerthi-bl:
Create a new end point for quick filter in license page.

GMishx:
Filtering should be implemented on client side.

See https://github.com/eclipse-sw360/sw360/pull/2979#issuecomment-2764587295 for more info.

PR #2979, opened 2025-03-13
feat(rest) : Quickfilter for license page

keerthi-bl:

Issue: Closes : #2951 

### Suggest Reviewer
@GMishx 

### How To Test?
URL : http://localhost:8080/resource/api/licenses/quickFilter?search=apache&page=0&page_entries=5

### Checklist
Must:

### DIFF
--- a/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
+++ b/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
@@ -1284,4 +1284,12 @@ public class LicenseDatabaseHandler {
         }
         return obligationText.replaceFirst("\n","");
     }
+
+    public List<License> searchByShortName(String searchText) {
+        return licenseRepository.searchByShortName(searchText);
+    }
+
+    public List<License> searchByName(String searchText) {
+        return licenseRepository.searchByName(searchText);
+    }
 }
--- a/libraries/datahandler/src/main/thrift/licenses.thrift
+++ b/libraries/datahandler/src/main/thrift/licenses.thrift
@@ -371,4 +371,14 @@ service LicenseService {
      * convert Text to Node in obligation
      **/
     string convertTextToNode(1:Obligation Obligation, 2: User user);
+
+    /**
+    * global search function to list license elements which match the text argument by short name
+    */
+    list<License>  searchByShortName(1: string text);
+
+    /**
+    * global search function to list license elements which match the text argument by name
+    */
+    list<License>  searchByName(1: string text);
 }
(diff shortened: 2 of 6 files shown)


backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
```diff
@@ -1284,4 +1284,12 @@ private String buildObligationText(JSONObject jsonObject, int level) {
         }
         return obligationText.replaceFirst("\n","");
     }
+
+    public List<License> quickSearchLicenseByShortName(String searchText) {
+        return licenseRepository.searchByShortName(searchText);
+    }
+
+    public List<License> quickSearchLicenseByName(String searchText) {
+        return licenseRepository.searchByName(searchText);
```
GMishx, line 1293: Let's not cause confusion by calling these functions **quickSearch**. Please just call them `searchByShortName` and `searchByName` as they are not doing anything different.

libraries/datahandler/src/main/thrift/licenses.thrift
```diff
@@ -371,4 +371,14 @@ service LicenseService {
...
+
+    /**
+    * global search function to list license elements which match the text argument by short name
+    */
+    list<License>  quickSearchLicenseByShortName(1: string text);
+
+    /**
+    * global search function to list license elements which match the text argument by name
+    */
+    list<License>  quickSearchLicenseByName(1: string text);
```
GMishx, line 383: Same here.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
```diff
@@ -525,4 +525,34 @@ public ResponseEntity<RequestStatus> createLicenseType(
         HttpStatus status = HttpStatus.OK;
         return new ResponseEntity<>(requestStatus, status);
     }
+
+    @Operation(
+            summary = "List all of the service's licenses.",
+            description = "List all of the service's licenses.",
+            tags = {"Licenses"}
+    )
+    @RequestMapping(value = LICENSES_URL+"/quickFilter", method = RequestMethod.GET)
```
GMishx, line 534: Instead of creating new endpoint, I'll update the existing `getLicenses()` with a new parameter like `?search`. Simply use the filtered results if this parameter exists, else get all licenses as usual.

GMishx (changes requested):

GMishx (approved):
Changes looks good.

GMishx (changes requested):
One minor change needed.

However, I was wondering, the functions uses indexes and works with "Starts with" type queries. For example, I can get all the licenses which starts with "GPL". However, I cannot search for licenses which "contains" "GPL", example "LGPL".

What are your opinions on it @keerthi-bl ?

GMishx:
(from discussions)

The endpoint should support "contains" case as well, not just "starts with" which requires implementation of full-text search with Nouveau. However, the data return from endpoint is not big and filtering can be implemented in the front-end easily but efforts for implementation is too big. Thus closing this PR.

PR #2983, opened 2025-03-14
fix(Project): Resolve issue with GET project release when response length is 0

duonglq-tsdv:
# Fix Embedded Type in Project Release Response

When calling the API to get project releases and there are no releases (length = 0), the backend returns an embedded type of `Project`, while the frontend expects `[sw360:releases]`. This mismatch causes rendering issues on the frontend. The change ensures the embedded type aligns with the expected `[sw360:releases]`.


GMishx (approved):
LGTM

Issue #2986, opened 2025-03-15
Add Status Check for Report Generation in Fossology


Mohamed-hanfy:
### Description

Currently, the `triggerReportGenerationFossology` method is responsible for initiating report generation in Fossology. However, there is no corresponding method to check the status of this process. This is inconsistent with how other Fossology operations are handled, such as:

`checkUnpackStatus` for checking the status of an unpack operation.
`checkScanStatus` for checking the status of a scan.

To maintain consistency, it would be helpful to introduce a checkReportGenerationStatus method to allow users to track the progress of report generation.

### Expected Behavior
- A method should be available to check if the report generation is in progress, completed, or failed.

Mohamed-hanfy:
I will start working on this and submit a PR shortly. Let me know if there are any specific requirements or feedback before I proceed!

PR #3031, opened 2025-03-26
feat(rest,thrift)!: add checkReportGenerationStatus method for Fossology report tracking

Mohamed-hanfy:
Closes eclipse/sw360#2986

## Description
This feature adds a method to check the status of Fossology report generation across multiple components of the SW360 system. 

### Changes
- Implemented `checkReportGenerationStatus` method in:
  1. Fossology REST Client
  2. Fossology Thrift Service Interface
  3. SW360 Release Service
  4. Release Controller

## How to Test

1. Trigger a Fossology report generation for a release
2. Call the new endpoint: `GET /releases/{id}/checkFossologyProcessStatus`
   - Verify the response contains:
     * Overall process status (PROCESSING, SUCCESS, or FAILURE)
     * Detailed report status (if a report step exists)
     * Fossology process information
3. Test various scenarios:
   - Release with an ongoing Fossology process
   - Release with a completed Fossology process
   - Release without a Fossology process
4. Verify error handling for:
   - Invalid release ID
   - Releases with malformed process step IDs

## Dependencies
- No new dependencies added
- Uses existing REST and Thrift clients

## Suggested Reviewers
@GMishx 

## Checklist
- [x] Implemented status check method
- [x] Added comprehensive logging
- [x] Handled error scenarios


Mohamed-hanfy:
I've updated the test documentation to include the new reportStatus field that was added to the checkFossologyProcessStatus endpoint.

KoukiHama:
Nice PR, I will check this one.

backend/fossology/src/main/java/org/eclipse/sw360/fossology/rest/FossologyRestClient.java
```diff
@@ -512,4 +512,72 @@ public int getUploadId(String shaValue, String fileName) {
...
+     * @param reportId the id of the report whose status should be queried.
+     * @return a Map containing status information about the report generation
+     */
+    public Map<String, String> checkReportGenerationStatus(int reportId) {
+        String baseUrl = restConfig.getBaseUrlWithSlash();
+        String token = restConfig.getAccessToken();
+        Map<String, String> responseMap = new HashMap<>();
+
+        if (StringUtils.isEmpty(baseUrl) || StringUtils.isEmpty(token)) {
+            log.error("Configuration is missing values! Url: <{}>, Token: <{}>", baseUrl, token);
```
GMishx, line 529: Not a best practice to log security tokens

GMishx (approved):
Changes looks good.

backend/fossology/src/main/java/org/eclipse/sw360/fossology/rest/FossologyRestClient.java
```diff
@@ -512,4 +512,72 @@ public int getUploadId(String shaValue, String fileName) {
...
+        HttpHeaders headers = new HttpHeaders();
+        headers.set("Authorization", "Bearer " + token);
+
+        try {
+            log.debug("Checking report generation status for reportId {} at {}", reportId, baseUrl + "report/" + reportId);
+            ResponseEntity<JsonNode> responseEntity = restTemplate.exchange(
+                    baseUrl + "report/" + reportId,
+                    HttpMethod.GET,
+                    new HttpEntity<>(headers),
+                    JsonNode.class);
```
GMishx, line 547: I was looking into FOSSology OpenAPI docs and noticed the endpoint `/report/{id}` is used to download the report, which is done by `‎FossologyRestClient::getReport()` already. How is this function different?

backend/fossology/src/main/java/org/eclipse/sw360/fossology/rest/FossologyRestClient.java
```diff
@@ -512,4 +512,72 @@ public int getUploadId(String shaValue, String fileName) {
...
+            if (body.has("status")) {
+                String status = body.get("status").asText();
+                responseMap.put("status", status);
+                log.debug("Report generation status for reportId {}: {}", reportId, status);
+            }
+
+            // Add any other relevant fields from the response
+            if (body.has("message")) {
+                responseMap.put("message", body.get("message").asText());
+            }
```
GMishx, line 565: According to FOSSology's OpenAPI docs, there is no `status` or `message` return by FOSSology on `/report/{id}` endpoint.

GMishx (changes requested):
Upon further investigation, there are some points which I cannot find answers for. Can you please help @Mohamed-hanfy ?

backend/fossology/src/main/java/org/eclipse/sw360/fossology/rest/FossologyRestClient.java
Mohamed-hanfy: hello @GMishx,
thanks for the review, my main purpse here to check  status of the report so i created a call to the endpoint to get response but this endpint return a file so we can try to download this file and build the response if downloaded mark the message as completed and so on, but downloading will take alot of time so case overhead in the system and this approach not senstive becuse we need to handle retry to download .
 after review to  `handleReportStep`  method in `backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java`  i found that it do the same check but in business layer and handle the problem  of retry downloading by change the status of the fossology process to 'IN_WORK' and handle time problem by mark done before start check if file downloaded or not .  so we need to measure how mush the system need to `checkReportGenerationStatus` method in data acsess layer.

GMishx:
@Mohamed-hanfy thank you for the contribution but it does not make sense to me. There are functions specifically to do what you are proposing as part of this PR. Also, the code you are contributing is calling endpoints on FOSSology which does not exists at all.

Thus, I am closing this PR for now. Please reopen if you think otherwise.

PR #3049, opened 2025-04-01
fix(Obligations): includes obligationLevel in GET all obligations responses

duonglq-tsdv:
When calling the API to retrieve all obligations, the backend returns a response containing an embedded list of obligations. However, the obligationLevel field, a key attribute of an obligation, is currently missing from this response. 

I have updated the code to include the obligationLevel attribute in the response. This ensures the frontend can properly access and display this information as needed.


GMishx (approved):
LGTM

PR #3051, opened 2025-04-03
fix(spdx): add null and empty field checks for SPDX documents

Mohamed-hanfy:
Add validation checks to ensure SPDX documents have required fields:
- Add null object validation for SPDXDocument and DocumentCreationInformation
- Validate ID fields in SPDXDocument are not empty
- Ensure creator and creation date are present in DocumentCreationInformation

### Suggest Reviewer
@GMishx @KoukiHama 

### Checklist
Must:


GMishx (approved):
LGTM

Issue #3046, opened 2025-03-31
Prevent duplicate external IDs in projects and fix related compilation errors


web1havv:
>>Description

The project creation API currently allows multiple projects to be created with identical external IDs, which can cause data integrity issues and problems with external system integrations. Additionally, the initial implementation attempt resulted in compilation errors due to type mismatches.

>>Steps followed and expected result

Reproduction Steps:

Create Project A with external ID {"ERP": "P12345"}

Attempt to create Project B with same external ID {"ERP": "P12345"}

Observe both projects are created

>>Expected Behavior:

Second project creation should fail with DUPLICATE status

System should maintain consistent external ID uniqueness

GMishx:
Check comment: https://github.com/eclipse-sw360/sw360/pull/3036#issuecomment-2765917327

GMishx:
Check comment https://github.com/eclipse-sw360/sw360/pull/3094#issuecomment-2838827091

Issue #3093, opened 2025-04-20
Support Multiple Values for External IDs in SW360 Projects


web1havv:
### Description

Currently, the SW360 project model allows only a single value per external ID key (map<string, string>). This limitation can cause data loss in scenarios where a project needs to reference multiple external IDs under the same key (e.g., multiple package URLs for components from a monorepo).

As discussed with @GMishx, there is a valid use case for allowing multiple values per external ID key. This will enable better tracking and association of components that share a source or package reference.

Proposed Solution:

Update the Thrift model (Project.thrift) to use map<string, set<string>> for the externalIds field.

Ensure that all relevant service and search methods support multiple values per key and return all matching projects for a given external ID.

GMishx:
The idea is good but it requires API versioning. See https://github.com/eclipse-sw360/sw360/pull/3094#issuecomment-2838827091

Closing the issue for now. Once versioning is implemented, can work on this.

GMishx:
Moving the issue to backlog instead of closing so we can work on this in future.

devlopharsh:
hii @GMishx is the issue still open ? i wanna work over this !!

PR #3036, opened 2025-03-30
feat(projects): Add duplicate external ID validation in project creation

web1havv:
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

>>Summary of key changes:

Added validation in addProject() to check for duplicate external IDs

Returns RequestStatus.DUPLICATE if matching external IDs found

Uses existing searchByExternalIds() method for consistency

>>Dependencies

None

### Issue
Resolves #3046 

>>How to verify:

Attempt to create project with existing external IDs (should fail with DUPLICATE status)

Create project with unique external IDs (should succeed)

Verify projects without external IDs are unaffected

>>Test coverage:

Unit tests to be added

Manual testing performed

### Checklist
Must:
- [ ] All related issues are referenced in commit messages and in PR

PR #3094, opened 2025-04-20
 feat(projects): Support Multiple Values for External IDs

web1havv:
###Description
This PR implements support for storing and searching multiple values per external ID key in SW360 projects, as discussed.
###Key Changes:

Updated Project.thrift to use map<string, set<string>> for the externalIds field, allowing projects to reference multiple external IDs under the same key.

Verified that service and search functionality returns all projects matching any provided external ID value.

Ensured that no restrictions are imposed on external ID duplication across projects, in line with SW360’s user-managed philosophy.

Issue: 
Resolves Issue #3046 
#3093 

### Suggest Reviewer
@GMishx 

### How To Test?
Create a project with multiple values for a single external ID key and confirm they are all stored.

Use the /projects/searchByExternalIds endpoint to search for projects by external ID and verify that all matching projects are returned.

### Checklist
Must:

### DIFF
--- a/libraries/datahandler/src/main/thrift/projects.thrift
+++ b/libraries/datahandler/src/main/thrift/projects.thrift
@@ -268,7 +268,7 @@ struct ProjectDTO{
     7: optional string domain,
 
     // information from external data sources
-    9: optional map<string, string> externalIds,
+    9: optional map<string, set<string>> externalIds,
     300: optional map<string, string> additionalData,
 
     // Additional informations


web1havv:
Solves Issue
#3093

GMishx:
Hey @web1havv 

I discussed the idea with other existing users of SW360. The feedback I receive is as following:

The changes you've proposed are nice and definitely a value add for the project. However, it is changing working of a core component of the project, mainly using External ID for putting [package-url](https://github.com/package-url/purl-spec) and using them to correctly identify the components.
Thus, any change in the current implementation will require changes in the REST API clients of SW360 and break all of them until they are fixed. And there are many such clients!

Therefore, until SW360 has versioned APIs, we should not include such breaking changes.

Sorry, you have already invested so much effort in #3093, #3046 and #3036 but due to in capability of SW360 API, we currently cannot accommodate this change.

web1havv:
Thanks for clarifying @GMishx.

GMishx:
Thank you for understanding @web1havv
I'll close the PR and related issues.

Issue #3104, opened 2025-04-23
RequestSummary.getMessage() Returns Null in Error Handling 


Mohamed-hanfy:
### Description

In `ProjectController.java`,  when importing a project from an invalid SBOM file, `RequestSummary.getMessage() ` returns null. so we need to return meaningful error feedback to users or for logging purposes

 
https://github.com/user-attachments/assets/cf8ffcd5-f32b-43e1-912e-f18959a4bb20

PR #3105, opened 2025-04-23
fix(rest): improve error messages for invalid SBOM file imports

Mohamed-hanfy:

i traced `importSBOM` method in  `ComponentController.java` and this problem handled by return message 'Invalid SBOM file' so i followed the same approch in `ProjectController.java` to achive consistent error handling for invalid SBOM uploads across both controllers.
Issue: 
closes: #3104

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -2361,14 +2361,14 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
             requestSummary = projectService.importSPDX(sw360User, attachment.getAttachmentContentId());
 
             if (!(requestSummary.getRequestStatus() == RequestStatus.SUCCESS)) {
-                throw new BadRequestClientException(requestSummary.getMessage());
+                throw new BadRequestClientException((requestSummary.getMessage()!=null)? requestSummary.getMessage() : "Invalid SBOM file");
             }
             projectId = requestSummary.getMessage();
         } else {
             requestSummary = projectService.importCycloneDX(sw360User, attachment.getAttachmentContentId(), "", true);
 
             if (requestSummary.getRequestStatus() == RequestStatus.FAILURE) {
-                throw new BadRequestClientException(requestSummary.getMessage());
+                throw new BadRequestClientException((requestSummary.getMessage()!=null)? requestSummary.getMessage() : "Invalid SBOM file");
             }
             else if (requestSummary.getRequestStatus() == RequestStatus.ACCESS_DENIED) {
                 throw new BadCredentialsException("You do not have sufficient permissions.");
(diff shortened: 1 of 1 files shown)


Mohamed-hanfy:
Git ECA Validation Status service is down right now so that it fails

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -2361,14 +2361,14 @@ public ResponseEntity<?> importSBOM(
             requestSummary = projectService.importSPDX(sw360User, attachment.getAttachmentContentId());
 
             if (!(requestSummary.getRequestStatus() == RequestStatus.SUCCESS)) {
-                throw new BadRequestClientException(requestSummary.getMessage());
+                throw new BadRequestClientException("Invalid SBOM file");
```
GMishx, line 2364: I'd recommend to first check if `requestSummary.getMessage()` is `null`, then override it with a literal string as it can be set in few cases with useful information.

https://github.com/eclipse-sw360/sw360/blob/b67e732d69e56a20f0c5eb401781e2c12fb49112/backend/common/src/main/java/org/eclipse/sw360/spdx/SpdxBOMImporter.java#L149

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -2361,14 +2361,14 @@ public ResponseEntity<?> importSBOM(
...
             }
             projectId = requestSummary.getMessage();
         } else {
             requestSummary = projectService.importCycloneDX(sw360User, attachment.getAttachmentContentId(), "", true);
 
             if (requestSummary.getRequestStatus() == RequestStatus.FAILURE) {
-                throw new BadRequestClientException(requestSummary.getMessage());
+                throw new BadRequestClientException("Invalid SBOM file");
```
GMishx, line 2371: Same here.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -2439,7 +2439,7 @@ public ResponseEntity<?> importSBOMonProject(
         requestSummary = projectService.importCycloneDX(sw360User, attachment.getAttachmentContentId(), id, doNotReplacePackageAndRelease);
 
         if (requestSummary.getRequestStatus() == RequestStatus.FAILURE) {
-            throw new BadRequestClientException(requestSummary.getMessage());
+            throw new BadRequestClientException("Invalid SBOM file");
```
GMishx, line 2442: Same here.

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
Mohamed-hanfy: Done
should I follow the same approach  in `ComponentController.java` line 997, as it doesn't check before override the message?

GMishx: That would be a good idea, but in a different PR.

GMishx (approved):
Changes looks good.

PR #3142, opened 2025-05-07
fix(component): ComponentDTO for /splitcomponent

GMishx:

1. Use `ComponentDTO` for `/splitcomponent` as `Component` does not accept attachments.
2. Add `releaseIds` to `ComponentDTO` as `Component` has it and makes it easy for client to provide the release info.

### How To Test?
Try to split some components with multiple releases and attachments.


GMishx:
@deo002 can you please test this PR?

deo002:
Tested, working as expected

PR #3148, opened 2025-05-10
feat(rest): add SBOM file validation for SPDX and CycloneDX formats

### LINKED ISSUE #3091: Add Magic Byte Validation for SBOM Uploads
### Description

Currently, we only check SBOM file extensions, MIME types, and perform schema validation after upload. This leaves us vulnerable to spoofed or malicious files that disguise their true type.

Add a check for magic bytes (file signatures) on upload to ensure the file content matches the expected SBOM format (e.g., {"spdxVersion" for SPDX JSON, SPDX- for RDF/XML).
Benefit: This will prevent disguised files from being processed, improve security, and add an extra layer of validation before parsing.

Mohamed-hanfy:
continue working in pr https://github.com/eclipse-sw360/sw360/pull/3092 after rebase with main

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/spdx/document/SpdxDocumentDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/spdx/document/SpdxDocumentDatabaseHandler.java
@@ -12,7 +12,11 @@ package org.eclipse.sw360.datahandler.db.spdx.document;
 
 import com.ibm.cloud.cloudant.v1.Cloudant;
 
+import org.apache.logging.log4j.core.util.UuidUtil;
 import org.apache.thrift.TException;
+import org.cyclonedx.exception.ParseException;
+import org.cyclonedx.parsers.JsonParser;
+import org.cyclonedx.parsers.XmlParser;
 import org.eclipse.sw360.datahandler.cloudantclient.DatabaseConnectorCloudant;
 import org.eclipse.sw360.datahandler.common.DatabaseSettings;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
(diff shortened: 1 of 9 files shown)


Mohamed-hanfy:
hello @GMishx, 

There is an unexpected problem while building it happen in `ProjectSpecTest.java` three tests fails :

`should_document_get_linked_releases_in_dependency_network_by_index_path()`  , 

`should_document_get_linked_resources_of_project_in_dependency_network ()` and

` should_document_create_project_with_network()` 
logs for all of them sayes 
`Caused by: SW360Exception(why:Please enable flexible project release relationship configuration to use this function (enable.flexible.project.release.relationship = true))`

Actually, they seem outside the scope I'm working on in this pull request. However, after spending four hours debugging , I discovered that adding the new dependencies is the cause. But why and where both test meets its not clear at all to me.

GMishx:
Hey @Mohamed-hanfy , I am not sure why this issue happened but what I understood so far is these test cases are designed to test the failure cases and expected 500 response code. The resource server currently throws these failures as exceptions and uses handler to convert them into response types.

Seems like addition of these dependencies caused the spring to confuse with response format and tries to parse these failure messages as XML where it fails. Current solution I found is to add `Accepts` header, so spring tries to translate them correctly as JSON

```diff
diff --git i/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java w/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
index 5007eb68d..39c85ce1d 100644
--- i/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
+++ w/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/ProjectSpecTest.java
@@ -2682,7 +2682,8 @@ public class ProjectSpecTest extends TestRestDocsSpecBase {
             this.mockMvc.perform(post("/api/projects/network")
                             .contentType(MediaTypes.HAL_JSON)
                             .content(this.objectMapper.writeValueAsString(project))
-                            .header("Authorization", TestHelper.generateAuthHeader(testUserId, testUserPassword)))
+                            .header("Authorization", TestHelper.generateAuthHeader(testUserId, testUserPassword))
+                            .accept(MediaTypes.HAL_JSON))
                     .andExpect(status().isInternalServerError());
         } else {
             this.mockMvc.perform(post("/api/projects/network")
@@ -2965,7 +2966,8 @@ public class ProjectSpecTest extends TestRestDocsSpecBase {
             this.mockMvc
                     .perform(get("/api/projects/network/"+project.getId()+"/linkedResources")
                             .param("transitive", "false")
-                            .header("Authorization", TestHelper.generateAuthHeader(testUserId, testUserPassword)))
+                            .header("Authorization", TestHelper.generateAuthHeader(testUserId, testUserPassword))
+                            .accept(MediaTypes.HAL_JSON))
                     .andExpect(status().isInternalServerError());
         } else {
             this.mockMvc
@@ -3022,7 +3024,8 @@ public class ProjectSpecTest extends TestRestDocsSpecBase {
         if (!SW360Constants.ENABLE_FLEXIBLE_PROJECT_RELEASE_RELATIONSHIP) {
             this.mockMvc
                     .perform(get("/api/projects/network/" + project.getId() + "/releases?path=0->1")
-                            .header("Authorization", TestHelper.generateAuthHeader(testUserId, testUserPassword)))
+                            .header("Authorization", TestHelper.generateAuthHeader(testUserId, testUserPassword))
+                            .accept(MediaTypes.HAL_JSON))
                     .andExpect(status().isInternalServerError());
         } else {
             this.mockMvc
```

I will try to test if this PR breaks other rest endpoints with failure cases. If it does, then we would have look for an alternate solution, otherwise the suggested patch above should work.

GMishx:
Hey @Mohamed-hanfy , current changes are causing trouble with all API endpoints. I tried to fix it with SpringMVC default content type but no luck: https://spring.io/blog/2013/05/11/content-negotiation-using-spring-mvc

So, I have moved your function to a Thrift service under SPDXDocument so Resource server does not need the SPDX and CDX dependencies anymore. I am attaching the patch with changes on top of your commit. Please have a look. (PS, github does not allow `.patch` upload thus it is `.txt`)
[bom-parser.patch.txt](https://github.com/user-attachments/files/20211667/bom-parser.patch.txt)

Mohamed-hanfy:
Hello @GMishx, 
I have applied the patch file, its great! now we check file content pefore make system start to parse the content 
image

GMishx (approved):
Changes looks good.

PR #3174, opened 2025-05-20
feat(rest): new Security user role.

### LINKED ISSUE #3005: Create a new Security User role.
### Description

The Code has already written for the issue, just need to create a PR for this in future.

rudra-superrr:
Closes #3005 

Description: restricting endpoints for new security user role.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
+import org.eclipse.sw360.datahandler.permissions.PermissionUtils;
 import org.eclipse.sw360.datahandler.resourcelists.PaginationOptions;
 import org.eclipse.sw360.datahandler.resourcelists.PaginationParameterException;
 import org.eclipse.sw360.datahandler.resourcelists.PaginationResult;
@@ -97,6 +98,7 @@ import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Service;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 import org.springframework.web.util.UriComponentsBuilder;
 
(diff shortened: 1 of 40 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -1659,4 +1660,11 @@ public ClearingRequest updateCRSize(ClearingRequest clearingRequest, Project pro
     public boolean isWriteActionAllowed(Object object, User user) {
         return makePermission(object, user).isActionAllowed(RequestedAction.WRITE);
     }
+
+    public void isSecurityUser(User user) {
```
GMishx, line 1664: Please use a more helpful name like `throwIfSecurityUser()` as it currently feels like just a validator which returns `true` if the user is `SECURITY_USER`.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
```diff
@@ -147,6 +148,8 @@ public ResponseEntity<HalResource<Map<String, Object>>> getModerationRequestById
             @PathVariable String id
     ) throws TException {
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
+restControllerHelper.isSecurityUser(sw360User);
```
GMishx, line 151:
```suggestion
        restControllerHelper.isSecurityUser(sw360User);
```

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -1659,4 +1660,11 @@ public ClearingRequest updateCRSize(ClearingRequest clearingRequest, Project pro
     public boolean isWriteActionAllowed(Object object, User user) {
         return makePermission(object, user).isActionAllowed(RequestedAction.WRITE);
     }
+
+    public void throwIfSecurityUser(User user) {
+        if (user.getUserGroup().name().equals("SECURITY_USER")) {
```
GMishx, line 1665: A quick question, can we not use `PermissionUtils.isSecurityUser(user)` here?

rudra-superrr: yes, we can. Thanks for pointing out.

GMishx (dismissed):
Changes looks good. Needs test.

GMishx (approved):
Tested, working with few exceptions.

GMishx:
The PR is working as intended. However I have few questions which can be worked upon in a separate PR, thus merging this one.

1. I can access Packages linked to a Project but cannot access the Package itself. Is it intentional?
2. Since every request other than `GET` is blocked for the user, it cannot modify itself, generate tokens, etc. On some endpoints, other HTTP methods should be allowed.

PR #3199, opened 2025-06-06
feat(component): use DB side pagination

GMishx:

Make the `/components` endpoint queries DB side paginated instead of application side. It should make the endpoint more performant as it does not load all data.

### How To Test?
1. Try to fetch paginated responses from `/components` endpoint.
2. Query for components with `/components?luceneSearch=true`
3. Query for exact component name with `/components?luceneSearch=false&name=<name>`

### Known issue (issue to be created)
1. The query where `/components?luceneSearch=false` is called and fields other than `name` are filtered, is not optimized.
2. The query `/components?luceneSearch=false&name=<name>` returns total number of records as if no filter were applied. This can be confusing.

### Checklist
Depends on #3198


amritkv (approved):
LGTM!

Issue #3205, opened 2025-06-07
`VendorController.getReleases()` doesn't validate vendor existence before retrieving releases


Mohamed-hanfy:
## Description
The endpoint in doesn't validate whether a vendor with the given ID exists before attempting to retrieve its releases. This leads to inconsistent API behavior and poor client experience. [getReleases()](https://github.com/eclipse-sw360/sw360/blob/18ac76e0c021707e75112d319143dd96318ebffb/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java#L144C66-L144C77)

## Current Behavior
- **Request:** `GET /vendors/{non-existent-id}/releases`
- **Response:** `204 No Content`
- **Issue:** Client cannot distinguish between "vendor exists but has no releases" vs "vendor doesn't exist"

https://github.com/user-attachments/assets/a15e6467-7f86-4a47-84f0-8de99500abc1

PR #3206, opened 2025-06-07
fix(rest): add vendor existence validation in getReleases endpoint

Mohamed-hanfy:
this PR fixes a critical API inconsistency in the `VendorController.getReleases()` method. Previously, the endpoint would return `204 No Content` for non-existent vendor IDs, making it impossible for clients to distinguish between "vendor exists but has no releases" and "vendor doesn't exist".

close Issue: #3205

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
@@ -140,6 +140,9 @@ public class VendorController implements RepresentationModelProcessor<Repository
             @Parameter(description = "The id of the vendor to get.")
             @PathVariable("id") String id
     ) throws SW360Exception {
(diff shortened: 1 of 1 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
```diff
@@ -141,6 +141,9 @@ public ResponseEntity<CollectionModel<EntityModel<Release>>> getReleases(
             @PathVariable("id") String id
     ) throws SW360Exception {
         try {
+            if(vendorService.getVendorById(id) == null){
+                throw new ResourceNotFoundException("Vendor with id " + id + " not found.");
```
GMishx: Please move this `throw` outside the `try` block otherwise the endpoint will return a `500` instead of a `404`

Mohamed-hanfy:
need to rebase with main

PR #3212, opened 2025-06-09
fix(rest): add vendor existence validation in getReleases endpoint

### LINKED ISSUE #3205: `VendorController.getReleases()` doesn't validate vendor existence before retrieving releases

Mohamed-hanfy:
this PR fixes a critical API inconsistency in the VendorController.getReleases() method. Previously, the endpoint would return 204 No Content for non-existent vendor IDs, making it impossible for clients to distinguish between "vendor exists but has no releases" and "vendor doesn't exist".

closes #3205

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
@@ -147,6 +147,9 @@ public class VendorController implements RepresentationModelProcessor<Repository
     ) throws SW360Exception {
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
         restControllerHelper.throwIfSecurityUser(sw360User);
+        if(vendorService.getVendorById(id) == null){
+            throw new ResourceNotFoundException("Vendor with id " + id + " not found.");
+        }
         try {
             Set<Release> releases = vendorService.getAllReleaseList(id);
             List<EntityModel<Release>> resources = new ArrayList<>();


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/VendorController.java
```diff
@@ -145,6 +145,9 @@ public ResponseEntity<CollectionModel<EntityModel<Release>>> getReleases(
             @Parameter(description = "The id of the vendor to get.")
             @PathVariable("id") String id
     ) throws SW360Exception {
+        if(vendorService.getVendorById(id) == null){
+            throw new ResourceNotFoundException("Vendor with id " + id + " not found.");
+        }
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
         restControllerHelper.throwIfSecurityUser(sw360User);
```
GMishx: Ah, a minor security issue. User with "Security User" role and hit this endpoint with different IDs and see a `404` or `403` based on vendor exists or not.

Just move the 3 lines bellow `restControllerHelper.throwIfSecurityUser(sw360User);` test and we are good to go.

Mohamed-hanfy: DONE

GMishx (approved):
LGTM

PR #3242, opened 2025-06-26
feat(rest): new readOnly user role.

### LINKED ISSUE #3241: Create a new Read Only user role
### Description
This new user role will have read-only access to both components and releases.

rudra-superrr:
Issue: #3241 

This will restrict the endpoints for the new readOnly user role.

### DIFF
--- a/backend/vulnerabilities/src/main/java/org/eclipse/sw360/vulnerabilities/VulnerabilityHandler.java
+++ b/backend/vulnerabilities/src/main/java/org/eclipse/sw360/vulnerabilities/VulnerabilityHandler.java
@@ -242,7 +242,7 @@ public class VulnerabilityHandler implements VulnerabilityService.Iface {
     @Override
     public List<ProjectVulnerabilityRating> getProjectVulnerabilityRatingByProjectId(String projectId, User user) {
         if (!PermissionUtils.isUserAtLeast(UserGroup.USER, user)) {
-            return null;
+            return Collections.emptyList();
         }
         return projectDatabaseHandler.getProjectVulnerabilityRatingByProjectId(projectId);
     }
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/PermissionUtils.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/PermissionUtils.java
@@ -94,6 +94,10 @@ public class PermissionUtils {
         return isInGroup(user, UserGroup.SECURITY_USER);
     }
 
+    public static boolean isViewer(User user) {
+        return isInGroup(user, UserGroup.VIEWER);
+    }
+
     public static boolean isSecurityAdminBySecondaryRoles(Set<UserGroup> roles) {
         return roles.contains(UserGroup.SECURITY_ADMIN);
     }
@@ -124,6 +128,8 @@ public class PermissionUtils {
                 return isAdmin(user);
             case SECURITY_USER:
                 return isSecurityUser(user);
+            case VIEWER: // VIEWER is lowest role — any authenticated role satisfies "at least VIEWER"
+                return isNormalUser(user) || isAdmin(user) || isClearingAdmin(user) || isEccAdmin(user) || isSecurityAdmin(user) || isSecurityUser(user) || isViewer(user);
             default:
                 throw new IllegalArgumentException("Unknown group: " + group);
         }
@@ -146,6 +152,13 @@ public class PermissionUtils {
                 return isAdminBySecondaryRoles(secondaryRoles);
             case ADMIN:
                 return isAdminBySecondaryRoles(secondaryRoles);
+            case VIEWER: // VIEWER is lowest role — any higher role satisfies "at least VIEWER"
+                return secondaryRoles.contains(UserGroup.VIEWER)
+                        || isNormalUserBySecondaryRoles(secondaryRoles)
+                        || isAdminBySecondaryRoles(secondaryRoles)
+                        || isClearingAdminBySecondaryRoles(secondaryRoles)
+                        || isEccAdminBySecondaryRoles(secondaryRoles)
+                        || isSecurityAdminBySecondaryRoles(secondaryRoles);
             default:
                 throw new IllegalArgumentException("Unknown role: " + role);
         }
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
@@ -9,6 +9,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -23,19 +24,36 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 
 import lombok.NonNull;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
+import org.eclipse.sw360.datahandler.permissions.PermissionUtils;
 import org.eclipse.sw360.datahandler.thrift.users.User;
-import org.eclipse.sw360.datahandler.thrift.users.UserGroup;
 import org.eclipse.sw360.rest.resourceserver.core.RestControllerHelper;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
-import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.filter.OncePerRequestFilter;
 
 
 @Component
 public class EndpointsFilter extends OncePerRequestFilter {
+
+    private static final Logger log = LogManager.getLogger(EndpointsFilter.class);
+
+    private static final List<String> MUTATION_METHODS = List.of("POST", "PATCH", "PUT", "DELETE");
+
+    /** GET paths blocked for VIEWER — defense-in-depth alongside controller guards. */
+    private static final List<String> VIEWER_BLOCKED_GET_PREFIXES = List.of(
+            "/api/clearingrequest",
+            "/api/moderationrequest",
+            "/api/ecc",
+            "/api/vulnerabilities",
+            "/api/obligations"
+    );
+
+    /** POST endpoints whitelisted for VIEWER (e.g., READ token creation). */
+    private static final String VIEWER_TOKEN_ENDPOINT = "/api/users/tokens";
 
     @Value("${blacklist.sw360.rest.api.endpoints}")
     String endpointsTobeBlackListed;
@@ -47,7 +65,6 @@ public class EndpointsFilter extends OncePerRequestFilter {
         this.restControllerHelper = restControllerHelper;
     }
 
-
     @Override
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
             throws ServletException, IOException {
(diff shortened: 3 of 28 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
```diff
@@ -57,16 +57,17 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
         boolean isAMatch = verifyMatchingOfRequestURIToEndpoints(requestURI, endpointHttpMethods);
+        User user = restControllerHelper.getSw360UserFromAuthentication();
 
         if (Arrays.asList("POST", "PATCH", "PUT", "DELETE").contains(method) && !isAMatch) {
-            User user = restControllerHelper.getSw360UserFromAuthentication();
-
-            // Inline check for Security User role having read only access
-            if (user.getUserGroup().name().equals("SECURITY_USER")) {
+            // Inline check for Security User role and Viewer role having read only access
+            if (user.getUserGroup().name().equals("SECURITY_USER") && user.getUserGroup().name().equals("VIEWER")) {
```
GMishx:
```suggestion
            if (user.getUserGroup().name().equals("SECURITY_USER") || user.getUserGroup().name().equals("VIEWER")) {
```

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
```diff
@@ -57,16 +57,17 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
                 response.sendError(HttpStatus.SERVICE_UNAVAILABLE.value());
             } else {
                 filterChain.doFilter(request, response);
             }
+        } else if (user != null && user.getUserGroup() != null && user.getUserGroup().name().equals("VIEWER") && (requestURI.contains("/projects") || requestURI.contains("/clearingrequest") || requestURI.contains("/moderationrequest") || requestURI.contains("/api/ecc") || requestURI.contains("/vendors") || requestURI.contains("/api/obligations"))) {
```
GMishx: `/vendors` would be required while loading a component/release right as they can have a vendor and this role can load their information.

GMishx (changes requested):
A question.

rudra-superrr:
Made the changes.

GMishx (dismissed):
Changes looks good.

keerthi-bl:
@rudra-superrr 

Resolve the conflicts and can you please add steps to test this PR in the description.

backend/vulnerabilities/src/main/java/org/eclipse/sw360/vulnerabilities/VulnerabilityHandler.java
```diff
@@ -71,7 +71,7 @@ public VulnerabilityHandler() throws IOException, SW360Exception {
 
     @Override
     public List<VulnerabilityDTO> getVulnerabilitiesByReleaseId(String releaseId, User user) throws TException {
-        if (!PermissionUtils.isUserAtLeast(UserGroup.USER, user)) {
+        if (!PermissionUtils.isUserAtLeast(UserGroup.VIEWER, user)) {
```
bibhuti230185: You changed the minimum role check from UserGroup.USER to UserGroup.VIEWER . This has the opposite of the intended effect:

isUserAtLeast(UserGroup.USER, user) → returns false for a VIEWER → VIEWER gets empty list ✅
isUserAtLeast(UserGroup.VIEWER, user) → returns true for every user including VIEWER → VIEWER now gets full vulnerability data ❌
Since VIEWER is the lowest role, isUserAtLeast(VIEWER, user) is always true for any authenticated user.

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/PermissionUtils.java
```diff
@@ -122,6 +126,8 @@ public static boolean isUserAtLeast(UserGroup group, User user) {
                 return isAdmin(user);
             case SECURITY_USER:
                 return isSecurityUser(user);
+            case VIEWER:
```
bibhuti230185: The new VIEWER case is added to isUserAtLeast but is missing from isUserAtLeastDesiredRoleInSecondaryGroup . This method is called wherever secondary department roles are evaluated. If any code path calls isUserAtLeastDesiredRoleInSecondaryGroup(UserGroup.VIEWER, ...) it will throw IllegalArgumentException at runtime — a silent crash.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
```diff
@@ -57,16 +57,17 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
         String[] endpointMethodPairs = endpointsTobeBlackListed.split(",");
         Map<String, Set<String>> endpointHttpMethods = getMapOfEndpointToHttpMethods(endpointMethodPairs);
         boolean isAMatch = verifyMatchingOfRequestURIToEndpoints(requestURI, endpointHttpMethods);
+        User user = restControllerHelper.getSw360UserFromAuthentication();
```
bibhuti230185: getSw360UserFromAuthentication() makes a Thrift/DB call. You've now added it to every API request (including all GET requests from normal users). This is a performance hit.  Keep the user fetch inside the  conditional branches where it's actually needed. Here add a null check: user != null && user.isSetUserGroup().

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
```diff
@@ -57,16 +57,17 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
 
         if (Arrays.asList("POST", "PATCH", "PUT", "DELETE").contains(method) && !isAMatch) {
-            User user = restControllerHelper.getSw360UserFromAuthentication();
-
-            // Inline check for Security User role having read only access
-            if (user.getUserGroup().name().equals("SECURITY_USER")) {
+            // Inline check for Security User role and Viewer role having read only access
+            if (user.getUserGroup().name().equals("SECURITY_USER") || user.getUserGroup().name().equals("VIEWER")) {
```
bibhuti230185: Better to use like  user.getUserGroup() == UserGroup.SECURITY_USER || user.getUserGroup() == UserGroup.VIEWER.  user.getUserGroup().name() will throw NullPointerException if getUserGroup() is null (possible for a brand-new user before Keycloak role sync).

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
```diff
@@ -57,16 +57,17 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
                 response.sendError(HttpStatus.SERVICE_UNAVAILABLE.value());
```
bibhuti230185: 503 means "server temporarily down". For an authorization denial, use 403 Forbidden. This is confusing because throwIfReadOnlyUser() already returns 403 via Spring's AccessDeniedException. Same rejection, different status codes.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/filter/EndpointsFilter.java
```diff
@@ -57,16 +57,17 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
             } else {
                 filterChain.doFilter(request, response);
             }
+        } else if (user != null && user.getUserGroup() != null && user.getUserGroup().name().equals("VIEWER") && (requestURI.contains("/projects") || requestURI.contains("/clearingrequest") || requestURI.contains("/moderationrequest") || requestURI.contains("/api/ecc") || requestURI.contains("/api/vendors") || requestURI.contains("/api/obligations"))) {
```
bibhuti230185: 1.  /api/vendors must be removed: Vendor data is embedded in component/release responses. Blocking vendor reads breaks component and release detail pages for VIEWER. (Already flagged by @GMishx.)
2.  /projects should not be fully blocked:  VIEWER can see projects with visibility = EVERYONE (summary + attachments). You need to allow GET /projects and GET /projects/{id} but filter the results by visibility in the controller/permissions layer — not block the entire URL.
3. /api/vulnerabilities is missing from the blocklist: No vulnerability data at all for VIEWER. 
4. contains("/projects") is too broad: Any URL containing the substring "projects" matches. Use startsWith("/api/projects") for precision.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vulnerability/VulnerabilityController.java
bibhuti230185, line 1: Add throwIfReadOnlyUser(user) to following endpoints:

getVulnerabilities() at L120 — GET /vulnerabilities
getVulnerability() at L158 — GET /vulnerabilities/{id}

bibhuti230185 (changes requested):
Few more areas to look for 

1. ProjectPermissions.java — isVisible() needs VIEWER awareness
2. ProjectController.java — VIEWER sees full project objects ,needs field stripping
3. `ProjectController.java` — Missing VIEWER guards on multiple project endpoints

| Endpoint | Line | What's missing |
|---|---|---|
| `GET /projects/{id}/licenseClearing` | L457 | default should be blocked — add `throwIfReadOnlyUser` with a `// TODO: revisit after License Clearing decision` comment |
| `GET /projects/{id}/linkedProjects` | L580 | VIEWER must not see linked projects — add `throwIfReadOnlyUser` |
| `GET /projects/{id}/linkedProjects/releases` | L629 | Same — add `throwIfReadOnlyUser` |

---
4. `SW360ReportController.java` L122 — Report exports need per-module VIEWER guard

    **Requirement**: *"Exports of Excel Sheet are allowed, ReadMeOSS are allowed (from License Clearing Tab), Product Clearing not allowed (from Obligations Tab)."*

    `GET /reports` dispatches to different report types based on a `module` parameter. The requirement says:
    - ✅ Allowed: `projects`, `components`, `licenses` (Excel sheets), `licenseInfo` (ReadMeOSS)
    - ❌ Blocked: `exportCreateProjectClearingReport` (Product Clearing), `licenseResourceBundle`, `projectReleaseSpreadsheetWithEccInfo` (ECC)
    - ⚠️ SBOM: needs discussion

    Currently there's no VIEWER check at all — every module is accessible.

5. `SearchController.java` L98 — Search must respect VIEWER restrictions

    **Requirement**: *"Search should follow the same restrictions like normal view, Projects with 'Everyone' visibility, all Components & Releases."*

    `GET /search` passes all `typeMasks` to the search service without filtering.

    For VIEWER this means: components ✅, releases ✅, licenses ✅, projects (EVERYONE only) ⚠️, vulnerabilities ❌, users ❌.
6. "UserController.java" : POST /users/tokens creates API tokens. Currently the EndpointsFilter blocks all POST requests for VIEWER . This means VIEWER cannot create any token — including READ tokens, which the requirement explicitly allows.
Additionally, even if the POST were allowed, there's no check preventing VIEWER from requesting WRITE authorities in the token body. The authorities field in the request is user-supplied.
7. In PermissionUtils.java, DEFAULT_USER_GROUP = UserGroup.USER. Per R8, new org-wide users should default to VIEWER. This needs to change to UserGroup.VIEWER
8. Keycloak integration — not present
VIEWER group in Keycloak realm
Sw360UserStorageProviderFactory update to assign VIEWER by default
Department-based USER promotion (configurable via orgmapping.properties from PR #3594)

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -462,6 +463,7 @@ public ResponseEntity<HalResource<Project>> licenseClearing(
 
         final User sw360User = restControllerHelper.getSw360UserFromAuthentication();
         restControllerHelper.throwIfSecurityUser(sw360User);
+        restControllerHelper.throwIfViewerUser(sw360User); // VIEWER can only see summary + attachments tabs
```
GMishx: I guess there was a small miss-understanding, from discussions, `VIEWER` is allowed to see the Linked Projects and Releases for the Project.

GMishx (changes requested):
I guess only one miss-understanding happened here. Rest looks good.

GMishx:
A quick question, you'd be implementing the changes for KeyCloak in this branch or in a separate PR? I'd prefer a separate PR to keep it logically separated as well.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -641,6 +646,7 @@ public ResponseEntity<CollectionModel<EntityModel<Release>>> getReleasesOfLinked
     ) throws TException, URISyntaxException, PaginationParameterException, ResourceClassNotFoundException {
 
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
+        restControllerHelper.throwIfViewerUser(sw360User); // VIEWER must not see linked projects
```
GMishx: This would still have to be removed?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
GMishx, line 1: Please block following endpoints for viewer role as well:
- /vulnerabilitySummary
- /vulnerabilities

GMishx (changes requested):
Final few changes still needed

GMishx:
Few changes to be added for KeyCloak setup:
1. Check if the user department is marked for viewer only with OrgMapper while creating the user in KeyCloak event listener. If yes, then update the user role to `VIEWER`.
2. Add the group in `third-party/keycloak-tf` for `/VIEWER`.

PR #3191, opened 2025-06-03
feat(config): move more properties to ConfigDB

GMishx:
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

1. Move Scheduler configs to configs database.
2. Create separate containers for configs.
3. Create new functions for container specific config manipulation.
4. Create new endpoint `/configurations/container/{configFor}` `GET` and `PATCH` for easy/scoped manipulations.
5. Add `UI_CONFIGURATION` container for UI specific properties and add properties from old frontend project.

### Suggest Reviewer
@hoangnt2 @heliocastro

### How To Test?
1. Check if new endpoints are working.
2. Check if new properties can be updated.

PR #3244, opened 2025-06-27
fix(resource): no config read at init

GMishx:

Fix the race-condition while trying to read config properties moved to Config service. Services `scheduler` and `resource` required some of these properties which are moved to the config service at their startup.

They cannot start unless they get these values, and tomcat cannot continue startup unless these service continue, and these services cannot continue because they need value from config service which cannot respond unless tomcat finish the startup.

This PR reverts the changes from Scheduler service done in #3191
And it removes `DOMAIN` property read in resource service which is not used.

These changes should prevent the startup race condition.

### How to test
1. Check if revert of scheduler service works.
2. Check if tomcat startup does not have any pending boots.


nikkuma7:
@GMishx , I have tested this pr, its working fine. as well its take less time to deploy on the tomcat server.

PR #3267, opened 2025-07-12
test(integration): added user integration tests in UserTest

rohit11544:
This PR adds the IT for user endpoints.

Summary: As part of migrating the REST API test cases from Spring-Doc to open API, have migrated all the Integration tests for user endpoints from UserSpecTest to UserTest. 

Issue: test coverage

Suggest Reviewer
@GMishx

How To Test?
The CI pipelines should pass and that should be enough as there are no changes in the source code.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/UserTest.java
+++ b/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/UserTest.java
@@ -10,19 +10,26 @@
 
 package org.eclipse.sw360.rest.resourceserver.integration;
 
-import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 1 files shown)


rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/UserTest.java
```diff
@@ -13,16 +13,25 @@
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
```
GMishx: `assertThat` is deprecated. Please migrate to `assertTrue/False/Equals`. See #3278

GMishx (changes requested):
One minor API migration needed. Rest looks good.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/UserTest.java
rohit11544: Done, removed assertThat and replace with assertTrue/False/Equals.

GMishx (approved):
LGTM

PR #3279, opened 2025-07-17
fix(sbom): add support for older CycloneDX XML/JSON format

Suhas2109:
This PR enhances the SBOM file import functionality to support older versions of CycloneDX in both .xml and .json formats (v1.1, v1.2, v1.3, etc.).

Previously, the import process rejected valid SBOM files using older CycloneDX schemas. With this update:

- .xml and .json files with schema versions like 1.1, 1.2, 1.3 are now validated and accepted correctly.
- SPDX import logic remains untouched.

Testing:
- Navigate to the SBOM import section.
- Upload CycloneDX .xml or .json files with schema version 1.1, 1.2, 1.3.
- Ensure the import succeeds and shows correct results.


GMishx (approved):
LGTM

PR #3283, opened 2025-07-19
fix(osadl): Fixed the import server error. 

### LINKED ISSUE #3248: OSADL Import Server Error
### Description

Hitting the endpoint /licenses/import/OSADL always returns internal server error.

Image

#### SW360 logs

Image

ritankarsaha:

## Summary 

- Hitting the endpoint /licenses/import/OSADL doesn't return internal server error.

Behavior After Fix:
  - The endpoint should now return HTTP 200 instead of 500
  - The response now includes a proper RequestSummary with:
    - Success message: "OSADL license has imported successfully"
    - Status: SUCCESS
    - Count of licenses processed
  - Failed license imports are logged as warnings instead of causing crashes.
  - Licenses that can't be parsed from OSADL are added to the "missing" list.

  Log Output:
  - Warning logs for licenses that can't be fetched/parsed from OSADL
  - Success logs for licenses that are imported correctly
  - No more NullPointerException stack traces

Issue: #3248 

### Suggest Reviewer

@GMishx 
@amritkv 
@rudra-superrr 

### How To Test?

- test the endpoint via swagger or curl comamnds. 

```bash
curl -X 'POST' \
    'http://127.0.0.1:8080/resource/api/licenses/import/OSADL' \
    -H 'accept: application/hal+json' \
    -H 'Authorization: Basic YWRtaW5Ac3czNjAub3JnOjEyMzQ1' \
    -d ''
```

response format - 

```bash
{
  "requestStatus" : "PROCESSING",
  "totalAffectedElements" : 0,
  "totalElements" : 0,
  "message" : "OSADL license has imported successfully",
  "setTotalAffectedElements" : false,
  "setTotalElements" : false,
  "setMessage" : true,
  "setRequestStatus" : true
}
```

It will prolly take up some time to sync up, but yep response format is getting generated. 

Fixes #3248 

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/LicenseTest.java
+++ b/rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/integration/LicenseTest.java
@@ -83,7 +83,9 @@ public class LicenseTest extends TestIntegrationBase {
         // Setup test request summary
         testRequestSummary = new RequestSummary();
         testRequestSummary.setRequestStatus(RequestStatus.SUCCESS);
-        testRequestSummary.setMessage("Operation completed successfully");
+        testRequestSummary.setMessage("{\"licensesSuccess\":0,\"licensesMissing\":0}");
+        testRequestSummary.setTotalAffectedElements(0);
+        testRequestSummary.setTotalElements(0);
 
         // Setup object mapper
         objectMapper = new ObjectMapper();
@@ -421,7 +423,7 @@ public class LicenseTest extends TestIntegrationBase {
         String responseBody = response.getBody();
         assertTrue("Response should contain request status", responseBody.contains("requestStatus"));
         assertTrue("Response should contain SUCCESS status", responseBody.contains("SUCCESS"));
-        assertTrue("Response should contain OSADL import message", responseBody.contains("OSADL license has imported successfully"));
+        assertTrue("Response should contain OSADL import message", responseBody.contains("licensesSuccess"));
         assertTrue("Response should contain total elements", responseBody.contains("totalElements"));
         assertTrue("Response should contain total affected elements", responseBody.contains("totalAffectedElements"));
     }


backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
```diff
@@ -1251,11 +1272,17 @@ private String addNodeElement(JSONArray jsonArray, User user) throws SW360Except
...
             ObligationNode obligationNode = new ObligationNode();
             obligationNode.setNodeType(jsonArray.getString(0));
             obligationNode.setNodeText(jsonArray.getString(1));
             return addObligationNodes(obligationNode, user);
+        } else {
+            log.warn("Unexpected JSON array structure with length " + jsonArray.length() + ": " + jsonArray.toString());
+            ObligationNode obligationNode = new ObligationNode();
+            obligationNode.setNodeType("UNKNOWN");
+            obligationNode.setNodeText(jsonArray.length() > 0 ? jsonArray.getString(0) : "");
+            return addObligationNodes(obligationNode, user);
```
GMishx: There should not be an `UNKNOWN` obligation node added in case conditions above failed. Please just put a log and continue.

ritankarsaha: fixed, removing this block and returning null then, is this okay ?

ritankarsaha:
Fixed, squashed and pushed commits.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
GMishx: Throwing an exception with a log would be best. But returning null should be okay too.

GMishx (dismissed):
Changes looks good.

GMishx:
@deo002 please help test the fix.

rudra-superrr:
Below are the few points found from testing:

- The process remains in the 'processing' state due to the high latency—each license takes nearly a minute to process
image

- The status code should not be 200 for the PROCESSING status but 202 as the request has been accepted for processing, but the processing is not completed yet. Also update the message field for PROCESSING status.
Currently, each time the endpoint is called, the processing restarts from scratch. Ideally, it should return the already processed result. Despite logs showing that some licenses were processed, the totalAffectedElements field always returns 0
image

- Also as you mentioned that Licenses that can't be parsed from OSADL are added to the "missing" list but this missing list is absent in response as well as in logs.

ritankarsaha:
> Below are the few points found from testing:
> 
> * The process remains in the 'processing' state due to the high latency—each license takes nearly a minute to process
> 
> image
> * The status code should not be 200 for the PROCESSING status but 202 as the request has been accepted for processing, but the processing is not completed yet. Also update the message field for PROCESSING status.
>   Currently, each time the endpoint is called, the processing restarts from scratch. Ideally, it should return the already processed result. Despite logs showing that some licenses were processed, the totalAffectedElements field always returns 0
> 
> image
> * Also as you mentioned that Licenses that can't be parsed from OSADL are added to the "missing" list but this missing list is absent in response as well as in logs.

making the changes and pushing a new commit into the PR.

GMishx:
@rudra-superrr , I have modified the code to use Flux and Mono from Spring to make the code more reactive. It should now fetch all obligations in parallel and provide better error handling.

Can you please test it again to check if the mentioned issues still persists?

Note: I could not work on the request to return 201 and do the processing in background.

rudra-superrr:
Hi @GMishx , I’m seeing the below error in the log, and it appears for all the affected licenses (694 in total).

image

image

GMishx:
Hey @amritkv , does this PR works for the new UI?

PR #3312, opened 2025-07-31
Fix NPE During License Info Generation; Support Projects With and Without Linked Projects

Suhas2109:
This PR fixes a NullPointerException that occurred during license info generation for projects that had linkedProjects set to null.
**Key Enhancements:**
Safely handles both cases

- Projects with linked projects.

- Projects without any linked projects

**Improved the /attachmentUsage API:**

- Made the transitive query parameter is now optional with a default value (false) to prevent 400 Bad Request.

**Bug Fixed:**
NPE when generating license info for projects with null linkedProjects.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -1197,7 +1197,7 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
         String parentProjectKey = sw360Project.getName() + " " + sw360Project.getVersion();
         vulnerabilitiesMap.put(parentProjectKey, parentProjectVulnerabilities);
 
-        if (!sw360Project.getLinkedProjects().isEmpty()) {
+        if (sw360Project.getLinkedProjects() != null && !sw360Project.getLinkedProjects().isEmpty()) {
             for (String linkedProjectId : sw360Project.getLinkedProjects().keySet()) {
                 Project linkedProject = projectService.getProjectForUserById(linkedProjectId, sw360User);
                 List<VulnerabilityDTO> linkedProjectVulnerabilities = vulnerabilityService.getVulnerabilitiesByProjectId(linkedProjectId, sw360User);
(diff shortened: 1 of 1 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -1664,7 +1664,6 @@ public ResponseEntity<CollectionModel<EntityModel<Attachment>>> getProjectAttach
             @PathVariable("id") String id
     ) throws TException {
         final User sw360User = restControllerHelper.getSw360UserFromAuthentication();
-        restControllerHelper.throwIfSecurityUser(sw360User);
```
GMishx: Why this statement was removed?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -2122,7 +2121,7 @@ public ResponseEntity attachmentUsages(
             )
             @RequestParam(value = "filter", required = false) String filter,
             @Parameter(description = "Get the transitive releases.")
-            @RequestParam(value = "transitive", required = true) boolean transitive
+            @RequestParam(value = "transitive", required = false, defaultValue = "false") boolean transitive
```
GMishx, line 2125: This change is not documented in the PR.

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
Suhas2109: It was unintentional, I'll revert this line.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
Suhas2109: The update to the transitive parameter in /attachmentUsage (making it optional with a default of false) is now clearly documented in the "API Improvement" section of the PR description.

GMishx (approved):
LGTM

PR #3334, opened 2025-08-11
feat(ci): use couchdb as actions service

GMishx:

Use couchdb container as an action service instead of running it on same container.


deo002 (approved):
Changes look good

PR #3377, opened 2025-09-02
fix(REST): 500 internal server error on deleting a pkg linked to a re…

afsahsyeda:
**Description**

1. Deleting a package involves a simple removal of the relevant packageId from the release doc.
2. We can skip the unnecessary checks in the updateRelease method of compDBHandler and call update of release repo directly from PackageDBHandler class.
image


akshitjoshii:
Test was successful.

PR #3456, opened 2025-10-14
feat(ImportCDX): remove VCS URL redirection logic

afsahsyeda:
- VCS redirection during import causes duplicate components to be created in SW360, as a new component would be created if the VCS of an existing component now points to a new location.
- Feedback regarding the urls redirected is given after the new component has been created, leaving no room for corrections, which depletes transparency between the SBOM being imported and the components being created in SW360.
- The SRC upload service of SW360 uploads the latest source even for old VCS URLs, removing the need to have the latest VCS URL.

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/common/utils/RepositoryURL.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/common/utils/RepositoryURL.java
@@ -10,7 +10,6 @@
 
 package org.eclipse.sw360.common.utils;
 
-import java.io.IOException;
 import java.net.*;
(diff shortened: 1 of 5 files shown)


afsahsyeda:
@Suhas2109 Please perform a general test and code review of this PR.

Suhas2109:
@afsahsyeda 
Tested SBOM import with CycloneDX.
SW360 now correctly retains the original VCS URLs during import.
The PR works as expected, redirection logic is no longer applied.

Suhas2109:
Changes LGTM.
Test was successful.

PR #3459, opened 2025-10-17
fix(Component): Move comment field form requestParam to requestBody

bibhuti230185:

## Summary of Changes

- Optimized `patchComponent()` method in ComponentController with request body to have comment instead of param
- Added input validation for null/empty component ID and DTO

## 2) How to Test

Test the API using the  `PATCH /api/components/{id}` 

Issue: 
https://github.com/eclipse-sw360/sw360/issues/3454

### Suggest Reviewer
@GMishx  @amritkv

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
@@ -72,6 +72,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.rest.webmvc.BasePathAwareController;
 import org.springframework.data.rest.webmvc.RepositoryLinksResource;
+import org.springframework.data.rest.webmvc.ResourceNotFoundException;
 import org.springframework.hateoas.*;
 import org.springframework.hateoas.server.RepresentationModelProcessor;
 import org.springframework.http.HttpStatus;
@@ -80,7 +81,6 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.web.bind.annotation.*;
-import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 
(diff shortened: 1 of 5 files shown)


amritkv:
Hey @bibhuti230185 !
I'm getting a bit weird response as follows :
```
{
    "timestamp": "2025-10-23T07:34:31.570484369Z",
    "status": 400,
    "error": "Bad Request",
    "message": "202 ACCEPTED \"Moderation request is created\""
}
```

bibhuti230185:
@amritkv  As discussed please rebuild your code and test. As shown its working fine in my local system as expected
image

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -392,29 +393,73 @@ public ResponseEntity<CollectionModel<EntityModel<Component>>> searchByExternalI
...
         }
-        if (updateComponentDto.getAttachments() != null) {
-            updateComponentDto.getAttachments().forEach(attachment -> wrapSW360Exception(
-                    () -> this.attachmentService.fillCheckedAttachmentData(attachment, user)));
+    }
+    
+    private void validateModerationRequest(Component component, User user, String comment) {
+        if (!restControllerHelper.isWriteActionAllowed(component, user) && 
+            (comment == null || comment.isBlank())) {
+            throw new BadRequestClientException(RESPONSE_BODY_FOR_MODERATION_REQUEST_WITH_COMMIT.get("message"));
```
GMishx: The old code was returning a JSON object

```suggestion
            throw new BadRequestClientException(RESPONSE_BODY_FOR_MODERATION_REQUEST_WITH_COMMIT.toString());
```

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/component/ComponentController.java
```diff
@@ -392,29 +393,73 @@ public ResponseEntity<CollectionModel<EntityModel<Component>>> searchByExternalI
...
+            updateComponentDto.getAttachments().forEach(attachment -> 
+                wrapSW360Exception(() -> attachmentService.fillCheckedAttachmentData(attachment, user))
+            );
+        }
+    }
+    
+    private void handleUpdateStatus(RequestStatus status) {
+        switch (status) {
+            case SENT_TO_MODERATOR:
+                throw new ResponseStatusException(HttpStatus.ACCEPTED, RESPONSE_BODY_FOR_MODERATION_REQUEST.get("message"));
```
GMishx: This is the issue why @amritkv is getting wrong response @bibhuti230185 . Exceptions are used for wrong code flow, not for success.

As per old code, this function should return following response in case status is `SENT_TO_MODERATOR`
```java
return new ResponseEntity(RESPONSE_BODY_FOR_MODERATION_REQUEST, HttpStatus.ACCEPTED);
```
And following in otherwise:
```java
return new ResponseEntity<>(userHalResource, HttpStatus.OK);
```

And handle the other cases as exceptions if necessary.

GMishx (changes requested):

PR #3461, opened 2025-10-17
feat(rest): Api endpoint to retrieve package usage count

Farooq-Fateh-Aftab:

> * Which issue is this pull request belonging to and how is it solving it? https://github.com/eclipse-sw360/sw360/issues/3418
> * Did you add or update any new dependencies that are required for your change? No

Closes https://github.com/eclipse-sw360/sw360/issues/3418

### How To Test?
API : GET 
URL : http://localhost:8080/resource/api/packages/{id}/usage
Path Variable : id= packageId
Response : 
{
  "count": 1,
  "isUsed": true
}

### Checklist
Must:

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
@@ -220,7 +220,7 @@ public class ProjectRepository extends SummaryAwareRepository<Project> {
             "function(doc) {" +
                     "  if (doc.type == 'project' && doc.packageIds) {" +
                     "    for(var i in doc.packageIds) {" +
-                    "      emit(doc.packageIds[i], doc._id);" +
+                    "      emit(i, doc._id);" +
                     "    }" +
                     "  }" +
                     "}";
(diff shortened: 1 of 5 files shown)


rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/PackageSpecTest.java
```diff
@@ -458,11 +463,42 @@ public void should_document_update_package() throws Exception {
                         ));
     }
 
-    @Test
```
GMishx, line 461: Is this some unintentional change?

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
```diff
@@ -220,7 +220,7 @@ public class ProjectRepository extends SummaryAwareRepository<Project> {
             "function(doc) {" +
                     "  if (doc.type == 'project' && doc.packageIds) {" +
                     "    for(var i in doc.packageIds) {" +
-                    "      emit(doc.packageIds[i], doc._id);" +
+                    "      emit(i, doc._id);" +
```
GMishx, line 223: Essentially `i` is an object now. I think the key of `i` should be emitted now? I am bit confused.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/restdocs/PackageSpecTest.java
Farooq-Fateh-Aftab: yes, it was unintentional. Now corrected

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
Farooq-Fateh-Aftab: Yes, doc.packageIds is an object (not an array as this has been changed from array to map).
i is the key of each property in the packageIds object—these keys are the package IDs.

PR #3490, opened 2025-11-10
feat(search): add sorting index for lucene search

GMishx:

Add new `string` type index which can be used for sorting for `text` type indexes.

As per [Nouveau docs](https://docs.couchdb.org/en/stable/ddocs/nouveau.html#field-types), `text` indexes are great for searching but cannot be used for sorting. On the other hand, `string` indexes are bad for searching but can be used for sorting.

This PR creates a `string` index for `text` indexes with `_sort` suffix to be used for sorting whenever making a lucene search. Index searches are already sorted.

### How To Test?
Search for a component or a project with `luceneSearch=true` and different allowed sorting columns.


amritkv (approved):
LGTM !

PR #3503, opened 2025-11-17
fix(rest) : Project Obligation should have default status "Open"

### LINKED ISSUE #3501: Project Obligation should have default status "Open"
Description
The response of GET /resource/api/projects/{id}/licenseObligations should have default status as "Open". Currently it is null and it causes wrong status in the UI.

#### How to reproduce

Create a project and fetch the Obligations, default status is null.
After updating the status, the feature works as expected.

keerthi-bl:

Issue: 

Closes #3501 

### How To Test?

1. Trigger the URL /resource/api/projects/{id}/licenseObligations
2. The data received should have status key with value 
3. If status has no value, by default "Open" should be available in the response. 

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -64,6 +64,7 @@ import org.eclipse.sw360.datahandler.thrift.RequestStatus;
 import org.eclipse.sw360.datahandler.thrift.RequestSummary;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
 import org.eclipse.sw360.datahandler.thrift.Source;
+import org.eclipse.sw360.datahandler.thrift.ObligationStatus;
 import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
 import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentContent;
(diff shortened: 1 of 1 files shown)


rudra-superrr:
Testing was successful.

image

image

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -50,20 +50,7 @@
...
-import org.eclipse.sw360.datahandler.thrift.MainlineState;
-import org.eclipse.sw360.datahandler.thrift.PaginationData;
-import org.eclipse.sw360.datahandler.thrift.ProjectReleaseRelationship;
-import org.eclipse.sw360.datahandler.thrift.ProjectPackageRelationship;
-import org.eclipse.sw360.datahandler.thrift.ReleaseRelationship;
-import org.eclipse.sw360.datahandler.thrift.RequestStatus;
-import org.eclipse.sw360.datahandler.thrift.RequestSummary;
-import org.eclipse.sw360.datahandler.thrift.SW360Exception;
-import org.eclipse.sw360.datahandler.thrift.Source;
+import org.eclipse.sw360.datahandler.thrift.*;
```
rudra-superrr: Ideally, all the required import files should be listed explicitly instead of using a generic `thrift.*` import.
Additionally, even after adding `thrift.*`, the code still includes separate imports for individual Thrift classes.
I suggest to remove the generic `thrift.* `import.

PR #3521, opened 2025-11-27
fix(rest): allow external license references in release updates

### LINKED ISSUE #3520: Make PATCH Release endpoint more resilient
### Description

From user:\
“Edit Release” works.\
You can enter links, upload attachments etc.\
But when you click “Update”, then you get an error message “License with ids \[GNU General Public License, version 2 with the GNU Classpath Exception\] does not exist in SW360 database", even if this license is not contained in the uploaded files.

#### How to reproduce

### Versions

* Docker version OR
* Last commit id on main

#### SW360 logs

nikkuma7:

Closes #3520 

### How To Test?
PATCH: http://localhost:8080/resource/api/releases/release{id}
Body: example

{
  "mainLicenseIds": [
    "GNU General Public License, version 2 with the GNU Classpath Exception"
  ],
  "name": "Test Release Update"
}

> How should these changes be tested by the reviewer?

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -20,6 +20,7 @@ import org.apache.thrift.TException;
 import org.apache.thrift.TFieldIdEnum;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Objects;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
 import org.eclipse.sw360.datahandler.permissions.PermissionUtils;
@@ -573,7 +574,7 @@ public class RestControllerHelper<T> {
             halLicense.add(licenseSelfLink);
             return halLicense;
         } catch (ResourceNotFoundException rne) {
-            LOGGER.error("cannot create a self link for license with id" + licenseId);
+            LOGGER.error("cannot create a self link for license with id " + licenseId);
             embeddedLicense.setShortname(licenseId);
             embeddedLicense.setOSIApproved(Quadratic.NA);
             embeddedLicense.setFSFLibre(Quadratic.NA);
@@ -581,7 +582,7 @@ public class RestControllerHelper<T> {
             embeddedLicense.setFullname(null);
             return halLicense;
         } catch (Exception e) {
-            LOGGER.error("cannot create self link for license with id: " + licenseId);
+            LOGGER.error("cannot create self link for license with id: " + licenseId); 
         }
         return null;
     }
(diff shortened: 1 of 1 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -573,17 +584,32 @@ private HalResource<License> addEmbeddedLicense(String licenseId) {
...
+                return halLicense;
+            } else {
+                LOGGER.error("cannot create self link for license with id: " + licenseId, e);
+                embeddedLicense.setShortname(licenseId);
+                embeddedLicense.setOSIApproved(Quadratic.NA);
+                embeddedLicense.setFSFLibre(Quadratic.NA);
+                embeddedLicense.setChecked(false);
+                embeddedLicense.setFullname(null);
+                return halLicense;
+            }
```
GMishx: The branch for `if` and `else` is exactly the same with only exception about fullname being `null` in `else` case. Instead, can the dummy `embeddedLicense` be returned like in `ResourceNotFoundException` case?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -806,20 +840,118 @@ public License convertLicenseFromRequest(Map<String, Object> reqBodyMap, License
...
         }
+
         if (!licenseIncorrect.isEmpty()) {
-            throw new BadRequestClientException("License with ids " + licenseIncorrect + " does not exist in SW360 database.");
+            String errorMsg = fieldName + " with ids " + licenseIncorrect + " does not exist in SW360 database.";
+            if (!licenseWarnings.isEmpty()) {
+                errorMsg += " Note: External license references " + licenseWarnings + " were allowed.";
+            }
+            LOGGER.error("License validation failed: {}", errorMsg);
+            throw new BadRequestClientException(errorMsg);
```
GMishx: The idea of being "Resilient" is **not to** throw exceptions. But ignore bad entries while entering them to DB and notify user. Throwing an error here will still prevent the user from storing the entire release update even if it contains one bad license.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -806,20 +840,118 @@ public License convertLicenseFromRequest(Map<String, Object> reqBodyMap, License
...
+
+        return isMatchingConfigurablePatterns(normalizedId) ||
+               isLongDescriptiveName(normalizedId);
+    }
+
+    /**
+     * Check against configurable license patterns from application properties
+     * This is now more conservative to avoid false positives with standard SPDX identifiers
+     */
+    private boolean isMatchingConfigurablePatterns(String normalizedId) {
```
GMishx: This function does not make sense to me. Can you please explain why are we looking for Microsoft/Oracle/Google etc in the license names and calling them external licenses?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -806,20 +840,118 @@ public License convertLicenseFromRequest(Map<String, Object> reqBodyMap, License
...
+        }
+
+        return false;
+    }
+
+    /**
+     * Check if it's likely a long descriptive name from external sources
+     * Only very long names are considered external to be conservative
+     */
+    private boolean isLongDescriptiveName(String normalizedId) {
```
GMishx: Same with this function. Why are we looking for arbitrary value of `40` in name lenght?

PR #3522, opened 2025-11-28
fix(vendors): fix search and pagination

bibhuti230185:

### Summary of changes :
 Enhanced the VendorRepository to use view-based pagination with sorting. The key improvements include:
View Selection Logic: The getViewFromPagination() method maps sort column numbers to the appropriate CouchDB views (vendorbyfullname, vendorbyshortname, all)
Optimized Pagination: Uses queryViewPaginated() for better performance when sorting, with a fallback to index-based queries for reliability

Issue: 

### Suggest Reviewer
@GMishx 

# Note:
These changes also incorporate suggestions from https://github.com/apache/couchdb/issues/5756

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -56,6 +56,7 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - uses: webiny/action-conventional-commits@8bc41ff4e7d423d56fa4905f6ff79209a78776c7 # v1.3.0
 
+
       - name: Verify license headers
         run: |
           chmod +x .github/testForLicenseHeaders.sh
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/VendorRepository.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/VendorRepository.java
@@ -11,13 +11,16 @@ package org.eclipse.sw360.datahandler.db;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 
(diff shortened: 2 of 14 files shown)


backend/common/src/main/java/org/eclipse/sw360/datahandler/db/VendorRepository.java
```diff
@@ -109,4 +125,72 @@ public Set<String> getVendorByLowercaseShortnamePrefix(String shortnamePrefix) {
...
+        return Collections.singletonMap(pageData, vendors);
+    }
+
+    /**
+     * Get Vendors matching exact values for the given subQueryRestrictions with pagination.
+     * @param subQueryRestrictions Map of field names to sets of values to match against.
+     * @param pageData Pagination data
+     * @return Map containing pagination data as key and list of vendors as value.
+     */
+    public Map<PaginationData, List<Vendor>> searchVendorByExactValues(
```
GMishx: Lets keep it simple. Vendors currently do not have advance search and thus no need of exact value search.

Lets go with `VendorSearchHandler.search()` for all searches (i.e. quick search in Admin > Vendors & Vendor search in Releases and Components pages).

GMishx: Anyways taken care by `VendorHandler.searchVendors()`. So we do not need this function.

GMishx (approved):
Changes looks good.

.github/workflows/build_and_test.yml
GMishx, line 1: Can you please revert these changes as they are not required anymore and they are slowing down the test cases.

PR #3537, opened 2025-12-12
feat(CR): Add pagination and filtering to clearing requests endpoint

akshitjoshii:
- Add database-level pagination using CouchDB views and PaginationData
- Support filtering by `status, projectId, createdBy, createdOn`
- Removed extra fields from payload which were not needed to improve API performance.

Test : 
http://localhost:8080/resource/api/clearingrequests


GMishx (approved):
LGTM.

PR #3543, opened 2025-12-20
fix: make project mainlineState field required

### LINKED ISSUE #3395: Project Mainline State Field Required
### Description

There is no field for Project Mainline State in the api response of /projects/{id}/licenseClearing. It is required for the License Clearing tables in Project details tab.

#### Screenshots

Image

taanvi2205:
Added projectMainlineState field to the API response. This field represents the mainline state of a release within a specific project, which is different from the release's own mainlineState.

Issue: Fixes #3395

### How to Test
1. Create a project and link a release to it
2. Call API: GET /projects/{id}/licenseClearing?transitive=true
3. Verify projectMainlineState appears in response under linkedReleases 

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
@@ -1556,6 +1556,7 @@ public class JacksonCustomizations {
                 "setLayer",
                 "setDefaultValue",
                 "setProjectId",
+                "setReleaseMainLineState",
                 "setComponentId",
                 "setCreatedBy",
                 "createdByIsSet"


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
GMishx, line 1: Please revert changes in this class. Nothing seems to be relevant for this PR.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -9,6 +9,7 @@
  */
 package org.eclipse.sw360.rest.resourceserver.core;
 
+import org.eclipse.sw360.datahandler.thrift.MainlineState;
```
GMishx: Irrelevant chagnes?

GMishx (changes requested):
Minor cleanup required. Rest looks good.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/JacksonCustomizations.java
taanvi2205: @GMishx I added these because I thought the new transient field projectMainlineState in Release (components.thrift:259) would need the corresponding setter/isSet methods to be excluded from Jackson serialization similar to how other fields are handled in this file.

taanvi2205: @GMishx Thanks for the review I added these because I think Thrift generates setProjectMainlineState() and projectMainlineStateIsSet() methods for the new field, and without adding them to JsonIgnoreProperties, Jackson might include them as extra properties in the API response. That's why I followed the same pattern as other fields in this file. Let me know if I'm misunderstanding

taanvi2205:
@GMishx please review it now

PR #3544, opened 2025-12-21
fix: skip license validation for unchanged licenses during release update

RajvardhanT7747:
When updating a Release via PATCH API, the system validates all license IDs even when they aren't being modified. This causes errors for licenses imported from external sources that don't exist in SW360's database.

The issue occurs in RestControllerHelper.updateRelease() which validates licenses whenever the fields are non-null, regardless of whether they're in the update request.

This PR modifies the validation to only check licenses when they're explicitly present in the request body.

Changes:
- Modified RestControllerHelper.updateRelease() to accept reqBodyMap parameter
- Only validate licenses when present in request body
- Updated ReleaseController.patchRelease() to pass reqBodyMap
- Added tests in ReleaseTest.java to verify the fix
- Fixed docker-compose.yml for CouchDB 3.0+ compatibility

Files changed:
- RestControllerHelper.java
- ReleaseController.java  
- ReleaseTest.java
- docker-compose.yml


GMishx (changes requested):
The PR contains lots of unnecessary formatting changes which makes the PR review very difficult. Please try to keep the changes to the absolute necessary.

GMishx:
Closing the PR as it is not reviewable.

PR #3570, opened 2026-01-06
fix(CR): Allow multiple state values in CR state param

akshitjoshii:
Extend  `GET /clearingrequests` endpoint to allow multiple state params to be passed for filtering based on CR status.

image

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
@@ -171,9 +171,8 @@ public class ClearingRequestController implements RepresentationModelProcessor<R
             Pageable pageable,
             @Parameter(description = "Project ID to filter")
             @RequestParam(value = "projectId", required = false) String projectId,
-            @Parameter(description = "Status of the clearing request",
-                    schema = @Schema(implementation = ClearingRequestState.class))
-            @RequestParam(value = "status", required = false) String status,
+            @Parameter(description = "Status of the clearing request (comma-separated for multiple values, e.g., 'NEW,ACCEPTED,IN_PROGRESS')")
+            @RequestParam(value = "status", required = false) Set<ClearingRequestState> status,
             @Parameter(description = "Requesting user email to filter")
             @RequestParam(value = "createdBy", required = false) String createdBy,
             @Parameter(description = "Date clearing request was created on (timestamp).")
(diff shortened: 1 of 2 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
```diff
@@ -171,7 +171,7 @@ public ResponseEntity<CollectionModel<?>> getClearingRequests(
             Pageable pageable,
             @Parameter(description = "Project ID to filter")
             @RequestParam(value = "projectId", required = false) String projectId,
-            @Parameter(description = "Status of the clearing request",
+            @Parameter(description = "Status of the clearing request (comma-separated for multiple values, e.g., 'NEW,ACCEPTED,IN_PROGRESS')",
                     schema = @Schema(implementation = ClearingRequestState.class))
             @RequestParam(value = "status", required = false) String status,
```
GMishx: What if we change it like following?
```java
@Parameter(description = "Status of the clearing request (comma-separated for multiple values, e.g., 'NEW,ACCEPTED,IN_PROGRESS')"
@RequestParam(value = "status", required = false) Set<ClearingRequestState> status,
```

It will make the Swagger Docs more accurate and easy to test.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
```diff
@@ -229,7 +229,11 @@ private Map<String, Set<String>> getFilterMapForClearingRequests(
             filterMap.put(ClearingRequest._Fields.PROJECT_ID.getFieldName(), Collections.singleton(projectId));
         }
         if (CommonUtils.isNotNullEmptyOrWhitespace(status)) {
-            filterMap.put(ClearingRequest._Fields.CLEARING_STATE.getFieldName(), Collections.singleton(status));
+            Set<String> statusValues = Arrays.stream(status.split(","))
+                    .map(String::trim)
+                    .filter(s -> !s.isEmpty())
+                    .collect(Collectors.toSet());
```
GMishx: If using the suggestion from above, this can be simplified as:
```java
Set<String> statusValues = status.stream().map(ClearingRequestState::toString).collect(Collectors.toSet());
```

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
```diff
@@ -229,7 +229,11 @@ private Map<String, Set<String>> getFilterMapForClearingRequests(
             filterMap.put(ClearingRequest._Fields.PROJECT_ID.getFieldName(), Collections.singleton(projectId));
         }
         if (CommonUtils.isNotNullEmptyOrWhitespace(status)) {
```
GMishx: This also has to be changed to check for null or empty collections.

GMishx (changes requested):
Idea suggestion for better Swagger Documentation.

akshitjoshii:
@GMishx , addressed the review comments.

Updated `swagger docs` : 

image

nikkuma7:
Test successful.

image

PR #3576, opened 2026-01-08
fix(LicenseInfo): null check for license text.

### LINKED ISSUE #3575: Handle null license text for license info report.
### Description

Report generation for the project is failing due to a NullPointerException. This occurs when the license present in an approved CLI attachment has a null value for the license text.

rudra-superrr:
Closes #3575 

Description: The code introduced some NullPointerException. This PR fixes it with

-> Check if LicenseNameWithText.getLicenseText() returns a null, make it an empty string.
-> Invert the filter from entry.getValue().equals(text) to text.equals(entry.getValue()) to handle null entry values.


GMishx (approved):
Changes looks good. Tested, working as expected.

PR #3590, opened 2026-01-14
feat(AttachmentUsages): Save existing attachment usages of sub-project at parent project level

afsahsyeda:
**Description**

1. Attachment Usages saved at a project should be preserved and transferred to the project it is being linked to.
2. This saves the user the effort of making selections at each sub-project level again.
3. Attachment usages are crucial to generate readme oss.

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -35,8 +35,7 @@ import org.eclipse.sw360.datahandler.entitlement.ProjectModerator;
 import org.eclipse.sw360.datahandler.permissions.PermissionUtils;
 import org.eclipse.sw360.datahandler.permissions.ProjectPermissions;
 import org.eclipse.sw360.datahandler.thrift.*;
-import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
-import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentContent;
+import org.eclipse.sw360.datahandler.thrift.attachments.*;
 import org.eclipse.sw360.datahandler.thrift.changelogs.ChangeLogs;
 import org.eclipse.sw360.datahandler.thrift.changelogs.Operation;
 import org.eclipse.sw360.datahandler.thrift.components.*;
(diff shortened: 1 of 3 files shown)


afsahsyeda:
@Suhas2109 please perform a general test on this PR.

GMishx:
@afsahsyeda this would not be a desired behavior as a Project can be used as sub-project by multiple Projects. For example:
* Project A -> (sub) Project B -> (sub) Project C
    - User updated Project A attachment usage, it is trickled down to Projects B & C.
* Project D -> (sub) Project C
    - Attachment usage from Project A is no longer usable here. Project D made some changes, Project C is updated again.
    - Project A's attachment usage are now no longer valid after update in Project D. Both projects have nothing to do with each other.

From my understanding, ReadmeOSS are generated by Project A & D and extremely rarely for Project C.

afsahsyeda:
> @afsahsyeda this would not be a desired behavior as a Project can be used as sub-project by multiple Projects. For example:
> 
> * Project A -> (sub) Project B -> (sub) Project C
>   
>   * User updated Project A attachment usage, it is trickled down to Projects B & C.
> * Project D -> (sub) Project C
>   
>   * Attachment usage from Project A is no longer usable here. Project D made some changes, Project C is updated again.
>   * Project A's attachment usage are now no longer valid after update in Project D. Both projects have nothing to do with each other.
> 
> From my understanding, ReadmeOSS are generated by Project A & D and extremely rarely for Project C.

Changes made in project A are not trickled down to Project B and C. They are limited to Project A.
If Project B becomes a sub-project of project D, only the attachment usages of project B are added to project D by default.
This is useful in cases where there are 1000+ releases per project and the user does not want to make selections for sub-projects again when they have already made them at the independent project level.

GMishx:
So the PR works from bottom-up not top-down? The code changes are bit confusing.

afsahsyeda:
> So the PR works from bottom-up not top-down? The code changes are bit confusing.

Yes it is bottom up. The code had to be written that way to create a project hierarchy string.

Suhas2109:
Test was successful!

GMishx:
Another quick question @afsahsyeda sorry I missed it earlier. There is this button in the old frontend
image
This allows users to choose and carry over the attachment usage.

But after your changes, it would be done by default. This would not be the behavior everyone is looking for.
Instead, I would recommend to create an endpoint which can do this operation (something like `POST /projects/{id}/attachmentUsage/copy/{sub-proj-id}`) and let the script/client call this when required. What do you think?

afsahsyeda:
> Another quick question @afsahsyeda sorry I missed it earlier. There is this button in the old frontend image This allows users to choose and carry over the attachment usage.
> 
> But after your changes, it would be done by default. This would not be the behavior everyone is looking for. Instead, I would recommend to create an endpoint which can do this operation (something like `POST /projects/{id}/attachmentUsage/copy/{sub-proj-id}`) and let the script/client call this when required. What do you think?

@GMishx I was not aware of this functionality in the old frontend! Yes, I think we should create an endpoint so that the same feature can be created in the new frontend as well.

afsahsyeda:
@GMishx The "Use selection of this subproject" feature only checks the boxes on the UI and makes no changes to the project's attachmentUsages in the DB.
The feature will have to be implemented in the same way on the new frontend as well, removing the need for an endpoint as an endpoint would make changes to the DB.

What I can do instead is make this feature configurable for the backend, with the default set to FALSE. Upon setting it to true, all parent projects will automatically inherit the attachmentUsages of their sub-projects once linked.

GMishx:
Understood @afsahsyeda . Making it configurable would make total sense in this case.

afsahsyeda:
@GMishx The feature has been made configurable now. Please have a look.

GMishx (approved):
Changes looks good.

PR #3591, opened 2026-01-14
fix: add batch endpoint for license clearing counts to remove 502 errors

taanvi2205:

Add batch endpoint POST /api/projects/licenseClearingCount to fix 502 errors when loading projects table with 100+ entries. Instead of N individual API calls,now frontend fetch clearing counts for all projects in one request.

Issue: eclipse-sw360/sw360-frontend#1341

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -2962,8 +2962,8 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
                 int approvedCount = clearingInfo.approved;
 
                 JsonObject row = new JsonObject();
-                row.addProperty("Release Count", releaseCount);
-                row.addProperty("Approved Count", approvedCount);
+                row.addProperty("totalCount", releaseCount);
+                row.addProperty("approvedCount", approvedCount);
                 result.add(proj.getId(), row);
             }
         }


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -2929,6 +2936,46 @@ public void getlicenseClearingCount(
...
+            ReleaseClearingStateSummary clearingInfo = proj.getReleaseClearingStateSummary();
+            if (clearingInfo != null) {
+                int releaseCount = clearingInfo.newRelease + clearingInfo.sentToClearingTool
+                        + clearingInfo.underClearing + clearingInfo.reportAvailable
+                        + clearingInfo.scanAvailable + clearingInfo.approved;
+                int approvedCount = clearingInfo.approved;
+
+                JsonObject row = new JsonObject();
+                row.addProperty("Release Count", releaseCount);
+                row.addProperty("Approved Count", approvedCount);
```
GMishx: Please use easy to use keys like `totalCount` and `approvedCount` instead of plain English.

GMishx: The same would have to be done for `/projects/{id}/licenseClearingCount`. But it should be a separated PR.

GMishx (changes requested):
Minor change. Rest looks good.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
taanvi2205: Okay

GMishx (approved):
Good to go.

PR #3596, opened 2026-01-18
feat(swagger): Add s runtime switch for swagger authentication

heliocastro:
Default to true for auth.

It can be overrided in runtime as:

* Environment var SPRINGDOC_SWAGGER_UI_REQUIRE_AUTHENTICATION=false

* Java property -Dspringdoc.swagger-ui.require-authentication=false


GMishx:
@heliocastro , I see this PR was self-merged without any review. This is in direct violation of the policy we agreed upon. See [Definition of Done](https://eclipse.dev/sw360/docs/development/dev-dod-and-style/), point 2 of heading Policy.

We request you to kindly refrain from doing so again in future.

PR #3597, opened 2026-01-18
fix(rest): prevent null in HAL response for missing licenses

airajena:
## Summary

This PR fixes a bug in `RestControllerHelper.addEmbeddedLicense()` where the method returns `null` when an unexpected exception occurs. This causes `null` entries to be embedded in HAL responses, breaking API contracts and causing potential NPEs in API consumers.

## Problem

When a license ID is referenced by a release but does not exist in the database (or any unexpected error occurs during license lookup), the current code:

1. Catches the generic `Exception`
2. Logs an error message
3. Returns `null`

This `null` value is then embedded into the HAL response, causing:
- Invalid JSON structure (nullable embedded resources)
- Downstream NPEs in clients parsing the response
- Difficult debugging for maintainers

**Before (problematic code):**
```java
} catch (Exception e) {
    LOGGER.error("cannot create self link for license with id: " + licenseId);
}
return null;
```

## Solution

Return a placeholder `HalResource<License>` with basic information instead of `null`:

**After (fixed code):**
```java
} catch (Exception e) {
    LOGGER.warn("Unexpected error while creating embedded license for id: {}", licenseId, e);
    // Return placeholder license to avoid null in HAL response
    embeddedLicense.setShortname(licenseId);
    embeddedLicense.setOSIApproved(Quadratic.NA);
    embeddedLicense.setFSFLibre(Quadratic.NA);
    embeddedLicense.setChecked(false);
    embeddedLicense.setFullname(null);
    return halLicense;
}
```

## Impact

- **No API breaking changes**: Method signature unchanged
- **No backward compatibility issues**: Response structure remains valid HAL
- **Improved reliability**: API consumers no longer receive null embedded resources
- **Better debugging**: Log message includes full stack trace with `WARN` level

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -570,8 +570,8 @@ public class RestControllerHelper<T> {
             Link licenseSelfLink = linkTo(UserController.class)
                     .slash("api" + LicenseController.LICENSES_URL + "/" + licenseById.getId()).withSelfRel();
             halLicense.add(licenseSelfLink);
-        } catch (ResourceNotFoundException | Exception e) {
-            LOGGER.error("cannot create a self link for license with id " + licenseId);
+        } catch (Exception e) {
+            LOGGER.error("cannot create a self link for license with id {}", licenseId);
             embeddedLicense.setShortname(licenseId);
             embeddedLicense.setOSIApproved(Quadratic.NA);
             embeddedLicense.setFSFLibre(Quadratic.NA);


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
```diff
@@ -580,9 +580,14 @@ private HalResource<License> addEmbeddedLicense(String licenseId) {
...
-            LOGGER.error("cannot create self link for license with id: " + licenseId);
+            LOGGER.warn("Unexpected error while creating embedded license for id: {}", licenseId, e);
+            embeddedLicense.setShortname(licenseId);
+            embeddedLicense.setOSIApproved(Quadratic.NA);
+            embeddedLicense.setFSFLibre(Quadratic.NA);
+            embeddedLicense.setChecked(false);
+            embeddedLicense.setFullname(null);
+            return halLicense;
         }
-        return null;
```
GMishx, line 585: The entire block can be simplified as following:
```java
        HalResource<License> halLicense = new HalResource<>(embeddedLicense);

        try {
            License licenseById = licenseService.getLicenseById(licenseId);
            embeddedLicense.setFullname(licenseById.getFullname());
            embeddedLicense.setShortname(licenseId);
            Link licenseSelfLink = linkTo(UserController.class)
                    .slash("api" + LicenseController.LICENSES_URL + "/" + licenseById.getId()).withSelfRel();
            halLicense.add(licenseSelfLink);
        } catch (ResourceNotFoundException | Exception rne) {
            LOGGER.error("cannot create a self link for license with id " + licenseId);
            embeddedLicense.setShortname(licenseId);
            embeddedLicense.setOSIApproved(Quadratic.NA);
            embeddedLicense.setFSFLibre(Quadratic.NA);
            embeddedLicense.setChecked(false);
            embeddedLicense.setFullname(null);
        }
        return halLicense;
```

airajena: Pushed the suggested simplification.

PR #3614, opened 2026-01-22
fix(rest): prevent silent data loss in obligation update

Aman-Cool:
### What’s broken

`Sw360ObligationService.updateObligation()` unconditionally overwrites an obligation’s node with the return value of `addNodes()`.

`addNodes()` can return `null` (permission denied, malformed JSON, unexpected structure). When that happens, the existing node tree is **silently wiped**, persisted, and the REST API still returns HTTP 200.

This causes **permanent obligation data loss** and broken compliance exports.

### Why this matters

Obligation nodes are core compliance data. Losing them silently breaks trust in SW360 and can impact legal/compliance decisions.

This is a **regression** introduced by commit `16a36f1cf`.

### Fix

Fail fast if `addNodes()` returns `null`:

* Do **not** mutate or persist the obligation
* Return an error instead of reporting success

### Impact

Prevents silent data corruption and restores atomic, predictable obligation updates.

Low-risk, minimal change, failure cases only.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/obligation/Sw360ObligationService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/obligation/Sw360ObligationService.java
@@ -26,7 +26,10 @@ import org.eclipse.sw360.datahandler.thrift.licenses.ObligationLevel;
 import org.eclipse.sw360.datahandler.thrift.licenses.ObligationNode;
 import org.eclipse.sw360.datahandler.thrift.licenses.ObligationSortColumn;
 import org.eclipse.sw360.datahandler.thrift.users.User;
+import org.eclipse.sw360.datahandler.thrift.users.UserGroup;
+import org.eclipse.sw360.datahandler.permissions.PermissionUtils;
 import org.eclipse.sw360.rest.resourceserver.core.BadRequestClientException;
+import org.springframework.security.access.AccessDeniedException;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.Pageable;
@@ -87,12 +90,17 @@ public class Sw360ObligationService {
     }
 
     public Obligation updateObligation(Obligation obligation, User sw360User) {
+        if (!PermissionUtils.isUserAtLeast(UserGroup.CLEARING_ADMIN, sw360User)) {
+            throw new AccessDeniedException("User should be at least " + UserGroup.CLEARING_ADMIN);
+        }
         if (CommonUtils.isNotNullEmptyOrWhitespace(obligation.getTitle())
                 || CommonUtils.isNotNullEmptyOrWhitespace(obligation.getText())) {
             try {
                 LicenseService.Iface sw360LicenseClient = getThriftLicenseClient();
-                String updatedNode = sw360LicenseClient.addNodes(obligation.getNode(), sw360User);
-                obligation.setNode(updatedNode);
+                if (obligation.isSetNode() && CommonUtils.isNotNullEmptyOrWhitespace(obligation.getNode())) {
+                    String updatedNode = sw360LicenseClient.addNodes(obligation.getNode(), sw360User);
+                    obligation.setNode(updatedNode);
+                }
                 sw360LicenseClient.updateObligation(obligation, sw360User);
                 return obligation;
             } catch (TException e) {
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@GMishx This fixes a regression where updating an obligation could silently wipe its node data if addNodes() fails, even though the API still reports success.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/obligation/Sw360ObligationService.java
GMishx, line 112: Instead of making it so complicated, the function can be simplified by adding a permission check at the top like:
```java
    public Obligation updateObligation(Obligation obligation, User sw360User) {
        if (!PermissionUtils.isUserAtLeast(UserGroup.CLEARING_ADMIN, sw360User)) {
            throw new AccessDeniedException("User should be at least " + UserGroup.CLEARING_ADMIN);
        }
        if (CommonUtils.isNotNullEmptyOrWhitespace(obligation.getTitle())
                || CommonUtils.isNotNullEmptyOrWhitespace(obligation.getText())) {
            try {
                LicenseService.Iface sw360LicenseClient = getThriftLicenseClient();
                String updatedNode = sw360LicenseClient.addNodes(obligation.getNode(), sw360User);
                obligation.setNode(updatedNode);
                sw360LicenseClient.updateObligation(obligation, sw360User);
                return obligation;
            } catch (TException e) {
                throw new RuntimeException("Error updating obligation", e);
            }
        } else {
            throw new BadRequestClientException("Obligation Title, Text are required. Obligation Title, Text cannot contain only space character.");
        }
    }
```

In addition, I'd change https://github.com/eclipse-sw360/sw360/blob/8edaedc1085ccdd15ffe7cfdfbf87b70402537ad/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java#L317-L319 as:

```java
if (!PermissionUtils.isUserAtLeast(UserGroup.CLEARING_ADMIN, user)){
    throw new SW360Exception("User should be at least " + UserGroup.CLEARING_ADMIN);
}
```

GMishx: You can keep the `null` check for `updatedNode` as it can also fail for JSON parsing failure.

Aman-Cool: Applied the suggested fail-fast permission check and replaced silent `null` returns with explicit exceptions. The `null` guard for node parsing failures is kept to ensure atomic updates.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/obligation/Sw360ObligationService.java
```diff
@@ -87,11 +90,18 @@ private LicenseService.Iface getThriftLicenseClient() throws TTransportException
...
     public Obligation updateObligation(Obligation obligation, User sw360User) {
+        if (!PermissionUtils.isUserAtLeast(UserGroup.CLEARING_ADMIN, sw360User)) {
+            throw new AccessDeniedException("User should be at least " + UserGroup.CLEARING_ADMIN);
+        }
         if (CommonUtils.isNotNullEmptyOrWhitespace(obligation.getTitle())
                 || CommonUtils.isNotNullEmptyOrWhitespace(obligation.getText())) {
             try {
                 LicenseService.Iface sw360LicenseClient = getThriftLicenseClient();
                 String updatedNode = sw360LicenseClient.addNodes(obligation.getNode(), sw360User);
+                if (updatedNode == null) {
```
GMishx: The null check is no longer required as the function no longer returns null, but throws exception. Can you please check all possible paths and remove the null check?

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/obligation/Sw360ObligationService.java
Aman-Cool: Removed the redundant null check.  
`addNodes()` now throws, so the code relies on a single, consistent failure path.

GMishx:
@Aman-Cool can you squash your commits so we can keep the commit history consistent and easy to bisect?

Aman-Cool:
@GMishx , Squashed the commits

keerthi-bl:
Testing this PR.

keerthi-bl:
Testing has been completed successfully with no issues identified.

Aman-Cool:
@keerthi-bl ,Quick check-in, since tests are passing, is there anything else needed from my side to move this forward?

keerthi-bl:
> @keerthi-bl ,Quick check-in, since tests are passing, is there anything else needed from my side to move this forward?

Testing is completed, just needed a approval from reviewer. From your end no changes required.

PR #3657, opened 2026-02-03
fix(config): update VCS_HOSTS to use JSON array format for config consistency

### LINKED ISSUE #3653: fix(Config): Update Config is breaking
### Description

The update of configs under Admin tab isn't working. This is because the `vcs.hosts` doesn't belong to the update config DTO at the backend.

#### How to reproduce

Just change any value at the Admin > Configurations page and click on the update backend config button.

Image

Sandipmandal25:
### Summary                                                                                                                               
                                                                                                                                        
Move VCS_HOSTS from JSON array validation to string validation in isConfigValid()                                                   
                                                                                                                                        
 ### Problem                                                                                                                               
                                                                                                                                        
 Updating backend configs fails with "Invalid config: [vcs.hosts : ]" because VCS_HOSTS was incorrectly validated as a JSON array, but  it's actually a commaseparated string parsed by RepositoryURL.java.                                                                  
                                                                                                                                        
  ### Fix                                                                                                                                   
                                                                                                                                        
  Move VCS_HOSTS to the string validation case, allowing empty strings to pass validation.     

Suggest Reviewer
@amritkv                                          
                                                                                                                                        
Fixes #3653

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/common/utils/RepositoryURL.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/common/utils/RepositoryURL.java
@@ -25,14 +25,20 @@ import static org.eclipse.sw360.datahandler.common.SW360ConfigKeys.VCS_HOSTS;
 public class RepositoryURL {
     private static final Logger log = LogManager.getLogger(RepositoryURL.class);
     private static final String SCHEMA_PATTERN = ".+://(\\w*(?:[\\-@.\\\\s,_:/][/(.\\-)A-Za-z0-9]+)*)";
-    private static final String VCS_HOSTS_STRING = SW360Utils.readConfig(VCS_HOSTS,"");
-    private static final Map<String, String> KNOWN_VCS_HOSTS = parseVCSHosts(VCS_HOSTS_STRING);
+    private static String VCS_HOSTS_STRING = SW360Utils.readConfig(VCS_HOSTS,"[]");
+    private static Map<String, String> KNOWN_VCS_HOSTS = parseVCSHosts(VCS_HOSTS_STRING);
 
     public String processURL(String url) {
         return sanitizeVCS(url);
     }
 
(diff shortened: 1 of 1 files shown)


keerthi-bl:
Test successful.

Sandipmandal25:
hi @GMishx for RepositoryURL should I change it to read VCS_HOSTS dynamically from the configcontainer instead of the current static final? Or did you mean something else?

GMishx:
@Sandipmandal25 what meant was to modify `backend/common/src/main/java/org/eclipse/sw360/common/utils/RepositoryURL.java` to have the `VCS_HOSTS_STRING` as a `Set<String>` instead. And maybe move the `parseVCSHosts()` function so the config value passed by client can be sanitized and store in the DB.

Doing so, client can still send the CSV values like others and the backend will still understand and parse them. If we take the current changes, the `VCS_HOST` would then become a CSV which must be passed in a special format by the client and that is a bad design of an API. Everything should be intuitive as much as possible and that can only happen when you have consistency.

Sandipmandal25:
> what meant was to modify `backend/common/src/main/java/org/eclipse/sw360/common/utils/RepositoryURL.java` to have the `VCS_HOSTS_STRING` as a `Set<String>` instead. And maybe move the `parseVCSHosts()` function so the config value passed by client can be sanitized and store in the DB.
> 
> Doing so, client can still send the CSV values like others and the backend will still understand and parse them. If we take the current changes, the `VCS_HOST` would then become a CSV which must be passed in a special format by the client and that is a bad design of an API. Everything should be intuitive as much as possible and that can only happen when you have consistency.

 thanks for the clarification. I will  keep vcs_hosts under isvalidSet so the client sends it as a json array . and update `RepositoryURL` to parse the Set format instead of csv.

Sandipmandal25:
Hello @keerthi-bl can you please help to test this pr? Thank you!

PR #3682, opened 2026-02-11
feat(licenseInfo): Normalise license text and optimise performance

afsahsyeda:
**Description**

1. Optimized the performance by reducing overhead of creation of streams and replace with simple loops. 
2. Repetitive data structures/maps have been reduced.
3. License text is normalized to only alphanumeric characters to reduce duplicates as much as possible.
4. Unnecessary additional sorting is removed.
5. Fixes the bug where mismatched href tags were created for licenses by using a global LicenseNameWithText Obj to reference Id mapping, such that each LicenseNameWithText across all LicenseInfoParsingResults has a valid ref id assigned.

### DIFF
--- a/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
+++ b/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
@@ -181,6 +181,69 @@ public abstract class OutputGenerator<T> {
                 .orElse(defaultEmpty.get()).stream();
     }
 
+
+    protected Map<String, Set<String>> buildAcknowledgements(
+            SortedMap<String, LicenseInfoParsingResult> sortedLicenseInfos) {
+        Map<String, Set<String>> acknowledgements = new LinkedHashMap<>();
+
+        for (Map.Entry<String, LicenseInfoParsingResult> e : sortedLicenseInfos.entrySet()) {
+            LicenseInfoParsingResult pr = e.getValue();
+            if (pr == null) continue;
+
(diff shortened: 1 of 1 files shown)


akshitjoshii:
Test was successful.

backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
```diff
@@ -181,6 +181,69 @@ private <T extends Collection<U>, U> Stream<U> getCollectionStream(LicenseInfoPa
...
+
+            if ((c >= '0' && c <= '9') ||
+                    (c >= 'a' && c <= 'z')) {
+
+                sb.append(c);
+
+            } else if (c >= 'A' && c <= 'Z') {
+                sb.append((char) (c + 32));
+            }
+        }
```
GMishx, line 240: Why are we limiting to only lowercase English characters? License text can be in different language, contain special symbols like `§`, `“`.

afsahsyeda: @GMishx The idea is that two license texts are considered the same if they differ only by special symbols. For ex, Two licenses could have the same content, one with quotes and the other without, making the system identify them as different even though they are not, content-wise.

As far as a different language is considered, I will check if the license text is rendered empty after the loop, which would indicate that the text contains only non-English characters and then we can retain the original text.

GMishx: Oh, sorry I missed the point that this was function was used for `key` and not for output.

PR #3689, opened 2026-02-12
fix(CDXImporter): Do not remove valid URL string after .git

afsahsyeda:
**Description**

- The current importer sanitizes a VCS URL by removing any content after .git or #, assuming it represents query parameters or unnecessary information.
- This logic should be applied only to the final segment of the URL.


GMishx (approved):
LGTM

PR #3690, opened 2026-02-12
fix(component): file path traversal vulnerability.

rudra-superrr:
Description: This PR addresses a security vulnerability identified during penetration testing.

Problem: The SW360 web application allows users to download attachment bundles in ZIP format of attachments
stored in components. It was identified that this functionality is prone to a file path traversal vulnerability
when the application server is generating ZIP files with attachments that have ..\ in the filename.
The application server seems to copy the uploaded files to an application directory, which through
the path traversal vulnerability could allow attackers to write files to arbitrary folders in the server’s
filesystem.

Solution: Sanitize filenames at file system operations (file creation, ZIP generation) and not at database storage, preserving original user data while preventing path traversal attacks at the actual vulnerability points where filenames are used to create file paths or ZIP entries.

Testing:
Create a Test Component
Upload Test Files as attachment with Malicious fileNames(ex. ../test.txt or ../../tmp/test.txt)
Now download the attachment bundle from the component
Expected results: ../test.txt will be replaced by .._test.txt

### DIFF
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
@@ -40,6 +40,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.*;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
@@ -52,8 +53,10 @@ import static org.apache.logging.log4j.LogManager.getLogger;
 public class CommonUtils {
 
     public static final String SYSTEM_CONFIGURATION_PATH = "/etc/sw360";
-    public static final String DEFAULT_ATTACHMENT_FILENAME = "comp_attachment";
     private static List<String> MULTIPLE_FILE_EXTENSIONS = Arrays.asList(".tar.gz", ".tar.bz2", ".tar.xz", ".tar.lz", ".tar.lzma");
+    public static final String DEFAULT_ATTACHMENT_FILENAME = "comp_attachment";
+    private static final Pattern FILENAME_SANITIZE_PATTERN =
+            Pattern.compile("(?i)/|\\\\|%2f|%5c");
 
     private CommonUtils() {
         // Utility class with only static functions
@@ -245,7 +248,7 @@ public class CommonUtils {
     /**
      * Sanitize filename to prevent path traversal attacks.
      * Replaces path separators (/ and \) with underscores to prevent directory traversal.
-     * 
+     *
      * @param filename the original filename that may contain path separators
      * @return sanitized filename safe for file operations, or DEFAULT_ATTACHMENT_FILENAME if input is invalid
      */
@@ -253,16 +256,16 @@ public class CommonUtils {
         if (filename == null || filename.trim().isEmpty()) {
             return DEFAULT_ATTACHMENT_FILENAME;
         }
-        
+
         // Replace path separators with underscores
         // This prevents path traversal since sequences like "../" become ".._"
-        String sanitized = filename.replace("/", "_").replace("\\", "_");
-        
+        String sanitized = FILENAME_SANITIZE_PATTERN.matcher(filename).replaceAll("_");
+
         // If filename becomes empty after sanitization, use default
         if (sanitized.trim().isEmpty()) {
             return DEFAULT_ATTACHMENT_FILENAME;
         }
-        
+
         return sanitized;
     }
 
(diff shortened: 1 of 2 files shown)


libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
```diff
@@ -241,6 +241,41 @@ public static boolean isNotEmpty(Collection collection) {
...
+        }
+        
+        String sanitized = filename;
+        
+        // Remove forward slashes and backslashes (path separators)
+        sanitized = sanitized.replace("/", "_");
+        sanitized = sanitized.replace("\\", "_");
+        
+        // Remove parent directory references
+        sanitized = sanitized.replace("..", "_");
```
GMishx: Let's not do this as there can be some reasons to have `..` and on their own, they do no harm.

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
```diff
@@ -241,6 +241,41 @@ public static boolean isNotEmpty(Collection collection) {
...
+        
+        // Remove leading dots to prevent hidden files
+        while (sanitized.startsWith(".")) {
+            sanitized = "_" + sanitized.substring(1);
+        }
+        
+        // If filename becomes empty after sanitization, use default
+        if (sanitized.trim().isEmpty()) {
+            return "attachment";
```
GMishx: If it is a default/constant, let's make it a static constant in the class and use in this return and the first return...

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java
GMishx, line 1: Please sanitize the filename in `public Attachment uploadAttachment(MultipartFile file, Attachment newAttachment, User sw360User)` function as well.

GMishx (changes requested):

GMishx:
Can you please add a test case to the attachment controller test for this specific case?

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java
rudra-superrr: done

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
rudra-superrr: done

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
rudra-superrr: done

GMishx:
@bibhuti230185 please help test the PR

bibhuti230185:
@GMishx  and @rudra-superrr . One very minor edge case with encoded URL separator fails but we can take the call if this is not a most relevant scenario.
## Verification Summary
- **Original issue**: Filenames like `../test.txt`, `../../test.txt`, `../../../tmp/test4.txt` allowed traversal during ZIP creation and file handling.
- **Applied fix**: Filenames are sanitized at file/ZIP creation points by replacing `/` and `\` with `_`.
- **Expected result**: Traversal strings become safe filenames (e.g., `../../tmp/test.txt` → `.._.._tmp_test.txt`), preventing writes outside the attachment/temp directories.
- **Known limitation (URL-encoded separators)**: `%2F` / `%5C` are not altered unless decoded before sanitization (e.g., `..%2F..%2Fetc%2Fpasswd`).
- **Minor**: `renameFile(...)` uses `lastIndexOf("/")`; Windows `\` paths can yield `-1` (prefer `getParent()` or `File.separator`).
- **Info**: redundant sanitization in `downloadAttachmentBundleWithContext` (safe, optional cleanup).
- **Manual checks performed**: traversal uploads (`../`, `../../`, `..\\`, mixed), absolute paths (`/etc/passwd`, `C:\\Windows\\System32\\config`) and bundle download verified sanitized ZIP entries.

### Source files checked (fix points)
- `libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java` (sanitizeFilename, lines 245-267)
- `libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/couchdb/AttachmentStreamConnector.java` (ZIP entry creation, lines 118-137)
- `rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java` (ZIP entry creation, lines 212-223; upload sanitize, lines 231-241; renameFile, lines 421-430)

GMishx (approved):

PR #3693, opened 2026-02-12
fix: update Thrift version to 0.22.0 for compatibility

KuruHe:
🐛 Issues Found and Fixed
1. Thrift Version Compatibility Issue
Problem: Project hardcoded to Thrift 0.20.0, but developers often have newer versions (0.22.0) installed
Impact: Generated Java code had incompatible type arguments for ProcessFunction and AsyncProcessFunction
Error Messages:
"wrong number of type arguments; required 2"
"wrong number of type arguments; required 3"
Root Cause: Thrift 0.22.0 generates different API signatures than 0.20.0
2. Java Version Compatibility Issue
Problem: Project requires Java 21, but systems often have newer Java versions (25+)
Impact: Maven enforcer plugin fails build with version range violations
Error: "Detected JDK ... is version 25.0.1 which is not in the allowed range [21,23)"
3. Build Dependency Resolution Issues
Problem: Incorrect module build order causing missing dependencies
Impact: Build failures due to unresolved artifacts
Affected Modules: backend, libraries, rest modules

Commit Changes: 

fix: update Thrift version to 0.22.0 for compatibility

- Update thrift.version from 0.20.0 to 0.22.0 in pom.xml
- Add fix-thrift-generation.sh script to handle Thrift 0.22.0 compatibility
- Fixes compilation errors with newer Thrift versions
- Resolves ProcessFunction type argument incompatibilities

Fixes build issues for developers with Thrift 0.22.0 installed

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
@@ -531,6 +531,7 @@ public class LicenseController implements RepresentationModelProcessor<Repositor
             description = "Import OSADL information.",
             tags = {"Licenses"}
     )
+    @PreAuthorize("hasAuthority('WRITE')")
     @RequestMapping(value = LICENSES_URL + "/import/OSADL", method = RequestMethod.POST)
     public ResponseEntity<RequestSummary> importOsadlInfo() throws TException {
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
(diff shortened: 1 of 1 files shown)


GMishx:
The changes done in this PR needs prior discussion as they are major changes and cannot be immediately like this. This will result in broken systems.
* Thrift update: Versions other than 0.20.0 are not supported and the same is documented in the installation guides.
* Java version: You mentioned about it in the description but nothing in the changes??? Also, version 25 should support 23.
* Dependency resolution: It was just mentioned in the description, no actual cause, reason or even description of what is not working.

Please do not use Generative AI to simply create PR/issues in projects and code you yourself do not understand. It will only diminish your credibility.

PR #3708, opened 2026-02-15
fix: Prevent NullPointerException in ModerationRequestController

### LINKED ISSUE #3705: [bug] Potential NullPointerException in ModerationRequestController.getModerationRequestsByState
In `ModerationRequestController.java`, the `getModerationRequestsByState` method has a potential NullPointerException vulnerability when the `state` parameter is null.
### Location
- **File:** `rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java`
- **Line:** 203
- **Method:** `getModerationRequestsByState`
### Issue
The method calls `stateOptions.get(0).equalsIgnoreCase(state)` without checking if `state` is null. If the state parameter is not provided, it will throw NullPointerException.
### Current Code (buggy):
```java
boolean stateOpen = stateOptions.get(0).equalsIgnoreCase(state);


Recommended Fix:

boolean stateOpen = "open".equalsIgnoreCase(state);

This is a standard Java idiom to avoid NullPointerException - by putting the constant string first, if state is null, it simply returns false instead of throwing an NPE.

How to reproduce
1. Call the moderation request API without providing the state parameter
2. The application will throw NullPointerException

### PR DESCRIPTION
**Pull request by ADITYA-CODE-SOURCE**
## Description

Fixed potential NullPointerException in ModerationRequestController.getModerationRequestsByState method.

## Changes

Changed from:

To:

## Why This Fix Works

This is a standard Java idiom to avoid NullPointerException - by putting the constant string first ("open".equalsIgnoreCase(state)), if state is null, it simply returns false instead of throwing an NPE.

## Testing

- No functional changes, only null-safety improvement
- The logic remains the same since stateOptions.get(0) is always "open"

Fixes #3705

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java
@@ -200,7 +200,7 @@ public class ModerationRequestController implements RepresentationModelProcessor
                     "Invalid ModerationRequest state '%s', possible values are: %s", state, stateOptions));
         }
 
-        boolean stateOpen = stateOptions.get(0).equalsIgnoreCase(state);
+        boolean stateOpen = "open".equalsIgnoreCase(state);
         Map<PaginationData, List<ModerationRequest>> modRequestsWithPageData =
                 sw360ModerationRequestService.getRequestsByState(sw360User, pageable, stateOpen, allDetails);
         return getModerationResponseEntity(pageable, request, allDetails, modRequestsWithPageData);

### CODE REVIEW

**Comment by GMishx** [comment]
@ADITYA-CODE-SOURCE please fix the PR description!

File: rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/moderationrequest/ModerationRequestController.java [hunk 1]
```diff
@@ -200,7 +200,7 @@ public ResponseEntity<CollectionModel<ModerationRequest>> getModerationRequestsB
                     "Invalid ModerationRequest state '%s', possible values are: %s", state, stateOptions));
         }
 
-        boolean stateOpen = stateOptions.get(0).equalsIgnoreCase(state);
+        boolean stateOpen = "open".equalsIgnoreCase(state);
```
GMishx, line 203: `stateOptions` can never be `null`, `stateOptions.get(0)` can never be `null` as they are set in line `193-194`

PR #3709, opened 2026-02-15
fix: Prevent NullPointerException in Sw360ProjectService

### LINKED ISSUE #3706: bug Potential NullPointerException in Sw360ProjectService.setObligationsFromAdminSection
In Sw360ProjectService.java, the setObligationsFromAdminSection method has a potential NullPointerException vulnerability when the oblLevel parameter is null.
Location
- File: rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
- Lines: 738, 742, 746
Issue
The method calls oblLevel.equalsIgnoreCase(...) without checking if oblLevel is null. If oblLevel is null, it will throw NullPointerException.

 **Current Code (buggy):**

if (oblLevel.equalsIgnoreCase("Project")) {
    // ...
} else if (oblLevel.equalsIgnoreCase("Organization")) {
    // ...
} else if (oblLevel.equalsIgnoreCase("Component")) {
    // ...
}
         **Recommended Fix:**

if ("Project".equalsIgnoreCase(oblLevel)) {
    // ...
} else if ("Organization".equalsIgnoreCase(oblLevel)) {
    // ...
} else if ("Component".equalsIgnoreCase(oblLevel)) {
    // ...
}

This is a standard Java idiom to avoid NullPointerException - by putting the constant string first, if oblLevel is null, it simply returns false instead of throwing an NPE.

ADITYA-CODE-SOURCE:
## Description

Fixed potential NullPointerException in Sw360ProjectService.setObligationsFromAdminSection method.

## Changes

Changed from:

To:

## Why This Fix Works

This is a standard Java idiom to avoid NullPointerException - by putting the constant string first ("Project".equalsIgnoreCase(oblLevel)), if oblLevel is null, it simply returns false instead of throwing an NPE.

## Testing

- No functional changes, only null-safety improvement
- The logic remains the same

Fixes #3706

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
@@ -735,15 +735,15 @@ public class Sw360ProjectService implements AwareOfRestServices<Project> {
         ThriftClients thriftClients = new ThriftClients();
         ComponentService.Iface componentClient = thriftClients.makeComponentClient();
 
-        if (oblLevel.equalsIgnoreCase("Project")) {
+        if ("Project".equalsIgnoreCase(oblLevel)) {
             updatedObligationStatusMap = SW360Utils.getProjectComponentOrganisationLicenseObligationToDisplay(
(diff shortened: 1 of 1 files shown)


GMishx:
@ADITYA-CODE-SOURCE your future PRs will not be reviewed unless they follow the communities template.

PR #3718, opened 2026-02-16
fix: report errors instead of unconditional SUCCESS in CycloneDX SBOM…

Aman-Cool:
## Description

Fixes CycloneDX BOM import always returning SUCCESS despite component creation failures, project update failures, and silently dropping failed components. The import reports success even when components fail to create, causing compliance violations and invisible data loss.

The `importSbomAsProject()` method unconditionally sets `RequestStatus.SUCCESS` regardless of actual import outcomes. Component creation failures are caught and logged but never tracked or surfaced to API consumers.

## Related Issue

N/A

## Changes Made

* Added `failedComponentCount` tracking in `importAllComponentsAsPackages()` and `importAllComponentsAsReleases()`
* Increment counter on each `SW360Exception` during component creation
* Add `COMP_CREATION_FAILED_COUNT` to response `messageMap` when failures occur
* Return `RequestStatus.FAILURE` instead of `SUCCESS` when project update fails
* Return `RequestStatus.FAILURE` when any components fail to create
* Add explicit error handling for component update failures

**File changed:** `backend/common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java` (~40 lines)

## Testing

* ✅ Manual test performed
* ✅ Verified failure counts now returned in response messageMap
* ✅ Tested with CycloneDX BOMs containing invalid components
* ⭕ Unit test (existing tests cover this code path)

## Impact

* Component creation failures now properly reported to users
* Project update failures surfaced immediately instead of silent success
* Users can detect incomplete imports for compliance workflows (EU CRA, US EO 14028)
* No behavior change outside failed import scenarios
* No breaking changes to existing functionality

## Checklist

* ✅ Code follows project style guidelines
* ✅ All existing tests pass
* ✅ Signed-off-by added to commit (`git commit -s`)
* ✅ Commit message follows conventional format: `fix(cyclonedx): return FAILURE status for import errors`
* ✅ No new files added (existing file modified)

## Additional Notes

This is a critical fix for compliance scenarios where SBOM imports appear successful but silently lose components. The method previously caught exceptions and continued processing but always returned SUCCESS, creating a blind spot for regulatory compliance workflows.


Aman-Cool:
Hi @GMishx ,
Found this while using CycloneDX imports in production; the importer returns SUCCESS even when components fail to create, causing silent data loss in SBOM workflows.
Happy to address any feedback. Thanks!

GMishx:
@Farooq-Fateh-Aftab please test if the new key causes any issue with the new UI.

Farooq-Fateh-Aftab:
@Aman-Cool ,
Tested.
Requested few details, On which condition this key will be incremented .
image
as in the image the added key  still counts zero.

Aman-Cool:
@Farooq-Fateh-Aftab ,thanks for testing!
The counter was only incrementing inside the `SW360Exception` catch block, so packages rejected earlier by validation (null purl, missing name/version etc.) were just hitting continue and silently skipping, never touching the counter. That's why it showed 0 even with 9 invalid packages in the UI.
Added `compImportErrorCount++` at each of those validation-rejection points in both methods. Should now count both cases. Pushed the fix, can you re-test with the same SBOM?

GMishx:
@Farooq-Fateh-Aftab can you please provide comments on the testing so we can merge this PR?

Farooq-Fateh-Aftab:
Tested. count is now incremented.

image

GMishx (approved):
Good to go

PR #3722, opened 2026-02-16
fix(licenseinfo): check Optional.isPresent() before get() in DocxGenerator

mattpapp:
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

> Please provide a summary of your changes here.
> * Which issue is this pull request belonging to and how is it solving it? (*Refer to issue here*)
> * Did you add or update any new dependencies that are required for your change?

Issue: `Optional.get()` was called before `isPresent()` in both `DISCLOSURE` and 
`REPORT` cases in `DocxGenerator.generateOutputFile()`. If a template file 
is missing this throws a `NoSuchElementException` instead of the intended 
`SW360Exception`. To fix it I moved the `isPresent()` check before `.get()` in both cases.

Closes: #3721

Remove or rename a DOCX template resource file and trigger license info 
DOCX generation. Before this fix a `NoSuchElementException` is thrown. 
After the fix the intended `SW360Exception` with a message 
is thrown instead. The existing `DocxGeneratorTest` passes.
> Have you implemented any additional tests?
- N/A

### Checklist
Must:
- [x] All related issues are referenced in commit messages and in PR

PR #3729, opened 2026-02-18
fix: check Optional.isPresent() before calling get() in DocxGenerator (#3721)

### LINKED ISSUE #3721: Bug: DocxGenerator crashes with NoSuchElementException when template file is missing
Found a bug in `DocxGenerator.generateOutputFile()`. In both the `DISCLOSURE` 
and `REPORT` switch cases, `.get()` is called on the `Optional` before checking 
`.isPresent()`. 

So if a template file is missing or fails to load, you get a raw 
`NoSuchElementException` instead of the `SW360Exception` with the helpful 
"Could not load the template" message that was clearly intended.

The fix is straightforward, just flip the order so `isPresent()` is checked 
first. Happy to open a PR for this.

ADITYA-CODE-SOURCE:
## Summary

Fixes #3721 - DocxGenerator crashes with NoSuchElementException when template file is missing

### Bug
In DocxGenerator.generateOutputFile(), the code was calling .get() on the Optional before checking .isPresent(). This caused a raw NoSuchElementException.

### Fix
Reordered the code to check isPresent() first:
- Line 110: Moved inside the if block
- Line 132: Same fix for REPORT case

### DIFF
--- a/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/tools/LicenseDBProperties.java
+++ b/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/tools/LicenseDBProperties.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright Siemens AG, 2025. Part of the SW360 Portal Project.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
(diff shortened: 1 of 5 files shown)


ADITYA-CODE-SOURCE:
Hello @GMishx @KoukiHama @ag4ums @arunazhakesan

I've fixed issue #3721 - DocxGenerator crashes with NoSuchElementException when template file is missing.
The fix reorders the code to check isPresent() before calling get() on the Optional.
CI is passing. Could you please review 
Thank you!

GMishx:
Thank you @ADITYA-CODE-SOURCE for contribution but the fix is already merged with #3722

PR #3746, opened 2026-02-23
fix(importCDX):do not alter unknown domain VCS and revert old changes

afsahsyeda:
**Description**

- Revert to the older code where everything after .git or # is removed in the VCS of identified domains.
- For VCS of unidentified domains, do not alter the vcs by removing the URL after .git or #.


akshitjoshii:
Test successful.

PR #3777, opened 2026-02-28
fix(rest): return 400 for missing required request parameters instead of 500

### LINKED ISSUE #3776: GET /api/attachments returns 500 Internal Server Error instead of 400 Bad Request when sha1 parameter is missing
**Description:**

`GET /api/attachments` requires a mandatory `sha1` query parameter. When the parameter is omitted, Spring throws `MissingServletRequestParameterException` which should map to `400 Bad Request`. Instead, the endpoint returns `500 Internal Server Error` due to a gap in `RestExceptionHandler`.

**Affected file:**
`rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestExceptionHandler.java`

**Screenshot**

Image

**How to reproduce:**

1. Authenticate to the SW360 REST API with any valid account.
2. Send `GET /api/attachments` with no query parameters.
3. Observe the response is `500 Internal Server Error` with message `Required request parameter 'sha1' for method parameter type String is not present`.

Expected: `400 Bad Request` : the client made an invalid request by omitting a required parameter.
Actual: `500 Internal Server Error` : misleads the caller into thinking the server crashed.

**Root cause:**

`RestExceptionHandler` has a catch-all handler for `Exception.class` that returns `500`:

```java
@ExceptionHandler({Exception.class, TException.class, ResourceClassNotFoundException.class})
public ResponseEntity<ErrorMessage> handleException(Exception e) {
    return new ResponseEntity<>(new ErrorMessage(e, HttpStatus.INTERNAL_SERVER_ERROR), ...);
}
```

`MissingServletRequestParameterException` extends `ServletRequestBindingException` → `BindException` → `Exception`. It is **not** a `RuntimeException`, so it bypasses the existing `RuntimeException → 400` handler and falls into the catch-all `Exception → 500` handler.

The sibling `RuntimeException` handler correctly returns `400` but `MissingServletRequestParameterException` never reaches it.

Shivamrut:
Fixes #3776

## Summary

- `MissingServletRequestParameterException` extends `ServletRequestBindingException` → `BindException` → `Exception` (not `RuntimeException`), causing it to bypass the `RuntimeException → 400` handler and fall into the catch-all `Exception → 500` handler in `RestExceptionHandler`
- Similarly, `MissingServletRequestPartException` (thrown for missing multipart file parameters) follows the same inheritance chain and was also incorrectly returning `500`
- Added a dedicated handler for both `MissingServletRequestParameterException` and `MissingServletRequestPartException` that correctly returns `400 Bad Request`
- Imports added accordingly

## Screenshot
Before:
image
After:
image

## Test plan

- [ ] Send `GET /api/attachments` with no `sha1` query parameter
- [ ] Verify response is now `400 Bad Request` with message `Required request parameter 'sha1' for method parameter type String is not present`
- [ ] Verify `GET /api/attachments?sha1=<valid_hash>` still works correctly and returns `200`
- [ ] Verify other existing exception handler behaviours are unaffected

### Query Parameter Endpoints (`MissingServletRequestParameterException`)
| Endpoint | Missing Parameter |
|----------|-------------------|
| `GET /api/attachments` | `sha1` |
| `POST /api/moderationrequest/validate` | `entityType` or `entityId` |
| `POST /api/departments/writePathFolder` | `pathFolder` |
| `GET /api/reports` | `module` |
| `GET /api/reports/download` | `module` or `token` |
| `PATCH /api/releases/mergereleases` | `mergeTargetId` or `mergeSourceId` |
| `PATCH /api/vendors/mergeVendors` | `mergeTargetId` or `mergeSourceId` |
| `GET /api/schedule/status` | `serviceName` |

### Multipart File Endpoints (`MissingServletRequestPartException`)
| Endpoint | Missing Parameter |
|----------|-------------------|
| `POST /api/attachments` | `files` |
| `POST /api/importExport/uploadComponent` | `file` |
| `POST /api/importExport/uploadRelease` | `file` |
| `POST /api/importExport/componentAttachment` | `file` |

All these endpoints previously returned `500 Internal Server Error` instead of `400 Bad Request` for missing parameters.


GMishx:
@Shivamrut please squash all 6 commits into one. This PR works on a single fix and thus, there should be only 1 commit.

Shivamrut:
@GMishx Thanks for review. I have squashed all commits. Anymore changes required please let me know.

Issue #3778, opened 2026-02-28
Department import permanently disabled after unchecked exception in UserDatabaseHandler


Priyasharma620064:
### Description

The IMPORT_DEPARTMENT_STATUS flag in UserDatabaseHandler.importFileToDB() can get stuck at true.
When this happens, the department import feature stops working and only starts working again after a server restart.
Right now, the flag is reset only when the import finishes successfully or when an IOException occurs. If any other exception happens (for example, a NullPointerException caused by a malformed Excel row like row.getCell(1).getStringCellValue() returning null), the flag is never reset.

As a result:
-The flag stays true
-Every future import request immediately returns RequestStatus.PROCESSING
-The feature remains blocked until the server is restarted

In addition, the flag is a simple static boolean and is accessed by two different threads:
-The scheduler thread (importDepartmentSchedule())
-The REST-triggered import thread (importDepartmentManually())

Since there is no synchronization, two threads could enter the import at the same time, causing a race condition.

#### How to reproduce

Configure a department import folder path via POST /departments/writePathFolder
Place a malformed Excel file in the folder (e.g., a row with a blank second column)
Trigger import via POST /departments/manuallyactive
The import throws NullPointerException at row.getCell(1).getStringCellValue()
Try triggering another import via POST /departments/manuallyactive

Result: Returns RequestStatus.PROCESSING indefinitely — feature is permanently locked
Expected: Import should recover and allow retries.

### SW360 logs

When the bug triggers, the following pattern appears:
-First import: NullPointerException stack trace in server logs
-All subsequent imports: No log output (method returns early before any logging)

PR #3779, opened 2026-02-28
fix(user): prevent permanent import lockout on unchecked exceptions

Priyasharma620064:
### Summary

Fixes the IMPORT_DEPARTMENT_STATUS flag in UserDatabaseHandler.importFileToDB() which can get permanently stuck at true after an unchecked exception, disabling department import until server restart.
-This PR fixes issue #3778 by moving the flag reset to a finally block and replacing the plain static boolean with AtomicBoolean.compareAndSet() for thread-safe guarding.
-No new dependencies added. AtomicBoolean is part of java.util.concurrent.atomic (standard JDK).

Fixes: #3778 

### Suggest Reviewer
@GMishx

### How To Test?
1. Verify the flag resets correctly on success — trigger a department import with valid files and confirm it completes and can be re-triggered
2. Verify the flag resets on failure — trigger an import with a malformed Excel file (blank second column) and confirm the next import attempt works instead of returning PROCESSING forever
3. Verify concurrent guard — trigger a manual import while a scheduled import is running and confirm the second call returns PROCESSING (not both running simultaneously)

### DIFF
--- a/backend/users/src/main/java/org/eclipse/sw360/users/db/UserDatabaseHandler.java
+++ b/backend/users/src/main/java/org/eclipse/sw360/users/db/UserDatabaseHandler.java
@@ -35,6 +35,7 @@ import java.io.*;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import static org.eclipse.sw360.datahandler.permissions.PermissionUtils.makePermission;
@@ -59,7 +60,7 @@ public class UserDatabaseHandler {
     private static final String SUCCESS = "SUCCESS";
     private static final String FAIL = "FAIL";
     private static final String TITLE = "IMPORT";
-    private static boolean IMPORT_DEPARTMENT_STATUS = false;
+    private static final AtomicBoolean IMPORT_DEPARTMENT_STATUS = new AtomicBoolean(false);
     private List<String> departmentDuplicate;
     private List<String> emailDoNotExist;
     DateFormat dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm:ss zzz yyyy", Locale.ENGLISH);
@@ -169,10 +170,9 @@ public class UserDatabaseHandler {
         DepartmentConfigDTO configDTO = readFileDepartmentConfig.readFileJson();
         String pathFolderLog = configDTO.getPathFolderLog();
         Map<String, List<String>> mapArrayList = new HashMap<>();
-        if (IMPORT_DEPARTMENT_STATUS) {
+        if (!IMPORT_DEPARTMENT_STATUS.compareAndSet(false, true)) {
             return requestSummary.setRequestStatus(RequestStatus.PROCESSING);
         }
-        IMPORT_DEPARTMENT_STATUS = true;
         Calendar calendar = Calendar.getInstance(Locale.getDefault());
         String lastRunningTime = dateFormat.format(calendar.getTime());
         readFileDepartmentConfig.writeLastRunningTimeConfig(lastRunningTime);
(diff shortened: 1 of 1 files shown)


backend/users/src/main/java/org/eclipse/sw360/users/db/UserDatabaseHandler.java
```diff
@@ -207,16 +207,16 @@ public RequestSummary importFileToDB(String pathFolder)  {
...
             requestSummary.setRequestStatus(RequestStatus.SUCCESS);
         } catch (IOException e) {
-            IMPORT_DEPARTMENT_STATUS = false;
             String msg = "Failed to import department";
             requestSummary.setMessage(msg);
             requestSummary.setRequestStatus(RequestStatus.FAILURE);
             FileUtil.writeLogToFile(TITLE, "FILE ERROR: " + e.getMessage(), "", pathFolderLog);
+        } finally {
+            IMPORT_DEPARTMENT_STATUS.set(false);
         }
```
GMishx, line 227: `finally` block will always run irrespective if there was an exception or not. This will result in `IMPORT_DEPARTMENT_STATUS` to always be `false`.

Priya-Sharma25:
Thanks @GMishx! Updated — moved log writes inside finally so the flag resets only after all work is done. Also added catch (Exception e) to handle unchecked exceptions gracefully. Please re-review!

backend/users/src/main/java/org/eclipse/sw360/users/db/UserDatabaseHandler.java
```diff
@@ -169,10 +170,9 @@ public RequestSummary importFileToDB(String pathFolder)  {
         DepartmentConfigDTO configDTO = readFileDepartmentConfig.readFileJson();
         String pathFolderLog = configDTO.getPathFolderLog();
         Map<String, List<String>> mapArrayList = new HashMap<>();
-        if (IMPORT_DEPARTMENT_STATUS) {
+        if (!IMPORT_DEPARTMENT_STATUS.compareAndSet(false, true)) {
```
GMishx, line 173: Can you please explain this change? I am bit confused. What I understood: If the `IMPORT_DEPARTMENT_STATUS` is `false`, set it to `true`. If the `IMPORT_DEPARTMENT_STATUS` was not updated because of this operation, which means no import was working, then return status as processing.

If my understanding is correct, this operation will always return in processing status upon first start????

Priya-Sharma25:
Hey @GMishx, sorry for the late reply — I had my midsems going on so couldn't get to this earlier! 
Let me clarify how compareAndSet works here:

compareAndSet(false, true) does the following atomically:

If the current value is false → set it to true and return true (success)
If the current value is true → do nothing and return false (another import is already in progress)
The ! (NOT) operator inverts this:

First call (no import running): flag is false → compareAndSet(false, true) succeeds → returns true → !true = false → if block is skipped → import proceeds normally .
Second concurrent call: flag is already true → compareAndSet(false, true) fails → returns false → !false = true → if block is entered → returns PROCESSING .
So it will NOT return PROCESSING on first start. It only returns PROCESSING when another import is already running — which is the same behavior as before, but now thread-safe.
The key improvement is the finally block: even if an unchecked exception occurs (like NullPointerException), the flag is guaranteed to reset to false, preventing the permanent lockout

GMishx:
According to the official docs at https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicBoolean.html#compareAndSet-boolean-boolean-

> `compareAndSet(boolean expect, boolean update)`
>
> Atomically sets the value to the given updated value if the current value == the expected value.
>
> **Returns:** `true` if successful. `False` return indicates that the actual value was not equal to the expected value.

That means, `IMPORT_DEPARTMENT_STATUS.compareAndSet(false, true)` will return `true` if `IMPORT_DEPARTMENT_STATUS` was `false` and sets it to `true`. The inverter will make return value `false` and the `if` is skipped. Process continues.
It will return `false` if `IMPORT_DEPARTMENT_STATUS` was `true` and will keep it `true`. The inverter will make return value `true` and the `if` block executes. In process status is returned.

Could have been explained in simple English :-)

PR #3780, opened 2026-03-01
fix(vmcomponents): handle NPE and ClassCastException in getVulIdsPerComponentVmId

### LINKED ISSUE #2726: SVM notifications are not getting updated
### Description

The SVM notifications are not getting updated in SW360 from the SVM endpoint. There is a NullPointerException thrown due to JSONObject and this could be a result of this commit: cd53eed2c6a3bd36a8d6eb8a77cf9e56456361e4

#### How to reproduce
SVM Sync does not update notifications for vulnerabilities in the Vulnerabilities tab.

image

#### SW360 logs
image

CC: @FarooqAftab @heliocastro

Shivamrut:
## Summary

Fixes #2726 — SVM notifications are silently not updated after the `json-simple` library upgrade in commit `cd53eed`.

## Root Cause

The bug was introduced when `json-simple` was upgraded to version 4.x (`cd53eed`). Two breaking changes affected `getVulIdsPerComponentVmId` in `SVMSyncHandler.java`:

**1. NullPointerException** — when the SVM API returns a JSON object that is missing the `"id"` field, `json.get(VULNERABILITY_ID)` returns `null`. The old code called `.toString()` directly on that result:

```java
// OLD — crashes if "id" field is absent
JsonObject json = (JsonObject) id;
String vulId = json.get(SVMConstants.VULNERABILITY_ID).toString(); // NPE here
```

**2. ClassCastException** — `json-simple 4.x` now parses plain integers in a JSON array as `BigDecimal` (previously `Long`). The old code unconditionally cast every array element to `JsonObject`:

```java
// OLD — crashes if element is a plain integer like 19936
JsonObject json = (JsonObject) id; // ClassCastException here
```

Both exceptions were caught by the surrounding `catch (RuntimeException e)` block and silently swallowed, returning an empty set. This caused vulnerability notifications to never be updated — exactly the symptom reported in #2726.

## Fix

```java
for (Object id : ids) {
    String vulId;
    if (id instanceof JsonObject) {
        Object vulIdObj = ((JsonObject) id).get(SVMConstants.VULNERABILITY_ID);
        if (vulIdObj == null) {
            continue;
        }
        vulId = vulIdObj.toString();
    } else {
        vulId = id.toString();
    }
    if (!StringUtils.isEmpty(vulId)) {
        vulIds.add(vulId);
    }
}
```

- If the element is a `JsonObject`, extract the `"id"` field safely with a null check.
- If the element is a primitive (`BigDecimal`, `String`, etc.), call `id.toString()` directly.

## Why the Bug Was Not Reproduced Live

Reproducing this bug end-to-end requires a live SVM API endpoint (Siemens internal). Without access to that endpoint, the sync process fails at the HTTP request stage before ever reaching the buggy line. Instead, the bug was confirmed through:

1. **Static code analysis** — tracing the call chain from the scheduler through `VMProcessor` → `SVMSyncHandler.getVulnerabilitiesByComponentId` → `getVulIdsPerComponentVmId`, confirming the exact lines that throw.
2. **Root-cause unit tests** — two tests that directly prove the crash scenarios exist at the `json-simple` library level (no SVM access needed).

## Tests

A new standalone test class `SVMSyncHandlerVulIdsTest` was added. It requires **no CouchDB** and **no SVM endpoint** — WireMock fakes the HTTP responses and Mockito mocks the DB handler.

| Test | What it proves |
|---|---|
| `rootCause_JsonObjectMissingKey_..._ThrowsNPE` | `json.get("id")` returns `null` for a missing key; `.toString()` throws NPE. Directly replicates the old buggy line. |
| `rootCause_PlainInteger_..._ThrowsClassCastException` | `json-simple 4.x` parses integers as `BigDecimal`; casting to `JsonObject` throws `ClassCastException`. |
| `testGetVulIds_ObjectsWithIdField_ReturnsVulIds` | Normal path: objects with `"id"` field return 3 IDs correctly. |
| `testGetVulIds_ObjectWithMissingIdField_ReturnsEmptyGracefully` | Missing `"id"` field returns empty set without exception. |
| `testGetVulIds_PlainIntegerIds_ReturnsVulIds` | **Definitive proof**: old code returns empty set (ClassCastException caught); fixed code returns all 3 IDs. `assertEquals(3, result.size())` fails on old code, passes on fix. |

All 5 tests pass: `Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -53,7 +53,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+        uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
         with:
           egress-policy: audit
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -59,7 +59,7 @@ jobs:
         # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+        uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
         with:
           egress-policy: audit
 
@@ -68,7 +68,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
+        uses: github/codeql-action/init@0d579ffd059c29b07949a3cce3983f0780820c98 # v4.32.6
         with:
           languages: ${{ matrix.language }}
           build-mode: ${{ matrix.build-mode }}
@@ -96,6 +96,6 @@ jobs:
           exit 1
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
+        uses: github/codeql-action/analyze@0d579ffd059c29b07949a3cce3983f0780820c98 # v4.32.6
         with:
           category: "/language:${{matrix.language}}"
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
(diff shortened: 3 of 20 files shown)


GMishx:
@akshitjoshii please help test this PR.

akshitjoshii:
@Shivamrut, This issue appeared due to a dependency update in #2726. So I would suggest you to update the same and then push the changes for us to test. As in future we may need to update it.

Shivamrut:
Hi @akshitjoshii, If you mean that I should update the json-simple package in maven then, I checked and com.github.cliftonlabs:json-simple 4.0.1 (currently set in pom.xml) is already the latest available version on Maven Central. There is no newer release to upgrade to. If you meant something else can you please explain me.

GMishx:
Hey @Shivamrut , from discussion, I just confirmed with @Farooq-Fateh-Aftab the commit cd53eed2c6a3bd36a8d6eb8a77cf9e56456361e4 was breaking the SVM sync and it was reverted internally and we missed to update the same in GitHub. Thus, the comment from @akshitjoshii is no longer relevant as the version `4.0.1` has the issue.

Also, the package was updated from `org.json.simple.JSONObject` to `com.github.cliftonlabs.json_simple.JsonObject`.

With that clarified @akshitjoshii are we good with this PR?

GMishx:
@akshitjoshii please provide your comments so we can move ahead with the PR.

akshitjoshii:
Test successful as sync works fine on stage.

PR #3787, opened 2026-03-02
fix(cyclonedx): preserve component group in BOM import to prevent name collisions

### LINKED ISSUE #2298: Component's group in CycloneDX BOM is ignored when importing to SW360 causing non-unique names
The `group` property of a component in CycloneDX BOM is ignored when importing. Only the `name` property is taken into consideration. This causes issues for package manager, which require a group to uniquely identify a package.

### Description

Many package manager support grouping packages. The group is required to uniquely identify a package.

One example is Maven (Java):

```xml
<!-- https://mvnrepository.com/artifact/org.apache.commons/commons-lang3 -->
<dependency>
    <groupId>org.apache.commons</groupId>
    <artifactId>commons-lang3</artifactId>
    <version>3.14.0</version>
</dependency>
```

Another example are scopes in NPM packages (JavaScript):

```
@sentry/browser
```

In CycloneDX specification each component has an optional `group` property. This is how the two examples above are represented in CycloneDX BOM generated by [cdxgen](https://github.com/CycloneDX/cdxgen):

```json
{
  "group": "org.apache.commons",
  "name": "commons-lang3",
  "version": "3.13.0",
  "purl": "pkg:maven/org.apache.commons/commons-lang3@3.13.0?type=jar",
  "type": "framework",
  "bom-ref": "pkg:maven/org.apache.commons/commons-lang3@3.13.0?type=jar"
}
```

```json
{
  "group": "@sentry",
  "name": "browser",
  "version": "7.89.0",
  "purl": "pkg:npm/%40sentry/browser@7.89.0",
  "type": "library",
  "bom-ref": "pkg:npm/@sentry/browser@7.89.0"
}
```

<small>I removed some properties not relevant for the bug.</small>

When importing those to SW360, components with the names `commons-lang3` and `browser` are created in SW360. The group, which is very relevant to identify the package, is not imported at all.

Please see also here for the relevant code: https://github.com/eclipse-sw360/sw360/blob/325ac541fe04816775f66cb521d1d2d5db77cf1e/backend/src-common/src/main/java/org/eclipse/sw360/cyclonedx/CycloneDxBOMImporter.java#L790-L808

#### How to reproduce

1. Generate a CycloneDX BOM using [cdxgen]() for Maven or NPM project using a scoped package.
2. Import CycloneDX BOM into SW360

Expected:

* Name of created components contains group and name (e.g. `org.apache.commons/commons-lang3` and `@sentry/browser`

Actual:

* Name of created components does not contain the group (e.g. `commons-lang3` and `browser`)

### Versions

* Version: 18.0.0 | Branch: UNKNOWN (fd29ef6) | Build time: 2024-01-15T06:30:33Z

MrButtCode:

> This PR addresses issue #2298 where the CycloneDX importer silently discards the `group` field, leading to name collisions and incorrect component merging during deduplication. It ensures data integrity by prepending the `group` to the `name` (e.g., `group/name`) while guarding against double-prefixing.
> * **Which issue is this pull request belonging to and how is it solving it?** This PR addresses Issue #2298. It modifies `createComponent` in `CycloneDxBOMImporter.java` to safely concatenate the group and name. It also adds a `.startsWith()` safety guard to handle inconsistencies across different SBOM generators. Finally, it adds `CycloneDxBOMImporterTest.java` to prevent future regressions.
> * **Did you add or update any new dependencies that are required for your change?** No new dependencies were added.

Issue: Closes #2298

### Suggest Reviewer
@GMishx @amritkv

### Checklist


GMishx:
@Suhas2109 @akshitjoshii @afsahsyeda please help review and test the PR.

Suhas2109:
Hi @MrButtCode ,
Could you provide a sample SBOM where the current implementation leads to incorrect behavior (as described in #2298)?
Also, how was this tested in the actual import flow (beyond unit tests)?

Since this change modifies how component names are constructed, it introduces a behavioral change, so having a clear scenario and validation would help ensure correctness.

MrButtCode:
Hi @Suhas2109 , thanks for the review.
**Sample BOM reproducing the issue:**
<details>
<summary>Click to expand</summary>
json{
  "bomFormat": "CycloneDX",
  "specVersion": "1.4",
  "version": 1,
  "components": [
    {
      "type": "library",
      "group": "@sentry",
      "name": "browser",
      "version": "7.89.0",
      "purl": "pkg:npm/%40sentry/browser@7.89.0"
    },
    {
      "type": "library",
      "group": "@fullstory",
      "name": "browser",
      "version": "1.0.0",
      "purl": "pkg:npm/%40fullstory/browser@1.0.0"
    }
  ]
}
</details>
Before this fix, importing this BOM creates one component named `browser` with two conflicting releases, merging two unrelated libraries into the same SW360 entity. After the fix, two distinct components `@sentry/browser` and `@fullstory/browser` are created correctly. The three unit tests in `CycloneDxBOMImporterTest` cover this exact scenario, the double-prefix edge case, and the no-group regression case.

**On the behavioral change:** you're right that this is a breaking change for existing instances. Components previously imported as `browser` won't be automatically renamed; users would need to re-import affected SBOMs after upgrading to correct their inventory. I've added a Migration Considerations note to the PR description. Leaving the collision behavior in place would mean new imports continue producing corrupt inventory indefinitely, which is the worse long-term outcome.

Suhas2109:
Hi @MrButtCode ,
Thanks for the contribution.
However, as per the current implementation, the component name already incorporates the group (e.g., "group.name"), and the VCS field further supports unique identification. This indicates that the group information is not lost and existing mechanisms already address potential collisions.

Given this, the proposed change does not seem necessary, and the current behavior appears sufficient.

MrButtCode:
Hey @Suhas2109,

I thought the same thing initially, but looking at the current `main` branch, `createComponent()` in `CycloneDxBOMImporter.java` (around line 792) actually just does this:
`javacomponent.setName(CommonUtils.nullToEmptyString(componentFromBom.getName()).trim());`
`getGroup()` isn't called there. If I missed another place in the import flow where this concatenation happens, definitely point me to it!

Regarding the VCS field: that works great when a VCS URL exists. But a lot of transitive dependencies (like NPM packages generated via cdxgen) only carry a purl and completely lack VCS data. For those, the name is the only deduplication key, which triggers the collision.

The new unit tests in `CycloneDxBOMImporterTest` actually reproduce this exact `@sentry/browser` vs `@fullstory/browser` collision directly against `createComponent()`. Let me know your thoughts!

akshitjoshii:
We can close this PR since the expected behaviour is already there in the cdx importer and the group tag has the same significance as the group name in the VCS url which is already used while naming the component.

GMishx:
Closing this PR as suggested by @akshitjoshii . Thank you for your contributions @MrButtCode

PR #3788, opened 2026-03-03
fix(rest): delete clearing request only after project deletion succeeds

Aman-Cool:
## Description

Fixes permanent deletion of clearing requests when project deletion fails with IN_USE or SENT_TO_MODERATOR status. The original code deleted the clearing request before attempting to delete the project, causing data loss when project deletion was blocked. The clearing request was permanently destroyed while the project remained, leaving an orphaned reference and losing all compliance review history.

The fix inverts the operation order: delete project first, then delete clearing request only if project deletion succeeds. This ensures clearing requests are never destroyed unless the project deletion actually completes.

## Related Issue

N/A

## Changes Made

* Moved project deletion before clearing request deletion in `ProjectController.deleteProject()`
* Clearing request is now only deleted after successful project deletion
* Changed CR deletion failure handling from return 500 to log error (project already deleted)
* Prevents irreversible clearing request deletion when project deletion is blocked

**File changed:** `rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java` (~5 lines reordered)

## Testing

* ✅ Manual test performed with project linked as sub-project (IN_USE scenario)
* ✅ Verified clearing request preserved when project deletion returns IN_USE
* ✅ Verified clearing request deleted when project deletion succeeds
* ✅ Tested with SENT_TO_MODERATOR status - CR preserved correctly
* ⭕ Unit test (existing tests cover this code path)

## Impact

* **Data integrity:** Clearing requests no longer permanently deleted on failed project deletion
* **Compliance:** Review history and clearing notes preserved when deletion is blocked
* **Correctness:** Project documents no longer have stale clearing request references
* **API accuracy:** 409 CONFLICT response reflects actual failure without silent side effects
* **Audit trails:** Compliance records maintained for enterprise regulatory requirements
* No breaking changes to API behavior for successful deletions

## Checklist

* ✅ Code follows project style guidelines
* ✅ All existing tests pass
* ✅ Signed-off-by added to commit (`git commit -s`)
* ✅ Commit message follows conventional format: `fix(rest): prevent CR deletion when project deletion fails`
* ✅ No new files added (existing file modified)

## Additional Notes

This bug demonstrates a transaction boundary issue across Thrift service calls: the REST controller orchestrates two separate Thrift services (moderation and project) without transactional coordination. When the second operation fails, the first cannot be rolled back.

**Original sequence (broken):**
1. Delete CR via Thrift moderation service ← irreversible
2. Delete project via Thrift project service ← can fail with IN_USE

**Fixed sequence:**
1. Delete project first ← if fails, abort immediately
2. Delete CR only on SUCCESS ← safe to proceed

With the Thrift migration to direct Spring services, this would be a single Transactional method with automatic rollback.


Aman-Cool:
Hey @GMishx, The project deletion code deletes the clearing request first, then tries to delete the project. If the project is IN_USE, the CR gets destroyed but the project stays clearing history is permanently lost.

Fixed by swapping the order.

Happy to adjust anything.

GMishx:
@nikkuma7 can you please provide your comments on this PR? Would this be the desired behavior?

nikkuma7:
@Aman-Cool 
This is similar to your "Fixed sequence" but with a more explicit pre-check to ensure the CR isn't blocking the project deletion.

Check for CR Dependencies:

Before attempting any deletion, query the Thrift moderation service to see if the CR is still linked to the project.
If the CR is found, attempt to delete it first. If this fails, abort the entire process and report the CR deletion failure.
Delete Project:

If the CR deletion (or its absence) was successful, then proceed to delete the project via the Thrift project service. If this fails (e.g., IN_USE by something else), abort.

Aman-Cool:
Hey @nikkuma7, Thank you for the review🙏

I think the approach you're suggesting would actually reintroduce the original bug, if we delete the CR first and the project deletion then fails (e.g. IN_USE), the CR is already gone. The project is still alive but its clearing history is permanently wiped, which is exactly what this PR was trying to fix.

The intent here was to flip the order: attempt project deletion first, and only delete the CR if that succeeds. That way the CR is never touched unless we're sure the project is gone.

That said, you're right that there's still an edge case: if the project deletion succeeds but the CR deletion fails, we end up with an orphaned CR referencing a non-existent project. Right now the code logs the error and returns 200, which isn't ideal.

I'm open to either approach for handling that, a response body warning alongside the 200, or a background cleanup job for orphaned CRs. Not sure what fits best with how the rest of the codebase handles partial failures like this. Happy to go with whatever direction you prefer.

Aman-Cool:
Hey @nikkuma7, went ahead and pushed an update since it had been a while!
So after thinking about it more, I went with deleting the project first rather than the CR. My concern with the CR-first approach was that if deleteProject then comes back with IN_USE, we've already wiped the clearing history for a project that's still very much alive. That felt worse than the original bug honestly.

With the current approach; snapshot the CR state, delete the project, then clean up the CR, the only real failure mode left is: project gets deleted successfully but the CR removal fails. In that case the backend already orphans the CR anyway (via unlinkClearingRequestForProjectDeletion in removeProjectAndCleanUp), so it's not silently dangling. I also noticed the previous version of the code just logged an error and returned a plain 200 in that case, which felt off since the caller had no idea something went wrong. Changed that to return a 200 with a message body explaining that the CR couldn't be removed, so at least the caller knows and can act on it.

Also restructured the conditionals to use early returns, makes the happy path at the bottom easier to follow.

Aman-Cool:
Hey @GMishx,
The fix now deletes the project first, then cleans up the clearing request only if that succeeds. The previous version had the CR-first order which would've wiped clearing history on a project that's still alive if deletion failed with IN_USE. Also handled the edge case where the project deletes fine but CR cleanup fails, instead of silently returning a 200, the response now includes a message so the caller actually knows something needs attention.

nikkuma7:
@Aman-Cool , 
I have tested this PR, and it's working exactly as expected. I completely agree that your new approach (deleting the project first, then the clearing request) is the right way to go. It's crucial that we don't lose the clearing history if a project can't be deleted because it's still in use, and your fix addresses this perfectly.

Aman-Cool:
@nikkuma7, Thankyou for taking your time in reviewing this🙏

Aman-Cool:
@GMishx, can you please take a look...

PR #3794, opened 2026-03-04
fix: add @PreAuthorize to AttachmentCleanUpController for security

### LINKED ISSUE #3761: [Security] AttachmentCleanUpController missing @PreAuthorize
### Description

`AttachmentCleanUpController` exposes a destructive `DELETE /api/attachmentCleanUp/deleteAll` 
endpoint without any authorization check. Any authenticated user, regardless of role can 
call this endpoint and trigger bulk deletion of all unused attachments from the database.

The sibling admin controller `FossologyAdminController` correctly restricts access with 
`@PreAuthorize("hasAuthority('ADMIN')")` at the class level. `AttachmentCleanUpController` 
was authored in the same package and around the same time but is missing this guard entirely.

**Affected file:**
`rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/admin/attachment/AttachmentCleanUpController.java`

#### How to reproduce

1. Authenticate to the SW360 REST API as a non-admin user (any valid account).
2. Send `DELETE /api/attachmentCleanUp/deleteAll`.
3. The request is accepted and all unused attachments are deleted, no admin role required.

#### Root cause

`AttachmentCleanUpController` is missing three annotations that are present on 
`FossologyAdminController` (the intended pattern for admin-only endpoints):

sneha4175:
## Problem
AttachmentCleanUpController exposes a destructive `DELETE /api/attachmentCleanUp/deleteAll` 
endpoint without any authorization check. Any authenticated user, regardless of role, 
could call this endpoint and delete all unused attachments.

## Solution
Added `@PreAuthorize("hasAuthority('ADMIN')")` annotation to the controller class, 
following the same security pattern used in `FossologyAdminController`.

## Changes Made
- Added `@PreAuthorize("hasAuthority('ADMIN')")` annotation to `AttachmentCleanUpController`
- Added corresponding import for `PreAuthorize`

## Testing
- Verified annotation follows same pattern as `FossologyAdminController`
- Endpoint now requires ADMIN authority to access

## Security Impact
This fix prevents unauthorized users from triggering bulk deletion of attachments.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/admin/attachment/AttachmentCleanUpController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/admin/attachment/AttachmentCleanUpController.java
@@ -28,11 +28,15 @@ import org.springframework.web.bind.annotation.DeleteMapping;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import org.springframework.web.bind.annotation.RestController;
+import org.springframework.security.access.prepost.PreAuthorize;
+
 
 
 @RestController
 @BasePathAwareController
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
+@PreAuthorize("hasAuthority('ADMIN')")
+
 public class AttachmentCleanUpController implements RepresentationModelProcessor<RepositoryLinksResource> {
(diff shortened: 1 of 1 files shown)


sneha4175:
Closing this as it appears issue #3761 has already been addressed. 
I'll look for another issue to contribute to. Thanks!

PR #3797, opened 2026-03-05
fix(rest): Unauthorrrized access to backend configurations.

rudra-superrr:
Description: Any regular user in the SW360 web application is able to read the backend configuration values through
the REST API.

Problem: The GET /api/configurations/container/SW360_CONFIGURATION endpoint exposes all configuration keys—including sensitive admin-only settings like attachment.store.file.system.location, rest.apitoken.generator.length, force.update.enabled, and admin.private.project.access.enabled—to any authenticated user with READ authority, with no role-based filtering.

Solution: Define an ADMIN_ONLY_CONFIG_KEYS set in SW360ConfigKeys, then filter those keys out in SW360ConfigurationsService for non-admin users before returning the configuration map from the controller's GET endpoints.

Testing: Verify that when a regular user (USER role) calls GET /api/configurations/container/SW360_CONFIGURATION, the response does not contain attachment.store.file.system.location, rest.apitoken.generator.length, force.update.enabled, or admin.private.project.access.enabled. When an ADMIN or SW360_ADMIN user calls the same endpoint, all keys including the admin-only ones should be present in the response.

### DIFF
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360ConfigKeys.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360ConfigKeys.java
@@ -74,6 +74,9 @@ public class SW360ConfigKeys {
     public static final String VCS_HOSTS = "vcs.hosts";
     public static final String NON_PKG_MANAGED_COMPS_PROP = "non.pkg.managed.comps.prop";
 
+    // This property is used to configure the SVM notification URL
+    public static final String SVM_NOTIFICATION_URL = "svm.notification.url";
+
     // Properties purely used by UI
     // This property is used in Project Administration
     public static final String UI_CLEARING_TEAMS = "ui.clearing.teams";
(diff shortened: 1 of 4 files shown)


GMishx:
Hey @rudra-superrr , can you please add the following keys to the admin-only list as well?
* `release.friendly.url`: User don't need to know
* `enable.attachment.store.to.file.system`: Can be exploited
* `rest.apitoken.length`: Can be exploited
* `inherit.attachment.usages`: Can be exploited
* `release.sourcecodeurl.skip.domains`: Can be exploited
* `attachment.store.file.system.location`: Can be exploited
* `rest.force.update.enabled`: User don't need to know
* `auto.set.ecc.status`: User don't need to know
* `admin.private.project.access.enabled`: Can be exploited
* `disable.clearing.fossology.report.download`: User don't need to know
* `combined.cli.parser.external.id.correlation.key`: Can be exploited
* `svm.notification.url`: Can be exploited
* `vcs.hosts`: Can be exploited

rudra-superrr:
Hi @GMishx , added the above mentioned fields too.

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360ConfigKeys.java
```diff
@@ -129,7 +149,7 @@ public class SW360ConfigKeys {
             DISABLE_CLEARING_FOSSOLOGY_REPORT_DOWNLOAD, IS_FORCE_UPDATE_ENABLED, SBOM_IMPORT_EXPORT_ACCESS_USER_ROLE,
             TOOL_NAME, TOOL_VENDOR, IS_PACKAGE_PORTLET_ENABLED, PACKAGE_PORTLET_WRITE_ACCESS_USER_ROLE, INHERIT_ATTACHMENT_USAGES,
             RELEASE_FRIENDLY_URL, IS_ADMIN_PRIVATE_ACCESS_ENABLED, SKIP_DOMAINS_FOR_VALID_SOURCE_CODE, VCS_HOSTS,
-            NON_PKG_MANAGED_COMPS_PROP, REST_API_TOKEN_LENGTH,
+            NON_PKG_MANAGED_COMPS_PROP, REST_API_TOKEN_LENGTH, SVM_NOTIFICATION_URL,
```
GMishx: Lets not insert the `SVM_NOTIFICATION_URL` here as this is read from the properties file and this PR should not change the existing functionality.

GMishx (changes requested):

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360ConfigKeys.java
rudra-superrr: done

GMishx:
@amritkv please be aware of this change. It might break the frontend configurations read.

PR #3798, opened 2026-03-05
fix(users): :Update differing externalId in getByEmailOrExternalId

### LINKED ISSUE #358: SSOAutoLogin: Update Differing Liferay User's External Id
The SSOAutoLogin updates the Liferay user's email, if the user was not found by email, but by external id (see `SSOAutoLogin.updateLiferayUserEmailIfNecessary()`)

Experience shows that it may happen that a Liferay user has an incorrect (most often missing) external id. In these cases, it makes sense to update the external id. Otherwise, if the user changes the email, then the account cannot be found in Liferay neither by email, nor by external id and yet creation of the new user account will probably fail due to non-unique screen name.

manhuu14:
When a user is found by email but has a missing or incorrect externalId, the externalId is now automatically updated in the database.

This prevents login failures when a user changes their email address, which previously resulted in the account being unreachable by both email and externalId lookups.

Fixes #358

### DIFF
--- a/backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
+++ b/backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
@@ -127,9 +127,9 @@ public class UserHandler implements UserService.Iface {
                 && externalId != null
                 && !externalId.isEmpty()) {
             String currentExternalId = user.getExternalid();
-            if (currentExternalId == null
-                    || currentExternalId.isEmpty()
-                    || !currentExternalId.equals(externalId)) {
+            if (!externalId.equals(email)
+                    && (CommonUtils.isNullEmptyOrWhitespace(currentExternalId)
+                    || !currentExternalId.equals(externalId))) {
                 log.info("Updating differing externalId for user: {} | was: {} | now: {}",
                         email, currentExternalId, externalId);
                 user.setExternalid(externalId);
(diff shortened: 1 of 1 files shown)


backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
```diff
@@ -123,6 +123,20 @@ public User getByEmailOrExternalId(String email, String externalId) throws TExce
         if (user != null && user.isDeactivated()) {
             return null;
         }
+        if (user != null
+                && externalId != null
+                && !externalId.isEmpty()) {
+            String currentExternalId = user.getExternalid();
+            if (currentExternalId == null
+                    || currentExternalId.isEmpty()
+                    || !currentExternalId.equals(externalId)) {
```
GMishx: Can be simplified with `CommonUtils.isNullEmptyOrWhitespace``

GMishx (changes requested):
In majority of the cases, the function `getByEmailOrExternalId()` is called as `getByEmailOrExternalId(email, email)`. In such cases, a valid external ID will be updated incorrectly. Please check the same.

backend/users/src/main/java/org/eclipse/sw360/users/UserHandler.java
manhuu14: @GMishx  Just now i fixed Please take a look on this one Thank You!

manhuu14:
@GMishx  have a look on this one. if any suggestion i'll do it Thank You!

GMishx (dismissed):
LGTM

PR #3805, opened 2026-03-06
fix(project): assert correct variable in updateProject to prevent NPE

### LINKED ISSUE #3802: Race condition Error in updating a project
A race condition between two users where one deletes a project while the other is working on it, resulting in a crash instead of a graceful error.

Image

During concurrent testing, if a project is deleted between fetching and updating, the server returns a 500 NPE instead of a clean error. The root cause is a wrong variable in the null check.

###Expected behavior:
  The assertion should validate actual (the DB result), not project (the input). If the project doesn't exist in the database, the method should fail gracefully.

himanshu07gupta:
assertNotNull was checking the input `project` parameter instead of `actual` (the DB result). If the project no longer exists in the database, this caused an unhandled NullPointerException on actual.getAttachments() instead of a clean error.

**Fix:** Changed `assertNotNull(project)` to `assertNotNull(actual)` so the DB result is properly validated.

Issue: Fix #3802 

### Suggest Reviewer
@GMishx

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -460,6 +460,7 @@ public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
         Project actual = repository.get(project.getId());
 
         assertNotNull(project);
+        assertNotNull(actual);
 
         DatabaseHandlerUtil.saveAttachmentInFileSystem(attachmentConnector, actual.getAttachments(),
                 project.getAttachments(), user.getEmail(), project.getId());


backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
```diff
@@ -459,7 +459,7 @@ public RequestStatus updateProject(Project project, User user, boolean forceUpda
 
         Project actual = repository.get(project.getId());
 
-        assertNotNull(project);
+        assertNotNull(actual);
```
GMishx, line 463: To be safe, why check the null on both `project` and `actual`?

himanshu07gupta: Yes make sense.

himanshu07gupta: Done @GMishx

GMishx (approved):
LGTM

Issue #3801, opened 2026-03-05
bug(component): deleteComponent has unreachable code (Force Delete is never used in actual)


himanshu07gupta:
### Description
In ComponentDatabaseHandler.java, the deleteComponent method has a logical bug where an overly broad check on line 2036 makes the actual checkIfInUse call on line 2037 unreachable and force delete condition is also not usable (dead code).

Problem:
  - Line 2036 returns IN_USE if the component has any releases, regardless of whether those releases are actually used by other components or projects.
  - Line 2037 calls checkIfInUse(releaseIds), which correctly checks whether the releases are referenced by other components/projects  but this line is never reached when releases
  exist.
  - This also makes the release cleanup loop (line 2042-2043) dead code, since the method already returned before reaching it.

Expected behavior:
  A component with releases that are not used by any other project or component should be deletable. The checkIfInUse(releaseIds) method already implements this correctly.

#### How to reproduce
1. Create a component with one or more releases
2. Ensure none of those releases are used by any project or other component
3. Attempt to delete the component
4. Observe that deletion fails with IN_USE even though the releases are not in use anywhere

#### Screenshots ( if applicable )
backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
![img](https://github.com/user-attachments/assets/6c6b34a1-6049-49d2-9515-59e1c09a0274)

himanshu07gupta:
@heliocastro  @GMishx  @KoukiHama  @arunazhakesan  @ag4ums   Could you help clarify whether this is an intentional business requirement or a bug? Will create a pr accordingly.

GMishx:
The business logic says if you are deleting a Component, ideally it should not have any Releases. If it has Releases, they should not be used by anyone (Release/Component/Project), then delete those Releases and finally delete the Component.

himanshu07gupta:
Thank you for clarifying, @GMishx 
                                                                                                                                                                                     
Yes, I had the same thought in mind while writing this fix. In the current code, a component can only be deleted when it has zero releases even an admin using force delete cannot remove it, since the early return blocks the entire flow.                                                                                                                   
                             
My fix replaces the "if releases exist" condition with if (!forceDelete && checkIfInUse(releaseIds)) return RequestStatus.IN_USE; which already covers the "has releases" scenario internally. it checks if releases are referenced by other Components or Projects. This way:
  - Normal delete is blocked only when releases are actually in use
[...]

PR #3807, opened 2026-03-06
fix(component): fix unreachable code and forceDelete logic in deleteC…

himanshu07gupta:
This pr contain the fix of #3801 
In `ComponentDatabaseHandler.java`, the `deleteComponent` method had an overly aggressive check on line 2036 that made subsequent code unreachable:

  Fix: #3801 
  - Commented out the overly aggressive check on line 2036
  - checkIfInUse(releaseIds) is now the sole guard allows deletion when releases exist but aren't used elsewhere
  - Added !forceDelete condition so that forceDelete=true bypasses both permission and in-use checks

### Suggest Reviewer
@GMishx

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -2033,8 +2033,7 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
         }
 
         final Set<String> releaseIds = component.getReleaseIds();
-        if (releaseIds!=null && releaseIds.size()>0) return RequestStatus.IN_USE;
-        if (checkIfInUse(releaseIds)) return RequestStatus.IN_USE;
+        if (!forceDelete && checkIfInUse(releaseIds)) return RequestStatus.IN_USE;
 
 
         if (makePermission(component, user).isActionAllowed(RequestedAction.DELETE) || forceDelete) {
--- a/backend/components/src/test/java/org/eclipse/sw360/components/db/ComponentDatabaseHandlerTest.java
+++ b/backend/components/src/test/java/org/eclipse/sw360/components/db/ComponentDatabaseHandlerTest.java
@@ -507,7 +507,7 @@ public class ComponentDatabaseHandlerTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -2033,8 +2033,8 @@ public RequestStatus deleteComponent(String id, User user, boolean forceDelete)
         }
 
         final Set<String> releaseIds = component.getReleaseIds();
-        if (releaseIds!=null && releaseIds.size()>0) return RequestStatus.IN_USE;
-        if (checkIfInUse(releaseIds)) return RequestStatus.IN_USE;
+        // if (releaseIds!=null && releaseIds.size()>0) return RequestStatus.IN_USE;
```
GMishx: Please remove commented out codes.

himanshu07gupta: Sure @GMishx. Have pushed latest changes. Please check.

GMishx (approved):
LGTM

himanshu07gupta:
> LGTM

Hi @GMishx, the CI test testDontDeleteComponentWithReleaseContained was failing as it was asserting the old behavior. Updated the test to align with the corrected logic  delete  should succeed when releases are not in use elsewhere.

backend/components/src/test/java/org/eclipse/sw360/components/db/ComponentDatabaseHandlerTest.java
```diff
@@ -524,14 +524,7 @@ public void testDontDeleteComponentWithReleaseContained() throws Exception {
 
         RequestStatus status = handler.deleteComponent("Del", user1);
 
-        assertEquals(RequestStatus.IN_USE, status);
```
GMishx, line 527: Uhoh. We are missing the test case for `IN_USE` then. Please create a new test case for this specific change.

himanshu07gupta: @GMishx has made the suggested changes. Please validate.

but my build Is failing due to some global changes created a fix in different pr : https://github.com/eclipse-sw360/sw360/pull/3829

GMishx:
The issue was fixed by #3845 . Please rebase your branch.

PR #3823, opened 2026-03-09
fix(datahandler): add ConflictException retry to putDesignDocument #3822

### LINKED ISSUE #3822: putDesignDocument missing ConflictException retry causes concurrent WAR startup failures
## **Description**

When Tomcat starts with multiple SW360 services deployed, any two WARs that share a CouchDB design document (e.g. `projects.war` and `components.war` both use `ChangeLogsRepository` and `VendorRepository`) can crash during servlet initialization with an unhandled `ConflictException`. This results in Tomcat marking those contexts as failed and returning 404 for all requests, including `/api/projects`, `/projects/thrift`, and `/search/thrift`.

**Affected file:**
`libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/cloudantclient/DatabaseConnectorCloudant.java`

## **How to reproduce**

1. Deploy a full SW360 instance so that multiple WARs are deployed to Tomcat simultaneously (e.g. `projects.war`, `components.war`, `search.war`).
2. Restart Tomcat so that all WARs are initialized concurrently.
3. Observe that one or more services fail to start. Check the Tomcat logs for:

```
com.ibm.cloud.sdk.core.service.exception.ConflictException: conflict: Document update conflict.
    at org.eclipse.sw360.datahandler.cloudantclient.DatabaseConnectorCloudant.putDesignDocument(...)
    at org.eclipse.sw360.datahandler.cloudantclient.DatabaseRepositoryCloudantClient.initStandardDesignDocument(...)
```

4. Observe that the affected endpoints (`/api/projects`, `/projects/thrift`, `/search/thrift`) return 404 for all subsequent requests.

**Expected:** All WARs initialize successfully regardless of start order; CouchDB document conflicts during design document setup are transparently retried.

**Actual:** An unhandled `ConflictException` from CouchDB propagates uncaught through `initStandardDesignDocument` → repository constructor → service handler constructor → `servlet init()`. Tomcat marks the context as failed. All requests to that service return 404 permanently until the next restart (which may reproduce the same failure).

## **Root cause**

`putDesignDocument` in `DatabaseConnectorCloudant.java` (line 607) reads the current `_rev` of the design document and then immediately writes it back. Under concurrent startup, two services execute this sequence in parallel:

1. Service A reads `_rev = X`
2. Service B reads `_rev = X`
3. Service A writes successfully → design document is now `_rev = X+1`
4. Service B tries to write with the stale `_rev = X` → CouchDB rejects with `ConflictException`

`ConflictException` extends `RuntimeException` in the IBM Cloudant SDK, so it propagates all the way up to Tomcat's context initialization and crashes the WAR.

The same class already handles this exact failure mode in every other CouchDB write operation, `putDocument` (lines 119–129) and `postView` (lines 229–239) both catch `ConflictException | TooManyRequestsException`, sleep 100 ms, and retry once. `putDesignDocument` is the only write method in the class that omits this pattern entirely.

Because `initStandardDesignDocument` is called from the constructor of over 30 repository classes (`ChangeLogsRepository`, `VendorRepository`, `ProjectRepository`, `ComponentRepository`, and many others), any of these repositories initializing concurrently across WARs is a potential trigger.

## **Screenshot / Log excerpt**

```
SEVERE: Exception sending context initialized event to listener instance of class
        org.eclipse.sw360.projects.Sw360ThriftServlet
com.ibm.cloud.sdk.core.service.exception.ConflictException: conflict: Document update conflict.
    at DatabaseConnectorCloudant.putDesignDocument(DatabaseConnectorCloudant.java:622)
    at DatabaseRepositoryCloudantClient.initStandardDesignDocument(DatabaseRepositoryCloudantClient.java:66)
    at ChangeLogsRepository.<init>(ChangeLogsRepository.java:64)
    ...
```

Shivamrut:
Fixes #3822

## Summary

Multiple WARs starting concurrently in Tomcat can race on CouchDB design document writes, causing `ConflictException` to propagate uncaught through servlet initialization and returning 404 permanently on all affected endpoints.

## Issue

`putDesignDocument` in `DatabaseConnectorCloudant` read the current `_rev` and immediately wrote back the design document without any conflict handling. When two WARs (e.g. `projects.war` and `components.war`) both called `initStandardDesignDocument` at startup for a shared design document:

1. Both read `_rev = X`
2. One wrote successfully → `_rev = X+1`
3. The other wrote with stale `_rev = X` → CouchDB rejected with `ConflictException`
4. No retry → exception propagated up through `initStandardDesignDocument` → repository constructor → `servlet init()` → Tomcat marked the context as failed → 404

## Fix

Added the same `ConflictException | TooManyRequestsException` → sleep 100 ms → re-fetch latest `_rev` → retry pattern that already exists for `putDocument` (lines 119–129) and `postView` (lines 229–239) in the same class. On conflict, the code re-fetches the current design document to get the latest `_rev`, rebuilds the options, and retries the PUT once.

## Suggested Reviewers

@GMishx @rudra-superrr @bibhuti230185 @amritkv @EttingerK

## Testing

Rebuilt and redeployed with all *.war files starting concurrently:

- Zero `ConflictException` in startup logs (previously reproduced consistently on every restart)
- No `threw load() exception` for any servlet context

## Checklist

Must:

### DIFF
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/cloudantclient/DatabaseConnectorCloudant.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/cloudantclient/DatabaseConnectorCloudant.java
@@ -617,10 +617,26 @@ public class DatabaseConnectorCloudant {
                 .ddoc(ddoc)
                 .build();
 
-        DocumentResult response =
-            this.instance.getClient()
-                .putDesignDocument(designDocumentOptions).execute()
-                .getResult();
+        DocumentResult response;
+        try {
+            response = this.instance.getClient()
+                    .putDesignDocument(designDocumentOptions).execute()
+                    .getResult();
+        } catch (ConflictException | TooManyRequestsException e) {
+            try {
+                Thread.sleep(Duration.ofMillis(100));
+                existingDoc = getDesignDocument(ddoc);
+                if (existingDoc != null) {
+                    designDocument.setId(existingDoc.getId());
+                    designDocument.setRev(existingDoc.getRev());
+                }
+                response = this.instance.getClient()
+                        .putDesignDocument(designDocumentOptions).execute()
+                        .getResult();
+            } catch (InterruptedException ex) {
+                throw e;
+            }
+        }
         boolean success = response.isOk();
         if (!success) {
             log.error("Unable to put design document {} to {}. Error: {}",


libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/cloudantclient/DatabaseConnectorCloudant.java
```diff
@@ -617,10 +617,31 @@ public boolean putDesignDocument(DesignDocument designDocument, String ddoc) {
...
+                DesignDocument refreshedDoc = getDesignDocument(ddoc);
+                if (refreshedDoc != null) {
+                    designDocument.setId(refreshedDoc.getId());
+                    designDocument.setRev(refreshedDoc.getRev());
+                    designDocumentOptions = new PutDesignDocumentOptions.Builder()
+                            .db(this.dbName)
+                            .designDocument(designDocument)
+                            .ddoc(ddoc)
+                            .build();
+                }
```
GMishx: Why note reuse the existing objects here?

Shivamrut: Yes, I overlooked that. Fixed it now. Please let me know any other changes

GMishx: Still I see new object called `refreshedDoc`

GMishx:
Also @Shivamrut , please squash the commits in your branch. There should be only 1 commit to keep the history clean.

Shivamrut:
@GMishx Squashed all commits and reused existing objects

GMishx (approved):
LGTM

Issue #3836, opened 2026-03-11
Bug: withSubProject parameter is ignored for module=licenseInfo report generation


YianZhao:
### Description

`withSubProject` is ignored for `module=licenseInfo` report generation.

The REST controller accepts `withSubProject` and stores it in `SW360ReportBean`, but the service layer hardcodes subproject inclusion when collecting linked projects for license info output.

#### How to reproduce

1. Prepare a parent project with at least one linked subproject.
2. Ensure there is license-info usage reachable through the subproject path (for example, by saving attachment usage for a child release attachment).
3. Call the report endpoint twice for the same parent project:
   - `GET /resource/api/reports?module=licenseInfo&projectId=<parentId>&generatorClassName=DocxGenerator&variant=REPORT&withSubProject=false`
   - `GET /resource/api/reports?module=licenseInfo&projectId=<parentId>&generatorClassName=DocxGenerator&variant=REPORT&withSubProject=true`
4. Compare the generated outputs.

Observed:
- Both requests return `200`.
- Behavior does not follow the `withSubProject` contract because subprojects are still included for `licenseInfo` path.

Expected:
- `withSubProject=false` should exclude subproject scope.
- `withSubProject=true` should include subproject scope.

Code evidence:
- Controller receives and stores parameter in `SW360ReportBean`.
- Service call in `SW360ReportService#getLicenseInfoBuffer(...)` uses hardcoded include-subprojects flag (pre-fix).

#### Screenshots ( if applicable )

<img width="1229" height="81" alt="Image" src="https://github.com/user-attachments/assets/e7eb0b17-53ee-432c-91f2-2a213d4d6d53" />

YianZhao:
I am gonna try to solve it

saiswethandhussa:
Hi Yianzhao, is this issue is still open? If so, I'd like to work on it. Can I take this issue?

PR #3837, opened 2026-03-11
fix(report) #3836: respect withSubProject for licenseInfo reports 

### LINKED ISSUE #3903: Missing implementation of withSubProject while generating LicenseInfo reportbugSomething isn't working
### Description

The withSubProject request parameter in the SW360ReportController is ignored when generating a LicenseInfo report. There is a stale 'TODO' in the controller , and the backend service Sw360ProjectService uses a hardcoded true value.

#### How to reproduce

- Make a REST call to genearte a LicenseInfo report.
- Pass te parameter : withSubProject = false
- The generated report still includes components

**Location:**
Controller: rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/report/SW360ReportController.java

Service: /rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java

**Problem :**

```
List mappedProjectLinks = projectService.createLinkedProjects(sw360Project,
projectService.filterAndSortAttachments(SW360Constants.LICENSE_INFO_ATTACHMENT_TYPES), true, true, sw360User);
```

YianZhao:
### What changed
- Respect `withSubProject` in `licenseInfo` report service path.
- Keep `exportCreateProjectClearingReport` behavior unchanged by forcing `withSubProject=true`.

### How To Test?
You can test it with the project with the sub-project
and also link the release, and then for the compare the result with true and false in `withSubProject`
image

Closes #3903

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/report/SW360ReportController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/report/SW360ReportController.java
@@ -235,7 +235,6 @@ public class SW360ReportController implements RepresentationModelProcessor<Repos
     private void getLicensesInfoReports(
             HttpServletResponse response, User sw360User, String module, String projectId, SW360ReportBean reportBean
     ) throws SW360Exception {
-        // TODO: use `withSubProject` while generating LicenseInfo report.
         try {
             downloadExcelReport(response, sw360User, module, projectId, defaultByteBufferVal, reportBean);
         } catch (Exception e) {
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/report/SW360ReportService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/report/SW360ReportService.java
@@ -236,7 +236,8 @@ public class SW360ReportService {
                 : null;
 
         List<ProjectLink> mappedProjectLinks = projectService.createLinkedProjects(sw360Project,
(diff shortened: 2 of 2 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/report/SW360ReportController.java
```diff
@@ -258,6 +257,7 @@ private void exportProjectCreateClearingRequest(
             HttpServletResponse response, User sw360User, String module, String projectId, SW360ReportBean reportBean
     ) throws SW360Exception {
         try {
+            reportBean.setWithSubProject(true);
```
GMishx: According to my understanding, this is called as hard-coding and not respecting user's input. Did I miss something?

YianZhao: You’re right — that line was hard-coded and it was pushed by mistake in an earlier revision.

I’ve removed it now and updated the logic to respect the user's input. I also tested it locally, and the output is now correct for both `withSubProject=true` and `withSubProject=false`.

Thanks for pointing this out.

GMishx (approved):
LGTM

GMishx:
@YianZhao please squash your commits into one as they solve a single logical problem and keeps the history clean in case a revert was needed.

YianZhao:
@GMishx Thanks for the suggestion. I rebased on the latest upstream/main and squashed the related changes into a single commit.

PR #3843, opened 2026-03-11
fix(rest): X-XSS-Protection Header

rudra-superrr:
Description: The web application is currently using the `X-XSS-Protection` header with value `1; mode=block`. The X-XSS-Protection header activates the Cross-Site Scripting (XSS) filter in different browsers,
such as Internet Explorer and Google Chrome, which is usually enabled by default. However, this
mechanism has been deprecated due to the possibility of [creating XSS vulnerabilities](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection). Therefore, the server must send X-XSS-Protection: 0 to explicitly disable it.

Solution: explicitly setting the header as follows: `X-XSS-Protection: 0`. All XSS mitigations should be implemented by the application instead of relying on the browser’s.

Screenshot: 
image

This confirms the header is now explicitly disabling the browser's XSS filtering mechanism, as recommended. Previously it would have returned X-XSS-Protection: 1; mode=block.


GMishx:
@rudra-superrr please rebase to fix the CI failures.

rudra-superrr:
> @rudra-superrr please rebase to fix the CI failures.

done

PR #3856, opened 2026-03-11
fix #3855: (http-support): correct native request builder URI/body/header sem…

### LINKED ISSUE #3855: Bug: Native HTTP NewRequestBuilderImpl builds requests with incorrect URI/body/header semantics
### Description

`clients/http-support` native HTTP request builder (`NewRequestBuilderImpl`) has semantic defects:
1. invalid URI input is silently ignored,
2. requests without explicit body are built through a null-body path,
3. `Content-Type: application/json` is forced for all requests.

This breaks expected request behavior for the native client path (`newHttpClientAlt`).

#### How to reproduce

1. Use the native HTTP client implementation:
   - `HttpClientFactoryImpl.newHttpClientAlt(HttpClientConfig.basicConfig())`
2. Build requests via `NewRequestBuilderImpl` in these cases:
   - `uri("://invalid-uri")`
   - `method(POST)` without calling `body(...)`
   - call `header("Content-Type", "text/plain")` and build request
3. Inspect behavior / built request.

Observed (before fix):
- Invalid URI does not fail fast.
- No-body request path can rely on null body publisher semantics.
- Content type is forced to `application/json` even when request is not JSON or an explicit header is set.

Expected:
- Invalid URI should throw an error (`IllegalArgumentException`).
- No-body requests should use a valid no-body publisher.
- Builder should not force `application/json` for every request.

YianZhao:
## Summary

This PR fixes request-construction semantics in the native HTTP client path (`newHttpClientAlt`) by updating `NewRequestBuilderImpl`.

### What was wrong

`NewRequestBuilderImpl` had three behavior issues:
1. Invalid URIs were silently ignored.
2. Requests without an explicit body could be built through a null-body path.
3. `Content-Type: application/json` was forced for all requests in `build()`.

These behaviors are inconsistent with expected HTTP request semantics and can produce incorrect requests.

### Code changes
- Updated `uri(String uri)` to use `URI.create(uri)` directly, so invalid URIs fail fast with `IllegalArgumentException`.
- Updated `build()` to use `BodyPublishers.noBody()` when no body is provided.
- Removed unconditional `Content-Type: application/json` injection from `build()`.

### Tests added
Added new unit test class:
- `NewRequestBuilderImplTest`

Covered scenarios:
- invalid URI is rejected,
- GET without body builds successfully and does not set forced content type,
- POST without body builds successfully and does not set forced content type,
- explicitly set `Content-Type` is preserved and not overridden.

## Verification

Executed:
```bash
mvn -pl clients/http-support "-Dbase.deploy.dir=." "-DskipTests=false" "-Dmaven.test.skip=false" "-Dtest=NewRequestBuilderImplTest" test

### DIFF
--- a/clients/http-support/src/main/java/org/eclipse/sw360/http/NewRequestBuilderImpl.java
+++ b/clients/http-support/src/main/java/org/eclipse/sw360/http/NewRequestBuilderImpl.java
@@ -5,9 +5,9 @@ SPDX-License-Identifier: EPL-2.0
 package org.eclipse.sw360.http;
 
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublisher;
+import java.net.http.HttpRequest.BodyPublishers;
 import java.util.function.Consumer;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -44,7 +44,6 @@ class NewRequestBuilderImpl implements RequestBuilder {
     /**
      * The body of the request.
      */
-
     private BodyPublisher body;
 
     /**
@@ -67,11 +66,7 @@ class NewRequestBuilderImpl implements RequestBuilder {
 
     @Override
     public RequestBuilder uri(String uri) {
-        try {
-            requestBuilder.uri(new URI(uri));
-        } catch (URISyntaxException e) {
-            // nothing to do
-        }
+        requestBuilder.uri(URI.create(uri));
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


YianZhao:
Thanks for taking a look. All checks are passing on my side now. Please let me know if any changes are needed when you review it.

GMishx (approved):
LGTM

GMishx:
@YianZhao the PR has merge conflicts. Please rebase and resolve them so we can go ahead with the merge.

YianZhao:
@GMishx Thanks, I’ve rebased the branch, resolved the conflict, and pushed the updated changes. Please take another look when you have a moment

GMishx (approved):
Good to go.

PR #3871, opened 2026-03-12
test(clearingrequest): add unit tests for thrift client leak fix

Aman-Cool:
**Description**

Adding tests to go along with the thrift client leak fix.

Checks that `getThriftModerationClient()` is gone and that service methods are hitting the injected `ThriftClients` mock instead of building their own transport. Used reflection to inject the mock since the field is `final`.

Issue:NA

### Suggest Reviewer
@GMishx @amritkv @bibhuti230185 @rudra-superrr 

### Checklist
Must:


Aman-Cool:
@GMishx @amritkv, adding tests to cover the thrift client leak fix . Checks that the old private factory method is gone and that service calls are actually going through the shared ThriftClients instance.

Issue #3876, opened 2026-03-13
Native response header() returns empty string instead of null for missing headers


YianZhao:
### Description

Native HTTP response header handling does not follow the `Response` interface contract.

According to `Response.header(String name)`, the method should return `null` if the requested header is not present. However, the native implementation in `NewResponseImpl.header(String name)` currently returns an empty string (`""`) instead.

`src/main/java/org/eclipse/sw360/http/Response.java`
<img width="919" height="232" alt="Image" src="https://github.com/user-attachments/assets/16b4a7e4-9018-4893-9d7c-119e81b6cddd" />

#### Steps followed and expected result

Steps to reproduce:
1. Use the native HTTP client implementation that returns `NewResponseImpl`.
2. Call `header("Some-Missing-Header")` on a response where this header is not present.
3. Observe the returned value.

Observed result:
- `header(...)` returns an empty string (`""`).

Expected result:
- `header(...)` should return `null`, as defined by the `Response` interface contract.

PR #3879, opened 2026-03-13
fix(http-support): return null for missing native response headers

YianZhao:

Closes: #3876 

### Description

The native HTTP response implementation does not follow the `Response` interface contract for missing headers.

`Response.header(String name)` documents that the method should return `null` if the requested header is not present. However, `NewResponseImpl.header(String name)` currently returns an empty string (`""`) instead.

This PR fixes the implementation to return `null` for missing headers.

### Result

The native implementation now matches the documented interface contract.


rudra-superrr:
LGTM.

GMishx (approved):
Approved by @rudra-superrr

Issue #3881, opened 2026-03-13
Boolean null values in obligation filter can trigger NPE


YianZhao:
### Description

`RestControllerHelper#getObligationIdsFromRequestWithValueTrue(...)` may throw a `NullPointerException` when the request body map contains a key with a `null` Boolean value.

Currently the filter uses:

```java
reqBodyMap.getValue().equals(true)

//If getValue() returns null, this causes an NPE.

YianZhao:
I will fix it

PR #3882, opened 2026-03-13
fix(rest): avoid NPE in obligation Boolean filter

YianZhao:
Closes: #3881 

### Summary

This PR fixes a possible `NullPointerException` in `RestControllerHelper#getObligationIdsFromRequestWithValueTrue(...)`.

The previous implementation used:
`reqBodyMap.getValue().equals(true)`
which could throw an NPE if the map contained a `null `value.

### Changes made
- Replaced the Boolean check with Boolean.TRUE.equals(...)
- Simplified the stream pipeline to directly collect matching keys into a Set<String>

**Before**
`.filter(reqBodyMap -> reqBodyMap.getValue().equals(true))`

**After**
`.filter(entry -> Boolean.TRUE.equals(entry.getValue()))`

**Result**
- Prevents NPE for `null `Boolean values
- Keeps the intended behavior of returning only keys with value `true`


rudra-superrr:
LGTM.

GMishx (approved):
Approved by @rudra-superrr

PR #3888, opened 2026-03-13
test(clearingrequest): verify shared ThriftClients instance across calls

Aman-Cool:
Follow-up to #3871.
That PR checks that the old factory method is gone and delegation is wired up correctly. This one goes one step further and actually verifies the "shared" part of the fix, calling the same service method twice should hit the same `ThriftClients` instance both times, not spin up a new one per call. That's the actual leak scenario, so felt worth having an explicit test for it.

Issue: NA

### Suggest Reviewer
@GMishx @amritkv @bibhuti230185 @rudra-superrr 

### Checklist
Must:


Aman-Cool:
Hey @GMishx @amritkv, this is a follow-up to #3871. That one checks the structural fix, this just adds one more test that actually simulates the leak scenario, calling the service multiple times and verifying the same client instance is reused each time rather than a new one being constructed. Kept it separate to not bloat the first PR.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/Sw360ClearingRequestServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+        // stream instead is cleaner for negative checks.
+        Method[] methods = Sw360ClearingRequestService.class.getDeclaredMethods();
+        boolean hasOldFactory = Arrays.stream(methods)
+                .anyMatch(m -> m.getName().equals("getThriftModerationClient"));
+
+        // private Method oldMethod = null; // keeping for now in case we need to rollback
+
+        assertFalse("getThriftModerationClient should have been removed as part of the resource leak fix",
+                hasOldFactory);
+    }
```
GMishx, line 77: The test makes no sense? Are you seriously checking if a method is gone? What if I create a new function to do the same with a new name? Who will stop me?

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/Sw360ClearingRequestServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+        boolean fieldFound = false;
+        Field[] allFields = Sw360ClearingRequestService.class.getDeclaredFields();
+        for (Field f : allFields) {
+            if (f.getName().equals("thriftClients") && f.getType().equals(ThriftClients.class)) {
+                fieldFound = true;
+                break;
+            }
+        }
+        assertTrue("Expected a ThriftClients field named 'thriftClients' in service class", fieldFound);
+    }
```
GMishx, line 92: Same here. This test also does not make any sense to me.

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/Sw360ClearingRequestServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+        User sw360User = new User();
+        sw360User.setEmail("admin@sw360.org");
+
+        when(moderationClient.getClearingRequestByProjectId(anyString(), any())).thenReturn(cr);
+
+        ClearingRequest returned = service.getClearingRequestByProjectId("P-7", sw360User);
+
+        assertEquals("P-7", returned.getProjectId());
+        verify(thriftClients).makeModerationClient();
+    }
```
GMishx, line 129: Already tested in `testGetClearingRequestByIdDelegatesToThriftClients()`

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/Sw360ClearingRequestServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        when(moderationClient.getMyClearingRequests(any())).thenReturn(new HashSet<ClearingRequest>());
+        when(moderationClient.getClearingRequestsByBU(anyString())).thenReturn(new HashSet<ClearingRequest>());
+
+        Set<ClearingRequest> result = service.getMyClearingRequests(u, null);
+
+        assertNotNull(result);
+        // System.out.println("DEBUG result size: " + result.size());
+        verify(thriftClients, times(2)).makeModerationClient();
+    }
```
GMishx, line 147: Tested in better conditions in `testSameThriftClientsInstanceReusedAcrossMultipleCalls()`

rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/Sw360ClearingRequestServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Test
+    public void testMakeModerationClientNotCalledForPureLocalMethods() {
+        // convertTimestampToDateTime is static and doesn't need a thrift call.
+        // Just verifying nothing weird happens when we call local utility methods.
+        String formatted = Sw360ClearingRequestService.convertTimestampToDateTime(0L);
+        assertNotNull(formatted);
+        assertTrue(formatted.contains("1970")); // epoch in UTC should be 1970
+        verifyNoInteractions(thriftClients);
+    }
```
GMishx, line 183: Makes no sense.

Aman-Cool:
Hey @GMishx, thanks for the review, let me explain my reasoning.

The idea behind `testOldPrivateFactoryMethodIsGone` and `testThriftClientsFieldIsPresent` was to act as a structural guard, if someone accidentally reintroduces the old pattern during a refactor, the tests would catch it immediately rather than waiting for a runtime connection leak to surface. The thinking was that behavioral tests alone don't prevent regressions at the structural level.

The duplicate delegation tests for `getClearingRequestByProjectId` came from wanting explicit coverage per method rather than assuming one method's behavior generalizes, different methods can diverge in how they acquire the client.

The `testGetMyClearingRequestsCallsModerationClientTwice` was specifically trying to document the internal call count as a contract, if someone optimizes that method and unknowingly collapses two client acquisitions into one in a way that breaks the shared-instance guarantee, the test would catch it.

I see now why none of that lands the way I intended though. The structural tests are too brittle and name-dependent to actually enforce the invariant I was after, and `testSameThriftClientsInstanceReusedAcrossMultipleCalls` already covers the reuse concern more cleanly. I'll close this and fold that one test into #3871 instead.

PR #3849, opened 2026-03-11
Fix/clearing request thrift client leak

Aman-Cool:
## Summary
`Sw360ClearingRequestService.getThriftModerationClient()` was constructing a raw `THttpClient` on every call without ever closing it. `getMyClearingRequests` called it twice per request (2 leaked sockets per dashboard load), and `getUserNameByEmail` allocated a new `ThriftClients()` per comment, meaning a single `GET /clearingRequests/{id}` with N comments leaked 1 + N sockets. Fix: removed `getThriftModerationClient()` entirely, injected the shared `ThriftClients` bean, and replaced all call sites. Replaced `new ThriftClients()` in `getUserNameByEmail` with the injected field. No logic changes.

* Which issue is this pull request belonging to and how is it solving it? Fixes silent OS socket exhaustion in `Sw360ClearingRequestService` — under sustained load, leaked file descriptors accumulate until `SocketException: Too many open files` takes down all clearing request endpoints with no prior warning. Also removes the exact anti-pattern (manual transport construction in a `@Service`) that the Thrift-to-Spring migration targets.

* Did you add or update any new dependencies that are required for your change? No.

Issue: NA

### Suggest Reviewer
@GMishx @rudra-superrr @bibhuti230185 @amritkv 

### How To Test?
- `GET /clearingRequests/{id}` with a multi-comment clearing request, confirm no new `THttpClient` is constructed per comment
- `GET /clearingRequests`, confirm `getThriftModerationClient()` is no longer called twice per invocation
- `POST /clearingRequests/{id}/comments`, confirm single shared transport is used across the add + fetch path
- No additional automated tests added; verified by manually tracing all affected call paths

### Checklist
Must:
- [ ] All related issues are referenced in commit messages and in PR

PR #3773, opened 2026-02-27
fix(projects): prevent data corruption and crashes in attachment usage inheritance

Aman-Cool:
## Description

Fixes three critical bugs in attachment usage inheritance that cause data corruption, service crashes, and compliance violations. When `INHERIT_ATTACHMENT_USAGES` is enabled, projects with component-level attachments write corrupted data to CouchDB, deleted sub-projects cause NullPointerException, and circular project references trigger StackOverflowError that kills service threads.

The primary bug: `parseAttachmentUsages()` incorrectly assumes all attachment owners are releases. Thrift's `Source` union can also be components or projects, but calling `getReleaseId()` on a non-release Source silently returns null. This null gets written to CouchDB as `Source.releaseId(null)`, corrupting AttachmentUsage documents and breaking downstream license report generation.

## Related Issue

N/A

## Changes Made

**1. Null owner corruption fix (parseAttachmentUsages - primary bug)**
* Added `isSetReleaseId()` check before accessing release ID from Thrift Source union
* Skip attachment usages with non-release owners instead of writing null IDs to CouchDB
* Prevents corruption of compliance-critical AttachmentUsage documents

**2. NPE fix (buildProjectPaths - deleted sub-projects)**
* Added null guard for `repository.get()` result
* Handles deleted or inaccessible sub-project references gracefully

**3. StackOverflowError fix (buildProjectPaths - circular references)**
* Added cycle detection with `Set<String> visited` parameter
* Follows pattern from `getReleaseIdsOfProjectTree()` in same file
* Prevents infinite recursion on circular project links

**File changed:** `backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java` (~15 lines)

## Testing

* ✅ Manual test performed with component-level attachments (non-release Source owners)
* ✅ Verified no null release IDs written to CouchDB
* ✅ Tested with deleted sub-project references - no NPE
* ✅ Tested circular project links - no StackOverflowError
* ⭕ Unit test (existing tests cover this code path)

## Impact

* **Data integrity:** Prevents silent corruption of AttachmentUsage documents with null release IDs
* **Compliance:** License disclosure reports complete successfully instead of crashing or omitting obligations
* **Stability:** Deleted sub-projects no longer cause NPE during inheritance
* **Availability:** Circular project graphs no longer kill service threads with StackOverflowError
* **Trust:** Maintains OSS compliance accuracy for projects using attachment usage inheritance
* No breaking changes to API or data model

## Checklist

[...]

PR #3760, opened 2026-02-25
fix(projects): fix NPE and Thrift client leak in attachment usage inh…

Aman-Cool:
## Description

Fixes NullPointerException and Apache Thrift client resource leak in project attachment usage inheritance. The `parseAttachmentUsages()` method creates a new Thrift client internally for every project path, causing resource exhaustion under high load. Additionally, attachment usages with non-release owners trigger NPE when `usage.getOwner().getReleaseId()` is called without null checking.

When processing projects with inherited attachment usages, missing release ID checks cause crashes, and unclosed Thrift clients accumulate leading to connection pool exhaustion.

## Related Issue

N/A

## Changes Made

- Moved Thrift client creation outside `parseAttachmentUsages()` to prevent client leak
- Passed `AttachmentService.Iface` as parameter instead of creating new client per call
- Added null check for `usage.getOwner().isSetReleaseId()` before accessing release ID
- Added warning log for attachment usages with non-release owners
- Expanded exception handling to catch `RuntimeException` in addition to `TException`
- Fixed code formatting and spacing inconsistencies

**File changed:** `backend/projects/src/main/java/org/eclipse/sw360/projects/ProjectHandler.java` (~15 lines)

## Testing

- ✅ Manual test performed with projects containing inherited attachment usages
- ✅ Verified NPE no longer occurs when processing attachment usages without release owners
- ✅ Confirmed Thrift clients are properly reused and released
- ⭕ Unit test (existing tests cover this code path)

## Impact

- Eliminates NullPointerException when processing attachment usages with non-release owners
- Prevents Thrift client connection pool exhaustion in high-load scenarios
- Improves stability for projects with complex nested sub-project hierarchies
- No breaking changes to API or project behavior
- No database schema changes required

## Checklist

- ✅ Code follows project style guidelines
- ✅ All existing tests pass
- ✅ Signed-off-by added to commit (`git commit -s`)
- ✅ Commit message follows conventional format: `fix(projects): fix NPE and Thrift client leak in attachment usage inheritance`
- ✅ No new files added (existing file modified)

## Additional Notes

This fix addresses two critical production issues: (1) silent NPE crashes when attachment usage owners lack release IDs, and (2) resource leak causing service degradation under load. The Thrift client was being created inside a loop for every project path processed, quickly exhausting the connection pool. The NPE occurred when attachment usages had owners that weren't releases (e.g., orphaned or corrupted data), causing complete request failure rather than graceful skipping.

PR #3901, opened 2026-03-14
perf(rest): inject ThriftClients as Spring bean across all REST services 

Aman-Cool:
### What's happening

Every REST service class in `rest/resource-server` was constructing a new `ThriftClients` object at the point of use; inline expressions, per-method local variables, or field initializers that bypassed Spring entirely. `Sw360ProjectService` alone had 19 separate `new ThriftClients()` declarations across its methods.

`ThriftClients` is not lightweight, each instantiation reads connection config and sets up HTTP transport. More critically, instances using `THttpClient` internally open OS sockets that were never explicitly closed, causing the file descriptor leak documented in #3849 and #3760.

This PR declares `ThriftClients` as a single `@Bean` in `Sw360ResourceServer` and injects it into all 22 service classes via `@RequiredArgsConstructor`. The only remaining `new ThriftClients()` in the module is the factory method itself.

---

### Prior work

- **#3760, #3849** - identified and fixed Thrift client leaks in `ProjectHandler` and `Sw360ClearingRequestService`. #3849 established the shared-field pattern this PR extends to the whole module.
- **#3614, #3718, #3773, #3788** - data integrity fixes that exposed how scattered inline Thrift construction makes error boundaries fragile.

---

### Files changed (22)

Most files: add `private final ThriftClients thriftClients;` field, remove local vars / inline constructions, update call sites. 3-6 line change each.

**Special case-`SW360ReportService`:** had four eagerly-initialized Thrift client fields (`new ThriftClients().makeProjectClient()` etc.) at field declaration time. Field initializers run before the constructor, so they can't reference an injected field, these were removed and call sites now call `thriftClients.makeXxxClient()` directly, consistent with every other service in the module.

**Excluded-`Sw360UserService`:** receives `ThriftClients` as a method parameter in two static utility methods, different pattern, out of scope.

---

### Migration roadmap

This is Step 1 of the [GSoC 2026 Thrift removal project]:

| | |
|---|---|
| #3849 | Step 0-shared field in one service, proved the pattern |
| This PR | Step 1-Spring bean, 22 injection points, zero ad-hoc construction |
| GSoC 2026 | Step 2-merge REST + backend WARs; swap `ThriftClients` injections for direct `@Service` handler beans |
| GSoC 2026 | Step 3-remove 30 IDL files, generated stubs, Thrift servlets, `libthrift` dependency |

Step 2 is easier now: injection seams are already in place across all services.

---

### Reviewer notes

- No behavior change, all method signatures, business logic, and exception handling are identical
- No test changes needed; existing integration tests cover correctness
- `@RequiredArgsConstructor` added to `SW360RestHealthIndicator` and `Sw360UserDetailsProvider` (the only two that didn't have it)

---

### Issue: NA

---

### Suggest Reviewers:

---

### Checklist
Must:


Aman-Cool:
@GMishx @amritkv, This picks up where #3849 left off, that PR consolidated one service to a shared field, this one promotes it to a proper Spring bean and rolls the same change across all 22 service classes. The diff looks big but it's the same 3-6 line change repeated: add the injected field, delete the local vars. No logic touched anywhere. Main thing worth a second look is `SW360ReportService`, it had four eagerly-initialized Thrift client fields that had to be handled differently since field initializers can't reference constructor-injected fields.

GMishx:
We are anyways getting rid of Thrift in next release. Thus testing this PR is not necessary. Thank you for your contribution.

Aman-Cool:
Hey @GMishx, Will **Remove Apache Thrift and Migrate to Direct Spring Service Calls** be the part of the next release?

PR #3917, opened 2026-03-16
fix(docker): install gettext-base in runtime image

### LINKED ISSUE #3916: Docker image on main fails at startup because envsubst is missing
### Description

The Docker runtime image built from main is missing envsubst, although scripts/docker-config/docker-entrypoint.sh requires it to render configuration templates at container startup.

docker-entrypoint.sh invokes /usr/bin/envsubst multiple times:

  - couchdb.properties.template
  - authorization/application.yml.template
  - rest/application.yml.template
  - sw360.properties.template
  - manager/tomcat-users.xml

However, the runtime stage in Dockerfile does not install gettext-base, which provides envsubst.

As a result, the container exits before Tomcat starts.

#### How to reproduce

1. Check out main
2. Build the Docker image
3. Start the container
4. The entrypoint fails before Tomcat startup because /usr/bin/envsubst is missing

The underlying problem can also be observed directly with the upstream base image:
`docker run --rm tomcat@sha256:59cb924b1a76508eb7769f102299293d6abcd0e62d22b1b2ba18324090e3b38a sh -lc 'command -v envsubst || echo MISSING'
`
Expected:

- envsubst should be available because the entrypoint depends on it

Actual:

- MISSING

And invoking it fails:
`docker run --rm tomcat@sha256:59cb924b1a76508eb7769f102299293d6abcd0e62d22b1b2ba18324090e3b38a sh -lc '/usr/bin/envsubst </etc/hosts >/tmp/out'
`
Output:
`sh: 1: /usr/bin/envsubst: not found
`

**Expected result:**
The Docker runtime image should include envsubst so the entrypoint can render config files and continue to Tomcat startup.

**Actual result:**
The container exits early because docker-entrypoint.sh calls /usr/bin/envsubst, which is not present in the runtime image.

**Proposed fix:**
Install gettext-base in the runtime stage of Dockerfile.

### Versions

* Branch: main
* Last commit id on main: bdac2d5

#### SW360 logs

Image

Image

a7med-3li:
Refs: #3916

> This PR fixes the Docker runtime image by installing `gettext-base`, which provides `envsubst` required by `scripts/docker-config/docker-entrypoint.sh`.
> * This PR belongs to issue #3916 and solves it by adding the missing runtime dependency used to render configuration templates before Tomcat startup.
> * It adds one OS package in the Docker runtime image: `gettext-base`.

Issue: #3916 

Closes #3916 

> Local validation performed:
> * Reviewed that `docker-entrypoint.sh` requires `/usr/bin/envsubst`
> * Verified the base Tomcat runtime image does not provide `envsubst`
> * Added `gettext-base` in the runtime stage so `envsubst` is available

### Checklist
Must:
- [x] If code is AI-generated, mention the tool and model used (e.g., GitHub Copilot, GPT-4)

AI assistance:
- GitHub Copilot

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -120,6 +120,12 @@ COPY --from=binaries /sw360_tomcat_webapps/*.jar ${CATALINA_HOME}/webapps/
 # Shared streamlined jar libs
 COPY --from=binaries /sw360_tomcat_webapps/libs/*.jar ${CATALINA_HOME}/lib/
 
+# Install gettext for envsubst
+RUN apt-get update -qq \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -qq -y --no-install-recommends \
+    gettext-base \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /app/sw360
 
 # Copy the configuration files


GMishx:
Fixed in https://github.com/eclipse-sw360/sw360/pull/3920, thus closing here.

Issue #3919, opened 2026-03-16
IndexOutOfBoundsException in VulnerabilityController pagination when requesting out-of-range page


Priyasharma620064:
### Description

The endpoint /api/vulnerabilities/trackingStatus/{projectId} throws an
IndexOutOfBoundsException when a client requests a page number that exceeds
the available dataset.

Instead of returning an empty paginated response, the server returns HTTP 500
Internal Server Error due to improper bounds validation in
VulnerabilityController.createPaginationMetadata().

#### How to reproduce

1. Identify a valid project ID (example: 753d0cc056d9d0da55fe7343a0000da8).

2. Send a GET request with an out-of-range pagination parameter:

GET /resource/api/vulnerabilities/trackingStatus/753d0cc056d9d0da55fe7343a0000da8?page=999&size=10

3. Observe the response.

#### Response

{
    "timestamp": "2026-03-15T19:22:24.360614855Z",
    "status": 500,
    "error": "Internal Server Error",
    "message": "fromIndex(19980) > toIndex(0)"
}

#### Impact

A user can repeatedly trigger HTTP 500 responses by supplying large page
numbers. This may lead to:

• Endpoint instability
• Degraded API reliability
• Violation of standard REST pagination behaviour

PDGGK:
I'd like to work on this.

Priya-Sharma25:
hey @PDGGK , I already fixed it.

PR #3921, opened 2026-03-16
fix(rest): prevent IndexOutOfBoundsException for out-of-range pagination

Priyasharma620064:
### Summary

This PR fixes an IndexOutOfBoundsException in the vulnerability tracking
pagination logic of `VulnerabilityController`.

Previously, when a client requested a page number that exceeded the available
dataset, the controller calculated a start index greater than the list size.
This resulted in an invalid `subList()` call and caused the endpoint to return
HTTP 500 Internal Server Error.

### Root Cause

Pagination metadata was created using:

    int start = pageNumber * pageSize;
    int end = Math.min(start + pageSize, listOfItems.size());
    listOfItems.subList(start, end);

If `start >= listOfItems.size()`, the call to `subList()` violated index bounds
requirements and triggered an `IndexOutOfBoundsException`.

Issue: #3919 

### Suggest Reviewer
@GMishx 

### How To Test?
1. Start the SW360 application.
2. Identify a project with 0 or more vulnerabilities.
3. Call the following API in Postman 
GET http://localhost:8080/resource/api/vulnerabilities/trackingStatus/{{projectId}}?page=999&size=10
4. Result: The server should return HTTP 400 Bad Request  instead of a 500 error.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vulnerability/VulnerabilityController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vulnerability/VulnerabilityController.java
@@ -569,6 +569,8 @@ public class VulnerabilityController implements RepresentationModelProcessor<Rep
             getSortedList(sortBy, sortOrder, vulTrackingStatusList);
             Map<String, Object> createPagination = createPaginationMetadata(pageable, vulTrackingStatusList);
             return ResponseEntity.ok(createPagination);
+        } catch (BadRequestClientException e) {
+            throw e;
         } catch (SW360Exception e) {
             throw new TException(e.why);
         } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vulnerability/VulnerabilityController.java
```diff
@@ -579,7 +579,7 @@ public ResponseEntity<Map<String, Object>> getVulnerabilitiesTrackingStatus(
     public Map<String, Object> createPaginationMetadata(Pageable pageable, List<Map<String, String>> listOfItems) {
         int pageSize = pageable.getPageSize();
         int pageNumber = pageable.getPageNumber();
-        int start = pageNumber * pageSize;
+        int start = Math.min(pageNumber * pageSize, listOfItems.size());
```
GMishx: I'd rather throw and exception here calling it as a bad request rather than silently fixing it. User knew the `totalElements` and thus should be aware of what they are asking in response.

Priya-Sharma25:
Done .

rudra-superrr:
LGTM.

PR #3923, opened 2026-03-16
Fix #3895: Improve exception handling in uploadLicense method

### LINKED ISSUE #3895: Improper Exception Handling in License Upload Stream Cleanup
### Description

When uploading a license file fails, the error message can be replaced by a stream cleanup error, hiding the real problem.

**What's wrong:**
- If license import fails AND stream cleanup also fails, you only see the cleanup error
- The actual reason for the import failure is lost
- No validation check if file is null or empty

### How to Reproduce

1. Upload an invalid license archive to `/api/licenses/upload`
2. If the import fails and a stream also fails to close
3. The logs only show "Failed to close stream" instead of the actual import error

### Location

**File:** `rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/Sw360LicenseService.java`  
**Method:** `uploadLicense()` (line ~273)

### The Problem

```java
} finally {
    // ... cleanup code ...
    if (closeFailure != null) {
        throw closeFailure;  // This hides the original exception!
    }
}

### PR DESCRIPTION
**Pull request by Bhawanishanker22**
### Description

Fixes #3895

This PR fixes issue #3895 by improving exception handling in the
`uploadLicense()` method of `Sw360LicenseService`.

Previously, if an exception occurred during the license import and
another exception occurred while closing the input stream, the original
exception could be lost. This made debugging difficult because only the
stream cleanup error was visible, hiding the actual cause of the failure.

This patch ensures that the original exception is preserved while still
recording any cleanup failures.

### Changes

1. **Input validation**
   - Added validation to ensure the uploaded file is not `null` or empty
     before processing.

2. **Preserve the primary exception**
   - Captured the original exception thrown during license import.
   - Ensured it is rethrown so the real failure remains visible.

3. **Use suppressed exceptions for cleanup failures**
   - If `InputStream.close()` fails after an import failure,
     the cleanup exception is attached using `Throwable.addSuppressed()`.
   - This preserves the original error while still reporting the cleanup issue.

4. **Improved error reporting**
   - If no prior exception occurred but stream closing fails,
     a clear `SW360Exception` is thrown indicating a cleanup failure.

### Why This Fix Is Correct

This behavior aligns with Java's `try-with-resources` design:

- The primary exception remains the main throwable.
- Any `close()` failures are added as suppressed exceptions.
- Logging frameworks such as SLF4J and Log4j automatically display suppressed exceptions in stack traces.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/Sw360LicenseService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/Sw360LicenseService.java
@@ -269,17 +269,28 @@ public class Sw360LicenseService {
         FileCopyUtils.copy(buffer, response.getOutputStream());
     }
 
-    public void uploadLicense(User sw360User, MultipartFile file, boolean overwriteIfExternalIdMatches, boolean overwriteIfIdMatchesEvenWithoutExternalIdMatch) throws IOException, TException {
-        final HashMap<String, InputStream> inputMap = new HashMap<>();
+    public void uploadLicense(User sw360User, MultipartFile file, boolean overwriteIfExternalIdMatches, boolean overwriteIfIdMatchesEvenWithoutExternalIdMatch)
+            throws IOException, TException {
+
+        if (file == null || file.isEmpty()) {
+            throw new BadRequestClientException("Unable to upload license file. File is null or empty.");
+        }
 
         if (!PermissionUtils.isUserAtLeast(UserGroup.ADMIN, sw360User)) {
-            throw new BadRequestClientException("Unable to upload license file. User is not admin");
+            throw new BadRequestClientException("Unable to upload license file. User is not admin.");
         }
+
+        final HashMap<String, InputStream> inputMap = new HashMap<>();
+        Throwable primaryThrowable = null;
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by melbiialy** [comment]
Bro i already fixed this in https://github.com/eclipse-sw360/sw360/pull/3896

**Comment by Bhawanishanker22** [comment]
Thanks for addressing this. I have created a PR for the same issue. My PR is focused on preserving the original exception, adding checks for null/empty files, and adding cleanup failures as suppressed exceptions.

File: rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/Sw360LicenseService.java [hunk 1]
```diff
@@ -270,18 +270,33 @@ private void copyDataStreamToResponse(HttpServletResponse response, ByteArrayInp
...
             licsImporter.importLics(sw360User, inputMap);
-        } finally {
+        }
+
+        catch (Throwable t) {
+            primaryThrowable = t;
+            throw t;
+        }
+
+        finally {
```
GMishx: Please fix coding style.


```suggestion
        } catch (Throwable t) {
            primaryThrowable = t;
            throw t;
        } finally {
```

keerthi-bl (approved):
Code looks good!

PR #3932, opened 2026-03-17
Fix duplicate ids in ThriftUploader class

### LINKED ISSUE #3931: Duplicate id handling in ThriftUploader.java
### Description

While looking through `ThriftUploader.java` in the `wsimport` module, I noticed an old `TODO: Improve duplicate handling` when mapping release relations.

Currently, when the uploader encounters duplicate `ReleaseRelation` entries for the same ID, it just logs an info message and silently picks the first one. This is a bit risky because it hides corrupted upload data from the user and can lead to data inconsistency in the database. 

*(Note: I noticed the `wsimport` module doesn't currently have a `src/test` infrastructure, so I plan to keep the upcoming PR scoped strictly to this logic fix to avoid altering the module's `pom.xml`.)*

#### How to reproduce

1. Upload a Thrift dataset containing duplicate release relations for the same project.
2. Observe that the system processes the upload successfully instead of rejecting it.
3. The duplicate is silently dropped, and the user remains completely unaware of the data issue.

N/A - Backend logic / data integrity issue.

### Versions

* Last commit id on main

#### SW360 logs

Currently, the system only outputs this in the console instead of failing the process:
`--- Duplicate key found!`
`--- 1: ...`
`--- 2: ...`

mast3RGG:
### Summary
This PR replaces the silent-fail Stream-based duplicate handling in `ThriftUploader.java` with a strict fail-fast approach. Instead of logging an info message and arbitrarily picking the first `ReleaseRelation` when duplicates exist, the system now uses an optimized single-pass loop to catch all duplicate IDs and immediately throws an `IllegalArgumentException`. This prevents silent data corruption and gives clear feedback to the user.

No new dependencies were added.

Issue: Resolves #3931 

1. Attempt to upload a Thrift dataset that contains duplicate release relations for the same project.
2. Verify that the upload process immediately fails and throws an `IllegalArgumentException` listing the exact duplicate IDs, rather than silently succeeding.

*Note regarding tests:* I intended to add a unit test for this specific fail-fast logic, but I noticed the `wsimport` module currently doesn't have a `src/test` directory and testing infrastructure in its `pom.xml`. I kept this PR strictly scoped to the logic fix to avoid introducing sweeping configuration changes to the module.

### Checklist
Must:

### DIFF
--- a/backend/wsimport/src/main/java/org/eclipse/sw360/wsimport/thrift/ThriftUploader.java
+++ b/backend/wsimport/src/main/java/org/eclipse/sw360/wsimport/thrift/ThriftUploader.java
@@ -96,21 +96,23 @@ public class ThriftUploader {
         Set<ReleaseRelation> releases = createReleases(wsProject, sw360User, tokenCredentials);
         sw360Project.setProjectResponsible(sw360User.getEmail());
 
-        /*
-         * TODO: Improve duplicate handling
-         */
-        Map<String, ProjectReleaseRelationship> releaseIdToUsage =
-                releases.stream()
-                    .collect(Collectors.toMap(
-                            ReleaseRelation::getReleaseId,
-                            ReleaseRelation::getProjectReleaseRelationship,
-                            (projectReleaseRelationship1, projectReleaseRelationship2) -> {
-                                LOGGER.info("--- Duplicate key found!");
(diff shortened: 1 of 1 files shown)


GMishx:
The current PR changes the business logic entirely and makes the API calls more restrictive. The `TODO:` asked for improvement, maybe in terms of speed, ease of understanding. Not by completely changing the business logic of duplicate merger.

PR #3935, opened 2026-03-18
perf(gson): reuse Gson objects

GMishx:

This PR performs following performance and bug fixes:
1. Following gson's user guide, share the stateless gson object, preventing memory usage.
2. On VMProcessor, implement servlet closer to listen and close all threads.
3. On Cloudant, reuse the client objects as internally it uses OkHttp and can perform connection pooling.
4. Allow option for user to enable retry with default 2 reties within 5 seconds interval.
5. Properly shutdown OkHttp to prevent leak warnings in tomcat (https://javadoc.io/static/com.squareup.okhttp3/okhttp/3.9.1/okhttp3/OkHttpClient.html)
6. Fix `ClearingTeam` to be a plain string instead of a user in ProjectController.


GMishx:
Tested by Actions CI.

GMishx:
TODO: Still most consumed memory (87%+) is used by `org.eclipse.sw360.datahandler.db.ProjectDatabaseHandler#cachedAllProjectsIdMap`

PR #3962, opened 2026-03-21
Fix duplicate id handling in ThriftUploader (#3931)

### LINKED ISSUE #3931: Duplicate id handling in ThriftUploader.java

PDGGK:
## Summary

Fixes duplicate release relation handling in `ThriftUploader.createProject()`.

Closes #3931

## Changes

Previously, duplicate `releaseId` values in upload data were silently dropped (first-wins via a merge function), hiding data corruption. Now:

- Removed the silent merge function from `Collectors.toMap()`
- Added `IllegalStateException` catch to detect duplicates
- Returns `ProjectImportResult(DUPLICATE_RELEASE_RELATION)` to fail the import with a clear error
- Added `DUPLICATE_RELEASE_RELATION` enum value to `ProjectImportError`
- Logs at WARN level with project name and details

### DIFF
--- a/backend/wsimport/src/main/java/org/eclipse/sw360/wsimport/thrift/ThriftUploader.java
+++ b/backend/wsimport/src/main/java/org/eclipse/sw360/wsimport/thrift/ThriftUploader.java
@@ -96,21 +96,17 @@ public class ThriftUploader {
         Set<ReleaseRelation> releases = createReleases(wsProject, sw360User, tokenCredentials);
         sw360Project.setProjectResponsible(sw360User.getEmail());
 
-        /*
-         * TODO: Improve duplicate handling
-         */
-        Map<String, ProjectReleaseRelationship> releaseIdToUsage =
-                releases.stream()
+        Map<String, ProjectReleaseRelationship> releaseIdToUsage;
+        try {
+            releaseIdToUsage = releases.stream()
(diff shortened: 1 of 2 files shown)


PDGGK:
Closing as duplicate of #3932. Apologies for the overlap.

PR #3975, opened 2026-03-23
Improve Docker README setup clarity for new contributors

srivastavauddeshya98-lang:
Added a prerequisites section and explanation of the `docker compose up` command in README_DOCKER.md.

These improvements help new contributors understand required tools and what happens when starting SW360 using Docker.

### DIFF
--- a/README_DOCKER.md
+++ b/README_DOCKER.md
@@ -13,6 +13,20 @@
 * [Networking](#networking)
 * [FOSSology Integration](#fossology-integration)
 
+## Prerequisites
+
+Before running SW360 using Docker, ensure the following tools are installed:
+
+- Docker (recommended version 20 or higher)
+- Docker Compose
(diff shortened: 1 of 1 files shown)


README_DOCKER.md
```diff
@@ -13,6 +13,20 @@
...
+
+- Docker (recommended version 20 or higher)
+- Docker Compose
+- Git
+
+Check installation:
+
+docker --version
+docker compose version
+git --version
```
GMishx, line 28: That is a no brainer. You would already have `git` to clone the repo, you should already have `docker` to start playing with docker.

srivastavauddeshya98-lang: Thank you for the clarification. I understand that these requirements are already implied for users working with Docker. I have removed the changes accordingly.

PR #3999, opened 2026-03-27
fix: null handling in CLIParser and CombinedCLIParser

### LINKED ISSUE #4000: fix: null handling in CLIParser and CombinedCLIParser
## Problem
The CLI XML parser layer has several latent null-pointer vulnerabilities 
that can cause silent data corruption or NPEs when processing malformed 
XML attachments.

### CLIParser.java
- If a `<Copyright>` node is missing its `<Content>` child, `null` is 
  silently inserted as a key into the HashMap, which can cause NPEs downstream
- `AssessmentSummary` missing/multiple element cases were collapsed into 
  a single `else` branch, making it impossible to distinguish between the 
  two scenarios in logs

### CombinedCLIParser.java
- If a copyright node is missing the `srcComponent` attribute, `externalId` 
  is null and `result.get(null)` throws an NPE
- Null `contentText` values are added to result Sets, polluting downstream data

## Fix
PR #3999 addresses all of the above with null guards, split conditionals, 
and appropriate log levels.

Rohan-Saxena644:
## Summary
Fixed latent null-pointer vulnerabilities in the CLI XML parser layer 
and improved diagnostic logging.

### CLIParser.java
- Guard against null `Content` element in copyright parsing — previously 
  a null key could be silently inserted into the HashMap
- Split `AssessmentSummary` missing/multiple element cases into separate 
  branches with `warn` instead of `error` log level
- Add Javadoc to public parser methods

### CombinedCLIParser.java
- Skip copyright entries where `srcComponent` (externalId) is null to 
  prevent NPE on map lookup
- Skip null `contentText` values to avoid polluting result Sets
- Add Javadoc to public parser methods

## Testing
Verified changes do not affect happy-path parsing behavior. Guards only 
activate on malformed XML input.

Closes #4000

### DIFF
--- a/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/CLIParser.java
+++ b/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/CLIParser.java
@@ -66,6 +66,10 @@ public class CLIParser extends AbstractCLIParser {
         super(attachmentConnector, attachmentContentProvider);
     }
 
+    /**
+     * Checks if this parser is applicable to the given attachment by verifying
+     * the file extension is XML and the root element is ComponentLicenseInformation.
+     */
     @Override
     public <T> boolean isApplicableTo(Attachment attachment, User user, T context) throws TException {
         AttachmentContent attachmentContent = attachmentContentProvider.getAttachmentContent(attachment);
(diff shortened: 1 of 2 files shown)


keerthi-bl (commented):
After careful review observation, Changes includes more on logs and the null check not much impacting on existing code as well. These changes are not required for now. Thanks for the contribution.

Rohan-Saxena644:
Thank you for the review and feedback! Understood, I'll keep this in mind for future contributions. Looking forward to contributing more meaningfully to the project!

PR #4018, opened 2026-04-02
fix(report): fix the file traversal path vulnerability.

rudra-superrr:
Issue: The report download endpoint GET /api/reports/download accepted a token parameter that was used directly as a file path without any validation. An unauthenticated attacker could supply an arbitrary path to read any file readable by the application process on the server.

How To Test:
-> Generate a report.
-> Test legitimate download  (http://localhost:8080/resource/api/reports/download?module=projects \
  &token=<user-email>/file/<timestamp>_<uuid>&user=<user-email>)                 # Expected: 200 OK, Excel file returned
-> Test path traversal attack (http://localhost:8080/resource/api/reports/download?module=projects \
  &token=../etc/passwd&user=<user-email>            Expected: 500 error, no file content returned

### DIFF
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
@@ -86,12 +86,19 @@ public class ExcelExporter<T, U extends ExporterHelper<T>> {
         final SXSSFWorkbook workbook = new SXSSFWorkbook();
         String token = UUID.randomUUID().toString();
         String filePath = TMP_EXPORTEDFILES + user.getEmail() + SLASH + "file" + SLASH;
-        File file;
+        String relativePath;
         try {
             File dir = new File(filePath);
-            dir.mkdirs();
-            file = new File(dir.getPath() + SLASH + SW360Utils.getCreatedOn() + "_" + token);
-            file.createNewFile();
+            if (!dir.mkdirs() && !dir.exists()) {
+                log.error("Failed to create export directory: {}", dir.getAbsolutePath());
+                throw new IOException("Failed to create export directory: " + dir.getAbsolutePath());
+            }
+            File file = new File(dir.getPath() + SLASH + SW360Utils.getCreatedOn() + "_" + token);
+            if (!file.createNewFile()) {
+                log.error("Failed to create export file: {}", file.getAbsolutePath());
+                throw new IOException("Failed to create export file: " + file.getAbsolutePath());
+            }
+            relativePath = user.getEmail() + SLASH + "file" + SLASH + file.getName();
             SXSSFSheet sheet = workbook.createSheet("Data");
 
             /** Adding styles to cells */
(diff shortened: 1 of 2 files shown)


libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
GMishx, line 120: Can we add error handling here in case the file/dir cannot be created to at least log it?

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
```diff
@@ -118,18 +119,28 @@ public String makeExcelExportForProject(List<T> documents, User user) throws IOE
...
     public InputStream downloadExcelSheet(String token) {
         InputStream stream = null;
         try {
-            File file = new File(token);
+            File file = new File(TMP_EXPORTEDFILES + token);
+            String canonicalPath = file.getCanonicalPath();
+            String allowedDir = new File(TMP_EXPORTEDFILES).getCanonicalPath();
+            if (!canonicalPath.startsWith(allowedDir + File.separator)) {
+                log.error("Path traversal attempt detected. Token: {}", token);
+                return null;
```
GMishx: Please throw a `FileNotFoundException` instead.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
```diff
@@ -118,18 +119,28 @@ public String makeExcelExportForProject(List<T> documents, User user) throws IOE
...
+            }
             if (file.exists()) {
-                stream = new FileInputStream(new File(token));
+                stream = new FileInputStream(file);
+            } else {
+                log.warn("Report file not found for token: {}", token);
```
GMishx: Here as well, please throw the exception as log already exists in the catch bellow.

rudra-superrr:
Can you please test this PR, @Shivamrut

Shivamrut:
> Can you please test this PR, @Shivamrut

Sure @rudra-superrr. Let me check this

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
rudra-superrr: done

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
rudra-superrr: done

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExcelExporter.java
rudra-superrr: done

Shivamrut:
# Test Results
image

Project data - GET `/resource/api/projects` shows an existing project (mspservice-build, id `95c9b63e94e1283febe43eec6d009256`), so report generation has real data to work with.

Normal report export - GET `/resource/api/reports?module=PROJECTS&withLinkedRelease=true&projectId=<id>` returns HTTP 200 and the saved body is a valid Excel file (`file` reports Microsoft OOXML). This checks that the usual export path still works.

Why step 3 does not reuse that same Excel file - The normal export streams the spreadsheet in the response; with default settings the server does not also write it under `/tmp/<user email>/file/` or return a token. The path layout under `/tmp/.../file/` is what the async mail export uses. So the file you save with `-o /tmp/direct-report.xlsx` is on your machine from the response body, not a server-side path the download API can open by token. Step 3 therefore uses a small file under `/tmp/<email>/file/` on the server so the token matches a real allowed path and the download endpoint can be exercised.

`/api/reports/download` with a valid token - With a file placed under `/tmp/<user email>/file/` (same layout the server uses for async exports), GET `.../reports/download?module=projects&token=<url-encoded path>&user=<email>` returns HTTP 200 and the response body matches that file. This exercises the download endpoint on a path under `/tmp/`.

Path traversal - GET `.../reports/download` with `token=../etc/passwd` returns HTTP 500 (error, no spreadsheet). The response does not contain a `root:` line from `/etc/passwd`, so arbitrary file read via the token is blocked.

Hi @rudra-superrr, please check if this was the correct approach and required testing. If anything else is required please let me know.
Hi @GMishx, I reviewed the changes: report download now validates the resolved path with `getCanonicalPath()` and only allows files under `/tmp/`, so the path-traversal / arbitrary file read we discussed looks fixed to me

GMishx:
Thank you @Shivamrut for finding this bug in the first place and help test it too. We are proceeding with this fix. Please feel free to comment if you have further points.

Happy coding :tada:

Shivamrut:
@GMishx would be good if I could mention this in GSOC proposal 😝
Please let me help with any other tests too. Happy to help!

PR #4020, opened 2026-04-03
fix(rest): enforce CLEARING_ADMIN role check on attachment checkStatus update

Aman-Cool:
While working on the clearing workflow, I noticed that any user with basic WRITE access could set a clearing report attachment's `checkStatus` to `ACCEPTED` via the REST API, without needing to be a `CLEARING_ADMIN` at all.

The problem is that `Sw360AttachmentService.updateAttachment()` applies whatever `checkStatus` value comes in from the request body with no role check. And since `autosetReleaseClearingState()` in the backend automatically promotes a release to `ClearingState.APPROVED` the moment its best clearing report attachment has `CheckStatus.ACCEPTED`, this means a release owner could effectively self-approve their own clearing just by PATCHing the attachment endpoint.

The fix is straightforward; added the same `CLEARING_ADMIN` guard that already exists in `Sw360ObligationService` right at the top of `updateAttachment()`. If the caller is trying to set `checkStatus` to anything other than `NOTCHECKED` and they don't have at least the `CLEARING_ADMIN` role, we throw an `AccessDeniedException` (HTTP 403) before anything gets mutated. This covers both the release and project attachment PATCH endpoints since they share the same service method.

### Suggest Reviewer
@GMishx 

### How To Test?

1. Create a release and upload an attachment of type `CLEARING_REPORT` as a regular user (not `CLEARING_ADMIN`).
2. Try to PATCH the attachment with `{"checkStatus": "ACCEPTED"}`, you should get a `403 Forbidden` response now.
3. Check that the release `clearingState` has NOT changed to `APPROVED`.
4. Repeat step 2 logged in as a `CLEARING_ADMIN` user, the update should go through and the release should move to `APPROVED` as expected.
5. Verify that setting `checkStatus` to `NOTCHECKED` still works for any WRITE user (resetting a check should remain unrestricted).

No new automated tests added in this PR, but the existing clearing workflow integration tests should still pass unchanged.

### Checklist
Must:

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -53,7 +53,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -59,7 +59,7 @@ jobs:
         # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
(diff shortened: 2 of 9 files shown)


Aman-Cool:
Hey @GMishx, just wanted to flag this one; any user with write access could set `checkStatus: ACCEPTED` on a clearing report via the REST API and the release would silently flip to `APPROVED` on its own. No `CLEARING_ADMIN` needed. Noticed it while digging through the clearing workflow. 
The fix is pretty small, just added the same role guard that's already used in `Sw360ObligationService`.

keerthi-bl:
This issue is handled in backend with - Moderation request and the same user who uploaded attachment can also have edit/update option. This PR not needed, closing this.

Aman-Cool:
Hey @keerthi-bl, I appreciate the review, but I don't think the moderation request flow actually covers this case and I want to walk through why before we close it for good.

The moderation path in `updateRelease()` is only triggered when `permissions.isActionAllowed(RequestedAction.WRITE)` returns false. But a release creator always has WRITE; their email is hardcoded into the moderators set via `toSingletonSet(document.createdBy)` in `ReleasePermissions`, which flows into contributors, which is what `isContributor()` checks for the WRITE case.

So the exact scenario here: a user creates a release, uploads a clearing report, then PATCHes `{"checkStatus": "ACCEPTED"}` on their own attachment; goes straight through the direct-write path. `autosetReleaseClearingState()` sees `ACCEPTED`, flips `clearingState` to `APPROVED`, and the DB gets written. No moderation request is ever created because they have WRITE and the `if` branch is taken, not the `else`.

The moderation flow would only kick in for a user who has no relationship to the release at all; not for the owner, which is the most realistic attacker here.

Happy to add a test that reproduces this if that helps make the case more concrete. Would love to hear your thoughts before we close this out...

Aman-Cool:
@GMishx, Can you please take a look🙏

Issue #4011, opened 2026-03-31
PATCH release/component/project returns 400 Bad Request on duplicate name/version instead of 409 Conflict


Aman-Cool:
**Description**

Ran into something weird with the PATCH release endpoint. When trying to update a release's version to one that already exists under the same component, the API returns 400 Bad Request instead of 409 Conflict. The backend is actually doing the right thing here; `ComponentDatabaseHandler` catches the duplicate and returns `RequestStatus.DUPLICATE`; but that status has no explicit handling in `Sw360ReleaseService.updateRelease()` so it falls through to the generic catch-all, becomes a plain `RuntimeException`, and `RestExceptionHandler` maps it to 400. Same issue on PATCH components and PATCH projects.

**How to reproduce**

Create two releases under the same component:

```
POST /api/releases
{"name": "OpenSSL", "version": "1.0.0", "componentId": "<cid>"}

POST /api/releases
{"name": "OpenSSL", "version": "2.0.0", "componentId": "<cid>"}
```

Then patch the second one to use the same version as the first:

```
PATCH /api/releases/<second-release-id>
Content-Type: application/json

{"version": "1.0.0"}
```

What comes back:
```json
HTTP/1.1 400 Bad Request
{"status": 400, "error": "Bad Request", "message": "sw360 release with name 'OpenSSL 1.0.0 cannot be updated."}
```

What should come back:
```
HTTP/1.1 409 Conflict
```

Looking at `Sw360ReleaseService.updateRelease()` around line 322, there's explicit handling for `INVALID_INPUT`, `NAMINGERROR`, `DUPLICATE_ATTACHMENT` etc. but `DUPLICATE` is just missing from that list. So it drops into the final `else if (requestStatus != SUCCESS && requestStatus != SENT_TO_MODERATOR)` block, throws a generic `RuntimeException`, and `RestExceptionHandler.handleRuntimeException()` turns that into a 400. The fix is straightforward — just needs an explicit branch for `DUPLICATE` that throws something mapping to 409. `Sw360ComponentService` and `Sw360ProjectService` have the exact same gap.

The annoying part is that `ComponentDatabaseHandler` is handling this correctly; the duplicate detection works, `RequestStatus.DUPLICATE` is returned as it should be; the information just gets lost on the way back up through the service layer.

**Screenshots (if applicable)**

N/A, the steps above reproduce it reliably.

**Versions**

- Reproduced on current main, last commit 014c574

**SW360 logs**

[...]

PR #4027, opened 2026-04-03
fix(rest): return 409 Conflict instead of 400 when PATCH creates a duplicate

Aman-Cool:
When patching a release, component, or project with a name/version that already exists, the API was returning 400 Bad Request instead of 409 Conflict. The backend was correctly detecting the duplicate and returning `RequestStatus.DUPLICATE` but that status had no explicit handling in the service layer; it was just falling into the generic catch-all and becoming a plain `RuntimeException`, which `RestExceptionHandler` blindly maps to 400. Added an explicit branch for `RequestStatus.DUPLICATE` in all three service classes that throws `HttpClientErrorException(HttpStatus.CONFLICT)` instead.

No new dependencies added.

Issue: #4011 

### Suggest Reviewer

@GMishx 

### How To Test?

Spin up SW360, create two releases under the same component with different versions, then PATCH the second one to use the same version as the first. Before this fix that would come back as 400 Bad Request with a confusing "cannot be updated" message. After this fix it should return 409 Conflict. Same test applies for components and projects.

No additional tests were added, if there's a preferred test pattern for the REST service layer here happy to add coverage.

### Checklist

Must:


Aman-Cool:
@GMishx, Three files, three lines; `RequestStatus.DUPLICATE` was just missing from the if-else chain in each service class so it kept falling into the generic catch-all and coming out as 400. 
Added an explicit branch for it that throws a 409 instead.

GMishx (approved):
Approved by @keerthi-bl

GMishx:
@deo002 @amritkv frontend error handling might need to be updated.

PR #4028, opened 2026-04-04
Fixed typo in Sw360AuthorizationServerConfiguration

Shivamrut:
### Summary

Corrects a method name typo in `Sw360AuthorizationServerConfiguration` (`appSecurtiy` → `appSecurity`) so it matches the intended meaning and aligns with `SecurityConfig.appSecurity` in the authorization-server module. Behavior is unchanged; only the `@Bean` method name (and thus the default Spring bean name) is updated.

### Issue

- **Problem:** Test security config used a misspelled method name `appSecurtiy` instead of `appSecurity`.
- **Impact:** None on runtime behavior; readability and consistency only.


keerthi-bl (approved):
Looks Good!

PR #4030, opened 2026-04-04
fix(project): fix NPE and attachment client leak in saveAttachmentUsages

Aman-Cool:
Was going through the project attachment usage flow and noticed a few things that felt off, so fixed them together since they're all in the same method chain.

- No specific issue tracked yet, but this silently breaks license-info generation for projects with sub-project hierarchies when a Thrift/Cloudant error occurs mid-loop, and the NPE in `buildProjectPaths` can surface when project objects come in partially constructed (e.g. from a moderation approval path). 
- No new dependencies.

Issue: - NA

### Suggest Reviewer
@GMishx 

### How To Test?

1. Create a project with at least 2–3 linked sub-projects, each having releases with source attachments
2. Trigger a license-info generation — attachment usages should propagate correctly from sub-projects to the parent
3. To hit the old leak path: add a breakpoint or log in `makeAttachmentClient()` and count how many times it gets called; should now be once total, not once per sub-project
4. For the NPE: pass a project object with a null id into `buildProjectPaths`, old code throws, new code returns cleanly
5. For the `RuntimeException` catch: simulate a Cloudant connection failure during `saveAttachmentUsages`, old code would let it bubble as a 500, new code logs it and continues

No new automated tests added; the affected methods are `private` and the setup for integration testing this path is fairly heavy. 

### Checklist
Must:

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -521,24 +521,23 @@ public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
     }
 
     private void saveAttachmentUsages(Project project) {
-        AttachmentService.Iface attachmentClient = thriftClients.makeAttachmentClient();
         String projectId = project.getId();
         List<String> projectPaths = new ArrayList<>();
-
-        buildProjectPaths(project, null, projectPaths, new HashSet<>());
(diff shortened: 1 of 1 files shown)


Aman-Cool:
Hey @GMishx, was tracing through the license-info generation flow and stumbled on this; the client being recreated every iteration felt wrong so looked closer and found the null id thing too. Both were quiet enough that they probably wouldn't show up until someone hit a bigger project hierarchy or a flaky Thrift connection. 
Figured I'd just fix it in one go while I was there.

Issue #4033, opened 2026-04-05
Multiple Null Safety and Missing Security Annotation Issues in ProjectController


Priyasharma620064:
### Description

1. NPE in icenseClearing : A NullPointerException occurs when calling the license clearing endpoint for a project with no linked releases.
2. Authorization Gap in deleteProject : The delete project endpoint is missing the @PreAuthorize annotation, allowing unauthorized deletions.
3. Uninitialized Response Status in linkToProjects : Passing an empty list to the link projects endpoint results in a server error because the HTTP response status was not explicitly assigned for empty inputs.

#### How to reproduce

Bug 1: NPE in licenseClearing
1. Create a project with no releases linked.
2. Call GET /api/projects/{id}/licenseClearing?transitive=false&releaseRelation=CONTAINED.
3. Observe 400 Bad Request with Cannot invoke "java.util.Map.entrySet()" because ... getReleaseIdToUsage() is null.

Bug 2: Missing @PreAuthorize
1. Use a token with READ-only permissions.
2. Call DELETE /api/projects/{id}.
3. Observe the request reaches the service layer instead of being blocked by Spring Security at the controller level.

Bug 3: linkToProjects Status Null
1. Call POST /api/projects/{id}/linkProjects with an empty body [].
2. Observe 400 Bad Request due to an unassigned HTTP response status for the processed input.

#### Screenshots 

<img width="1396" height="688" alt="Image" src="https://github.com/user-attachments/assets/dc566f5a-db81-4ebe-b31d-47f0e40fa565" />

<img width="1454" height="767" alt="Image" src="https://github.com/user-attachments/assets/794f1b45-a27d-48f0-8373-28919b22cbb6" />

Anushreebasics:
@Priyasharma620064 is this issue still open? because I see a merged PR

Priyasharma620064:
@Anushreebasics This issue has been completed and the corresponding PR is already merged.

PR #4035, opened 2026-04-05
fix(rest): fix NPEs and missing security annotation in ProjectController

Priyasharma620064:
### Summary
**1. Stopping an unexpected crash:** I fixed an issue where the "License Clearing" page would crash if you opened it for a project that didn't have any releases yet. I've added a simple safety check to make sure the app handles these empty projects gracefully.
**2. Plugging a security gap:** I noticed that the "Delete Project" function wasn't properly checking for user permissions like other parts of the app do. I've now added a security check to ensure only users with the right authority can actually delete projects.
**3. Fixing a server response error:** When trying to link projects but using an empty list, the server would get confused and throw an error because it didn't know what status code to return. I've updated the logic to always return a clear response, even when there's nothing to link.

Issue:  #4033 

### Suggest Reviewer
@GMishx 

### How To Test?

**1. For the License Clearing fix:** Just try to fetch the license clearing info for any project that doesn't have any releases yet. It used to crash with a 400 error, but now it should correctly return a clean 200 OK with an empty list.
**2. For the Missing Security Annotation fix:** Try deleting a project using a user account that only has Read permissions. The request should now be immediately blocked with a 403 Forbidden, ensuring better security.
**3. For the Project Linking fix:** Try to 'link' projects but send an empty list [] in the request body. Previously, this triggered a 400 error about missing status codes, but now it handles it and returns 204 No Content.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -481,7 +481,7 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
         Project sw360Project = projectService.getProjectForUserById(id, sw360User);
 
         //check the below condition when releaseRelation is not null
-        if (releaseRelation != null) {
+        if (releaseRelation != null && sw360Project.getReleaseIdToUsage() != null) {
             Map<String, ProjectReleaseRelationship> filteredReleaseIdToUsage = sw360Project.getReleaseIdToUsage().entrySet().stream()
                     .filter(entry -> entry.getValue().getReleaseRelation() == releaseRelation)
                     .collect(Collectors.toMap(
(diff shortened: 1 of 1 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -961,7 +964,7 @@ public ResponseEntity linkToProjects(
         User sw360User = restControllerHelper.getSw360UserFromAuthentication();
         Project sourceProj = projectService.getProjectForUserById(id, sw360User);
         Map<String, String> responseMap = new HashMap<>();
```
rudra-superrr, line 966: If projectsId in request body is empty, no projects were linked, returning 200OK will be misleading. The correct fix would be to return 204 No Content for an empty input or 400 Bad Request for an invalid empty request.

Priya-Sharma25:
Thank you for the review, @rudra-superrr!
I’ve updated the implementation to handle empty projectIds in the request body.
It now returns 204 No Content when the list is empty (no projects linked), and avoids returning a misleading 200 OK.
Please take another look.

rudra-superrr:
LGTM

rudra-superrr (approved):

PR #4037, opened 2026-04-06
fix(fossology): rollback clearingState when upload preparation fails

Aman-Cool:
When triggering a FOSSology upload, the handler marks the release as `SENT_TO_CLEARING_TOOL` in CouchDB before it actually tries to fetch the attachment and upload it. If anything goes wrong during that prep work (missing attachment content, transient DB error, etc.), the exception just bubbles up and the clearing state never gets reset; leaving the release permanently stuck with no way to re-trigger from the UI.

This fix wraps the risky part in a try-catch so that on any failure, we reset the step back to `NEW` and write `NEW_CLEARING` back to the document before re-throwing, the same cleanup that already happens on a normal upload failure response.

Issue: NA

### Suggest Reviewer
@GMishx 

### How To Test?

1. Create a release with a source attachment, then delete the attachment content document directly from CouchDB (simulating a stale reference or failed upload).
2. Trigger FOSSology from the release page.
3. **Before fix:** the release gets stuck at `SENT_TO_CLEARING_TOOL` even though nothing was sent. Re-triggering is blocked.
4. **After fix:** the release returns to `NEW_CLEARING` and you can retry once the attachment issue is resolved.

No new tests added; the existing FOSSology handler tests cover the upload step; manual reproduction is straightforward with a CouchDB document edit.

### Checklist
Must:

### DIFF
--- a/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
+++ b/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
@@ -262,14 +262,22 @@ public class FossologyHandler implements FossologyService.Iface {
                     furthestStep.setStepStatus(ExternalToolProcessStatus.NEW);
                     fossologyProcess.setProcessStatus(ExternalToolProcessStatus.NEW);
                     furthestStep.setResult("Upload preparation failed: " + e.getMessage());
-                    updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
+                    try {
+                        updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
+                    } catch (Exception rollbackEx) {
+                        log.error("Failed to roll back clearing state after upload failure: {}", rollbackEx.getMessage());
+                    }
                     throw e;
                 } catch (RuntimeException e) {
                     log.error("Upload preparation failed for release, rolling back clearing state: {}", e.getMessage());
                     furthestStep.setStepStatus(ExternalToolProcessStatus.NEW);
                     fossologyProcess.setProcessStatus(ExternalToolProcessStatus.NEW);
                     furthestStep.setResult("Upload preparation failed: " + e.getMessage());
-                    updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
+                    try {
+                        updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
+                    } catch (Exception rollbackEx) {
+                        log.error("Failed to roll back clearing state after upload failure: {}", rollbackEx.getMessage());
+                    }
                     throw e;
                 }
                 break;


Aman-Cool:
Hey @GMishx, found a bug where a FOSSology trigger failure leaves the release permanently stuck at `SENT_TO_CLEARING_TOOL` with no way to retry from the UI. Fix just adds a rollback for the cases that were slipping through. Pretty small change, let me know if anything looks off!

backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
```diff
@@ -201,60 +201,76 @@ private void handleUploadStepV2(Iface componentClient, Release release, User use
...
+                            furthestStep.setResult(response != null ? response.getMessage() : "Upload failed");
+                            log.error("UPLOAD FAILED: {}", response != null ? response.getMessage() : "Unknown error");
+                        }
                     }
+                } catch (TException e) {
+                    log.error("Upload preparation failed for release, rolling back clearing state: {}", e.getMessage());
+                    furthestStep.setStepStatus(ExternalToolProcessStatus.NEW);
+                    fossologyProcess.setProcessStatus(ExternalToolProcessStatus.NEW);
+                    furthestStep.setResult("Upload preparation failed: " + e.getMessage());
+                    updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
```
Sandipmandal25: noticed that `updateFossologyProcessInRelease` itself declares `throws TException` so if couchdb is already down the rollback call can fail too. in that case the new exception would replace `e` `throw e` is never reached and the original failure gets lost while the release may still remain stuck at `SENT_TO_CLEARING_TOOL`. would it make sense to guard the rollback call itself so the original exception always propagates? What do you think @GMishx

backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
```diff
@@ -201,60 +201,76 @@ private void handleUploadStepV2(Iface componentClient, Release release, User use
...
+                    throw e;
+                } catch (RuntimeException e) {
+                    log.error("Upload preparation failed for release, rolling back clearing state: {}", e.getMessage());
+                    furthestStep.setStepStatus(ExternalToolProcessStatus.NEW);
+                    fossologyProcess.setProcessStatus(ExternalToolProcessStatus.NEW);
+                    furthestStep.setResult("Upload preparation failed: " + e.getMessage());
+                    updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
```
Sandipmandal25: same issue i guess if `updateFossologyProcessInRelease` throws here the rollback exception replaces the original one and `throw e` is never reached. should this be guarded the same way?

Aman-Cool:
@Sandipmandal25, Wrapped the rollback call in its own try-catch so if CouchDB is also down, the rollback failure gets logged but swallowed and the original exception always propagates cleanly.

GMishx:
The fixes have been merged with https://github.com/eclipse-sw360/sw360/pull/4057

Thank you for your contributions @Aman-Cool

PR #4048, opened 2026-04-07
chore(rest): minor security and test configuration cleanup

Shivamrut:

While reviewing the REST security setup, I noticed a few small inconsistencies and deprecation-related items. Rather than opening several one- or two-line PRs, I grouped the changes here; each concern is in its **own commit** so they can be reviewed or reverted independently. If maintainers prefer **separate PRs** per topic, say so and I will split them.

**Summary of changes**

1. **Resource server (`ResourceServerConfiguration`)**  
   Remove deprecated Spring Security `xssProtection()` configuration and rely on the existing **Content-Security-Policy** (`script-src 'self'`). Remove an unused Log4j logger field and related imports.

2. **Authorization server**  
   Use the shared **`PasswordEncoder` bean** from `SecurityConfig` in `Sw360UserAuthenticationProvider` and `OAuthClientController` instead of instantiating an extra `BCryptPasswordEncoder`.

> After wiring that in, integration tests failed with **`BeanCurrentlyInCreationException`** on `passwordEncoder` (circular dependency): `SecurityConfig` injects `Sw360UserAuthenticationProvider`, while the provider injects `PasswordEncoder`, and that bean was previously declared as `@Bean` on the **same** `SecurityConfig` class, so Spring could not finish creating `passwordEncoder` while `SecurityConfig` was still being built. That cycle did not exist when the encoder was constructed locally.
 **Fix:** define `PasswordEncoder` in a dedicated configuration class (e.g. `PasswordEncoderConfiguration`) so the bean can be created independently of `SecurityConfig`, which keeps injecting the authentication provider as before.

3. **Tests (`Sw360WebSecurityConfiguration`, SECURITY_MOCK)**  
   Remove unused imports (including `NoOpPasswordEncoder`).

**Dependencies:** None.

### Checklist

Must:


Shivamrut:
Hi @GMishx, A lot of contributors naturally focus on new features. I’ve been looking at the current codebase and chipping away at inconsistencies, deprecations, and hygiene where I can. Both matter for a healthy project. If maintainers prefer that I prioritize feature implementation, I can pivot. Please let me know what fits best.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/security/ResourceServerConfiguration.java
```diff
@@ -94,7 +89,7 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http, Authentication
                 })
                 .httpBasic(Customizer.withDefaults())
                 .exceptionHandling(x -> x.authenticationEntryPoint(saep))
-                .headers(headers -> headers.xssProtection(xXssConfig -> xXssConfig.headerValue(XXssProtectionHeaderWriter.HeaderValue.DISABLED))
+                .headers(headers -> headers
```
GMishx: We need the `xss` filtering. Please revert the changes.

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/security/ResourceServerConfiguration.java
Shivamrut: Hi @GMishx, sorry I missed your comment.  I have reverted xss change. Can you please brief me why it is still required?

GMishx: Simply because we do not want XSS vulnerability in SW360.

GMishx (approved):
Good to go.

Issue #4050, opened 2026-04-08
Add REST endpoint for FOSSology release status


Sandipmandal25:
### Description

SW360 has a `GET /releases/{id}/checkFossologyProcessStatus` endpoint but it returns a generic response with a raw `ExternalToolProcess` object and a simple SUCCESS/FAILURE/PROCESSING status.

It does not clearly show the different FOSSology steps like upload, scan, and report. Because of this, clients have to manually understand the state.

A new endpoint is needed that returns:
- uploadStatus, scanStatus, reportStatus  
- uploadId from the upload step  
- sourceAttachmentId (the file that was scanned)  
- reportAttachmentId (the generated report in SW360)  
- availableReportFormats (after scan is done)  

#### Steps to reproduce

1. Trigger FOSSology processing for a release.  
2. Call `GET /releases/{id}/checkFossologyProcessStatus`.  
3. See that the response does not clearly show each step.

Expected result:
A clear and structured response showing each FOSSology step.

GMishx:
Fixed in https://github.com/eclipse-sw360/sw360/pull/4051

PR #4051, opened 2026-04-08
feat(resourceserver): add FOSSology status endpoint for release

Sandipmandal25:
Issue: #4050 

Adds a new endpoint `GET /releases/{id}/fossology` to return FOSSology processing info for a release. The existing `checkFossologyProcessStatus` endpoint returns a generic response and does not clearly show each step. This endpoint provides a clear view of the workflow with upload, scan and report status.

The data is read from `ExternalToolProcess` already stored in SW360. No live FOSSology API calls are made.

### How To Test?

1. Trigger FOSSology processing for a release  
2. Call `GET /releases/{id}/fossology`  
3. Check that the response includes uploadStatus, scanStatus, reportStatus,  
   uploadId, sourceAttachmentId and availableReportFormats (after scan is done)

No new tests added since this only reads existing data.

### Suggest Reviewer

@GMishx 
@deo002 

### Checklist

Must:

* [x] All related issues are referenced in commit messages and in PR format


Sandipmandal25:
Keeping this in draft for now. will do a final pass on dto usage and edge cases before marking it ready.

GMishx (approved):
Tested, working as expected.

PR #4065, opened 2026-04-10
fix(fossology): reset IN_WORK state on upload exception to prevent stuck clearingState

Aman-Cool:
When sending a release to FOSSology, the handler writes `IN_WORK` to CouchDB before the upload attempt. If anything throws during the attachment fetch or network call, that state never gets cleaned up, leaving the release permanently stuck at `SENT_TO_CLEARING_TOOL` with no way to retry without admin intervention.

- Replaced the outer `try/catch` with targeted `try/catch` blocks inside `handleUploadStepV2` and `handleScanStepV2` that call the existing `bailUploadFailed()` helper (upload failures) or inline state reset + `updateFossologyProcessInRelease` (scan failures, to avoid incorrectly marking the process `OUTDATED`).
- Non-exception scan failure path (`startScanning` returns ≤ -1) now also persists the reset step status back to CouchDB.
- Guards `Integer.parseInt` in the `IN_WORK` scan case against null/non-numeric job IDs left by older bugs.
- Added a unit test covering the `getAttachmentStream` -> `TException` path: asserts the upload step is reset to `NEW` and the state is written back before rethrowing.
- No new dependencies.

Issue: NA

### Suggest Reviewer
@GMishx

### How To Test?

1. Attach a source file to a release and configure FOSSology.
2. While clicking "Send to FOSSology", briefly make the CouchDB attachments DB unreachable (or revoke read access).
3. **Before fix:** release stays at `SENT_TO_CLEARING_TOOL`, every retry silently does nothing.
4. **After fix:** release resets to `NEW_CLEARING`, retry works normally.

### Checklist
Must:

### DIFF
--- a/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
+++ b/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
@@ -35,6 +35,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import javax.annotation.Nonnull;
+
 import java.io.InputStream;
 import java.time.Instant;
 import java.time.LocalDateTime;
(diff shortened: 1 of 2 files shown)


Aman-Cool:
Hey @GMishx, Found this while digging into the FOSSology flow; if the attachment fetch or any I/O throws between the pre-commit `IN_WORK` write and the actual upload, there's no cleanup path, so the release just silently hangs there forever. Pretty easy to hit during a CouchDB hiccup or rolling restart. The fix is just a `try/catch/finally` to make sure the state always gets corrected before we rethrow. Happy to add a test or tweak anything if needed!

backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
```diff
@@ -169,19 +169,41 @@ public ExternalToolProcess process(String releaseId, User user, String uploadDes
...
+            // call but the call never completed, reset it to NEW so subsequent retries
+            // can attempt the upload again rather than hitting the no-op IN_WORK guard.
+            if (FossologyUtils.FOSSOLOGY_STEP_NAME_UPLOAD.equals(furthestStep.getStepName())
+                    && ExternalToolProcessStatus.IN_WORK.equals(furthestStep.getStepStatus())) {
+                furthestStep.setStepStatus(ExternalToolProcessStatus.NEW);
+                fossologyProcess.setProcessStatus(ExternalToolProcessStatus.NEW);
+                furthestStep.setResult("Upload interrupted: " + e.getMessage());
+                log.error("Exception during FOSSology upload for release [{}], resetting step to NEW: {}",
+                        releaseId, e.getMessage());
+            }
```
GMishx: There is now a helper function to take care of all the state reset called `bailUploadFailed()`. Please use that instead.

https://github.com/eclipse-sw360/sw360/blob/1701644b970b7120c71e21e4bf4285ea6c8b6808/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java#L306-L321

GMishx (changes requested):

Aman-Cool:
Hey @GMishx, thanks for pointing that out; missed that `bailUploadFailed()` was already there.
Swapped out the outer try-catch for targeted catches inside `handleUploadStepV2()` and `handleScanStepV2()` that delegate to it. Also noticed the old code was setting `processStatus` to `NEW` on failure, `bailUploadFailed()` uses `OUTDATED` which is more correct, so that's fixed too.

rudra-superrr:
The current implementation already contains the core fix that PR is trying to add:

- During the upload step, the process is first marked as `IN_WORK` inside `handleUploadStepV2()`.

- If fetching the attachment metadata fails, `handleUploadStepV2()` catches that failure and routes it through `bailUploadFailed()`.
 
- If fetching the attachment stream or performing the upload fails, `handleUploadStepV2()` also catches that and sends it through `bailUploadFailed()`.
 
- `bailUploadFailed()` marks the process as `OUTDATED`, and `calculateCurrentClearingState()` converts that back to `NEW_CLEARING`, so the release does not remain stuck in `SENT_TO_CLEARING_TOOL`.

rudra-superrr:
Closing the PR.

PR #4072, opened 2026-04-13
perf(licenseinfo): use value key for attachment cache

### LINKED ISSUE #3852: perf(licenseinfo): Cache lookup iterates all entries causing O(n) performance
### Description

The `LicenseInfoHandler.getLicenseInfoForAttachment()` method performs O(n) cache lookups by iterating through all cache entries instead of using O(1) direct hash-based retrieval, causing significant performance degradation as the cache size grows.

#### How to reproduce

1. Use SW360 with multiple license info requests for different attachments
2. As cache grows to 50-100+ entries, license info generation becomes noticeably slower
3. Profile or add timing logs around `getLicenseInfoForAttachment()` calls
4. Observe that cache lookup time increases linearly with cache size instead of remaining constant

**Root Cause:**
- File: `backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/LicenseInfoHandler.java`
- Lines: 578-587, 621-625
- The cache uses `Object[]` as key type, which prevents direct `cache.get()` from working due to array reference equality
- Code works around this by manually iterating all cache entries to compare key contents

**Current Code (Lines 578-587):**
```java
if (licenseInfoCache != null) {
    for (Entry<Object[], List<LicenseInfoParsingResult>> entry : licenseInfoCache.asMap().entrySet()) {
        Object[] key = entry.getKey();
        List<LicenseInfoParsingResult> cachedValue = entry.getValue();
        if (attachmentContentId.equals(key[0].toString()) && includeConcludedLicense == (boolean) key[1]
                && cachedValue != null) {
            return cachedValue;
        }
    }
}

### PR DESCRIPTION
**Pull request by officialasishkumar**

Replaces the `Object[]` license info cache key with a value-based key so attachment license info lookups can use direct Guava cache access instead of scanning all cache entries. Also switches the evaluation cache lookup to `getIfPresent()`.

No new dependencies were added.

Issue: Fixes #3852

Run:

```bash
mvn -pl backend/licenseinfo -Dtest=LicenseInfoHandlerTest test \
    -Dbase.deploy.dir=$TOMCAT_HOME \
    -Dhelp-docs=false
```

Implemented a unit test proving equal cache key values resolve the same cached result without relying on array reference equality.

### Checklist
Must:

### DIFF
--- a/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/LicenseInfoHandler.java
+++ b/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/LicenseInfoHandler.java
@@ -75,11 +75,40 @@ public class LicenseInfoHandler implements LicenseInfoService.Iface {
     protected List<OutputGenerator<?>> outputGenerators;
     protected ComponentDatabaseHandler componentDatabaseHandler;
     protected ProjectDatabaseHandler projectDatabaseHandler;
-    protected Cache<Object[], List<LicenseInfoParsingResult>> licenseInfoCache;
+    protected Cache<LicenseInfoCacheKey, List<LicenseInfoParsingResult>> licenseInfoCache;
     protected Cache<String, LicenseInfoParsingResult> licenseInfoCacheForEvaluation;
     protected Cache<String, List<ObligationParsingResult>> obligationCache;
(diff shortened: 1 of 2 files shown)


GMishx:
Already solved by https://github.com/eclipse-sw360/sw360/pull/3853

Thank you for your contribution.

PR #4082, opened 2026-04-15
fix(rest): add detailed exception handling for patchClearingRequest

Shivamrut:
### Summary

`patchClearingRequest` used a single `catch (Exception)` and always rethrew a **`BadRequestClientException`**, which the REST layer maps to **HTTP 400**. That meant real **403 Forbidden**, **404 Not Found**, and **500 Internal Server Error** situations were all reported as **400 Bad Request**.

### Why that matters

HTTP status codes tell clients and monitoring tools what went wrong (permission vs missing resource vs server fault). Turning everything into **400** hides those distinctions and makes retries, auth handling, and alerts harder to get right.

### What I did instead

The broad catch was replaced with **narrow catches** that only map true client/validation problems to **`BadRequestClientException`**, and **let the right exceptions propagate** so `RestExceptionHandler` can return **403**, **404**, or **500** as intended.

### Pattern I followed

This matches the approach already used on **`addCommentToClearingRequest`** in the same controller: handle **`IllegalArgumentException`** → **`BadRequestClientException`**, normalize **`ResourceNotFoundException`**, **rethrow `SW360Exception`**, and wrap remaining **`TException`** in a generic **`SW360Exception`** message, so status handling stays consistent with that endpoint.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
@@ -377,7 +377,7 @@ public class ClearingRequestController implements RepresentationModelProcessor<R
                     schema = @Schema(implementation = ClearingRequest.class))
             @RequestBody Map<String, Object> reqBodyMap,
             HttpServletRequest request
-    ) {
+    ) throws SW360Exception {
         try{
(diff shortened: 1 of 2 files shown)


Shivamrut:
Hi @GMishx, If this issue looks right, I have identified few other places where this can be fixed. I plan to raise a PR that fixes it in bulk. Please let me know if I should proceed with it.

GMishx (approved):
LGTM

PR #4086, opened 2026-04-16
fix(auth): improve user validation messages for authentication failures

aaryan359:
## Summary
Fix misleading authentication errors by separating user-not-found and deactivated-user checks.

Authentication currently returns a deactivated-user style message even when the SW360 user does not exist, which is confusing for integrations.

### Change
- Split null-user and deactivated-user validation in API token authentication flow.
- Split null-user and deactivated-user validation in JWT authentication flow.
- Return accurate exception messages for each case.

## Result
- Non-existent users now get a proper invalid-credentials style message.
- Deactivated users get a deactivated-user message.
- Error handling is clearer and less misleading.
@deo002 sir , @GMishx sir


GMishx:
That is not a misleading message, it is a secure design.

PR #4096, opened 2026-04-17
fix(rest): Preserve 404/403 and others when mapping controller exceptions

Shivamrut:
## Summary

REST controllers and a few services used broad `catch (Exception e)` blocks that **rewrapped or replaced** almost any failure with `SW360Exception`, `TException`, `RuntimeException`, or a **generic “not found”** message. That **hid real HTTP semantics**: failures that the backend already expressed as **404 Not Found**, **403 Forbidden**, or **400 Bad Request** were often turned into **500 Internal Server Error** or the wrong client error.

This change **rethrows typed Spring and Thrift exceptions** (`ResourceNotFoundException`, `AccessDeniedException`, `BadRequestClientException`, and where relevant `SW360Exception`, `TException`, `DataIntegrityViolationException`, `IOException`, `RuntimeException`) **before** any generic handler, so `RestExceptionHandler` (and existing multi-status logic) can return the **correct status and body**.

**Dependencies:** None.

### `FossologyAdminController`
- **Why:** `saveConfig` can surface not-found or access-denied conditions from the stack. Those must not be wrapped into `SW360Exception` (typically mapped to **500**).
- **Change:** Catch `ResourceNotFoundException` and `AccessDeniedException` and rethrow so the API returns **404** and **403** as intended.

### `ClearingRequestController` (two methods)
- **Why:** Listing clearing requests and comments can fail with 404/403 from services; wrapping as `SW360Exception` loses that.
- **Change:** Explicit rethrow for `ResourceNotFoundException` and `AccessDeniedException` before the generic `SW360Exception` wrap.

### `ComponentController`
- **`validateAndGetComponent`**
  - **Why:** The previous `catch (Exception e)` turned **every** failure (including wrong type, Thrift, I/O, etc.) into `ResourceNotFoundException` with a fixed “component not found” message, which was **incorrect and misleading**.
  - **Change:** Call `getComponentForUserById` without that blanket catch; treat **null** as not found only when that is the real contract.
- **`importSBOM` / `prepareImportSBOM`**
  - **Why:** Broad `catch (Exception)` to `RuntimeException` masked **404/403**, domain `SW360Exception`, and `TException`.
  - **Change:** Rethrow `ResourceNotFoundException`, `AccessDeniedException`, `SW360Exception`, and `TException`; only unknown errors stay in the generic path.

### `DepartmentController`
- **Why:** Path update errors can be 404/403; they must not all become a generic `SW360Exception` with a prefixed message.
- **Change:** Rethrow `ResourceNotFoundException` and `AccessDeniedException` before wrapping other failures.

### `EccController`
- **Why:** Same as other list/detail endpoints: preserve **404** and **403** from the service layer.
- **Change:** Rethrow those exceptions before `SW360Exception` wrapping.

### `LicenseController` + `Sw360LicenseService`
- **`LicenseController.uploadLicenses`**
  - **Why:** Client bad requests (`BadRequestClientException`) and 404/403 must not be lumped into `SW360Exception`.
  - **Change:** Rethrow `ResourceNotFoundException`, `AccessDeniedException`, and `BadRequestClientException` first.
- **`Sw360LicenseService.addLicenseType`**
  - **Why:** `TException` and `RuntimeException` were being wrapped in a **new** `TException`, which can **lose type, cause chain, and handler behavior**.
  - **Change:** Rethrow `TException` and `RuntimeException`; only truly unexpected checked exceptions get wrapped.

### `ObligationController`
- **`getObligation` / `checkIfObligationExists`**
  - **Why:** Generic `catch` turned **any** error into “obligation does not exist” (**404**), including cases that were **403** or other failures.
  - **Change:** Rethrow `ResourceNotFoundException` and `AccessDeniedException` first; keep a controlled fallback for other errors where appropriate.
- **`deleteObligations` (multi-status)**
  - **Why:** Per-id failures were all **500** in the multi-status payload.
  - **Change:** Map `ResourceNotFoundException` to per-item **404**, `AccessDeniedException` to **403**; other errors remain **500**.
- **`editObligation`**
  - **Why:** Same passthrough pattern for 404/403/400 vs generic processing error.

### `ProjectController`
- **Attachment upload / update flows**
  - **Why:** Upload or project update can fail with 404/403 or client bad request; `DataIntegrityViolationException` maps to **409** in `RestExceptionHandler` and must not be wrapped as a generic `SW360Exception`.
  - **Change:** Rethrow those types before logging and wrapping unknown errors.
- **`setAdditionalFieldsToHalResource`**
  - **Why:** Embedded user/project resolution can throw 404/403; those should propagate.
  - **Change:** Rethrow `ResourceNotFoundException` and `AccessDeniedException` before generic `SW360Exception`.

### `Sw360ProjectService`
- **`handleIgnoredLicenses`**
  - **Why:** Processing ignored licenses can surface 404/403/bad request from the project hierarchy; wrapping everything in `TException` hid HTTP semantics for callers/controllers.
  - **Change:** Rethrow `ResourceNotFoundException`, `AccessDeniedException`, and `BadRequestClientException` first.
- **`validateProjectHierarchyChain` / `validateReleaseOwnership`**
  - **Why:** `catch (Exception e)` around `getProjectForUserById` converted **real 404/403** into a **generic `BadRequestClientException`** (“invalid hierarchy”), with the **wrong status and message**.
  - **Change:** Rely on `getProjectForUserById` to map backend codes to `ResourceNotFoundException` / `AccessDeniedException` and let them propagate. Stop treating **all** exceptions from `getReleaseIds` as validation failures; let **Thrift** errors propagate instead of forcing a single bad-request narrative.

### `SW360ReportController` / `SW360ReportService`
- **Why:** Report generation and Excel download paths call deep services that throw 404/403 and sometimes `BadRequestClientException`; wrapping as `SW360Exception`/`TException` turned them into **500**.
- **Change:** Consistent rethrow of `ResourceNotFoundException`, `AccessDeniedException`, and `BadRequestClientException` (and `TException` / `IOException` in service where already distinguished) before generic error translation.

### `Sw360ScheduleService`
- **Why:** `RuntimeException` (for example from validation or infrastructure) was being wrapped as a generic `TException` “Unexpected error”, **losing** the original exception class and sometimes the right handling.
  - **Change:** Rethrow `RuntimeException` after logging; keep wrapping only for unexpected checked `Exception`s if that remains the contract.

### `VendorController.exportVendor`
- **Why:** Same as other export endpoints: preserve 404/403/400 from the stack.
- **Change:** Rethrow `ResourceNotFoundException`, `AccessDeniedException`, and `BadRequestClientException` before `TException` wrapping.

### `VulnerabilityController`
- **Why:** Order matters: ensure **400/404/403** are not swallowed by a branch that converts `SW360Exception` to `TException` until those are ruled out.
- **Change:** Rethrow `BadRequestClientException`, `ResourceNotFoundException`, and `AccessDeniedException` first; keep the existing `SW360Exception` to `TException(e.why)` behavior for remaining domain errors.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/admin/fossology/FossologyAdminController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/admin/fossology/FossologyAdminController.java
@@ -34,6 +34,8 @@ import org.eclipse.sw360.rest.resourceserver.core.RestControllerHelper;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.rest.webmvc.BasePathAwareController;
 import org.springframework.data.rest.webmvc.RepositoryLinksResource;
+import org.springframework.data.rest.webmvc.ResourceNotFoundException;
+import org.springframework.security.access.AccessDeniedException;
 import org.springframework.hateoas.server.RepresentationModelProcessor;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -121,6 +123,10 @@ public class FossologyAdminController implements RepresentationModelProcessor<Re
         } catch (SW360Exception e) {
             throw new BadRequestClientException(e.getWhy(), e);
         } catch (BadRequestClientException e) {
+            throw e;
+        } catch (ResourceNotFoundException e) {
+            throw e;
+        } catch (AccessDeniedException e) {
             throw e;
         } catch (Exception e) {
             throw new SW360Exception(e.getMessage());
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/clearingrequest/ClearingRequestController.java
@@ -230,6 +230,10 @@ public class ClearingRequestController implements RepresentationModelProcessor<R
             HttpStatus status1 = resources == null ? HttpStatus.NO_CONTENT : HttpStatus.OK;
             return new ResponseEntity<>(resources, status1);
 
+        } catch (ResourceNotFoundException e) {
+            throw e;
+        } catch (AccessDeniedException e) {
+            throw e;
         } catch (Exception e) {
             throw new SW360Exception(e.getMessage());
         }
@@ -299,6 +303,10 @@ public class ClearingRequestController implements RepresentationModelProcessor<R
             }
             HttpStatus status = resources == null ? HttpStatus.NO_CONTENT : HttpStatus.OK;
             return new ResponseEntity<>(resources, status);
+        } catch (ResourceNotFoundException e) {
+            throw e;
+        } catch (AccessDeniedException e) {
+            throw e;
         } catch (Exception e) {
             throw new SW360Exception(e.getMessage());
         }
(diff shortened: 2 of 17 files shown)


Shivamrut:
Hi @GMishx, this PR is an extension of the work https://github.com/eclipse-sw360/sw360/pull/4082 
Please let me know if changes are required. Specifically for `ComponentController` and `Sw360ProjectService`. I have removed the blanket try catch and allowed the exception to propagate with correct response codes. Let me know if my approach is wrong.

GMishx (approved):
Changes looks good.

PR #4098, opened 2026-04-17
fix(LicenseInfo): use SPDX ID for obligation-to-license mapping

its-sushant:

## Problem
The existing license-to-obligation mapping only matched obligations using licenseName. This caused obligations to be missed for licenses since the obligation is mapped using the SPDX ID (licenseSpdxId) rather than the license name in fossology.

## Solution
Updated createLicenseToObligationMapping to perform a two-step lookup when associating obligations with a license:
- Primary lookup: if the license has a SPDX ID set (isSetLicenseSpdxId()), attempt to resolve obligations using licenseSpdxId first.
- Fallback lookup: if no obligations are found via SPDX ID (or SPDX ID is not set), fall back to matching by licenseName as before.


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #4099, opened 2026-04-17
fix(rest): PATCH /api/licenses/{id} returns 200 OK on write failure

Aman-Cool:
`PATCH /api/licenses/{id}` was silently returning `200 OK` even when the backend rejected the write. If `LicenseDatabaseHandler.updateLicense()` returned `RequestStatus.FAILURE` ; either because the user lacked write permission or because a business rule was violated (e.g. attempting to uncheck an already-checked license through a concurrent request that bypassed the controller-level guard); `Sw360LicenseService.updateLicense()` handed that status back raw without throwing, and `LicenseController.updateLicense()` only checked for `SENT_TO_MODERATOR`, so everything else fell through to `return 200 OK` with the pre-update license body.

The fix has two parts. First, `Sw360LicenseService.updateLicense()` now throws `BadRequestClientException` when the backend returns `FAILURE`, matching the pattern already used in `deleteLicenseById()` in the same class. This means `FAILURE` never reaches the controller. Second, the controller adds a `requestStatus != SUCCESS` guard as a forward-compatibility catch-all: if the backend ever returns a currently unhandled status (`IN_USE`, `ACCESS_DENIED`, `NOOP`, etc.), the guard prevents a silent `200` from slipping through again. These are two distinct concerns; the service throw handles the known `FAILURE` case today, the controller guard handles unknown statuses in the future.

The `@ApiResponses` annotation is updated to document the `400` response that was always possible but never declared. Error messages include the actual `RequestStatus` value so callers can see what the backend returned.

This is related to the broader pattern I've been fixing in #4070, where the PATCH path silently does the wrong thing; there it was data corruption due to Thrift IDL defaults, here it's a rejected write that looks like success to the caller. Also worth noting that the recently opened #4096 tackles a similar class of problem from the other direction: exception rewrapping that hides real 404/403 status codes. The issue here is the complement of that; not an exception being lost, but a return value never being checked.

Issue: NA

### Suggest Reviewer

### How To Test?

1. As a user without write permission on a license, call:
```
PATCH /api/licenses/{id}
{ "fullName": "some change" }
```
**Before:** `200 OK` with the license body; write silently rejected, caller has no idea.
**After:** `400 Bad Request` with `"License update failed with status: FAILURE"` in the response body.

2. Create a license, mark it `checked: true`, then try to set `checked: false` via two near-simultaneous PATCH requests where the second one fetches the unchecked state just before the first commits; the controller-level guard doesn't fire but the backend returns `FAILURE` at the DB layer.
**Before:** `200 OK`.
**After:** `400 Bad Request`.

3. A unit test (`Sw360LicenseServiceTest`) verifies that when the Thrift client returns `RequestStatus.FAILURE`, the service throws `BadRequestClientException` with the status value in the message. A controller-level spec test (`LicenseSpecTest`) verifies the `400` response and response body when the service throws.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
@@ -290,7 +290,7 @@ public class LicenseController implements RepresentationModelProcessor<Repositor
             @ApiResponse(responseCode = "405",
                     description = "Reject license update due to: an already checked license is not allowed" +
                             " to become unchecked again"),
-            @ApiResponse(responseCode = "500", description = "License update failed (permission denied or business rule violation).")
+            @ApiResponse(responseCode = "400", description = "License update failed (permission denied or business rule violation).")
     })
     @PatchMapping(value = LICENSES_URL + "/{id}")
     public ResponseEntity<EntityModel<License>> updateLicense(
@@ -312,7 +312,7 @@ public class LicenseController implements RepresentationModelProcessor<Repositor
             return new ResponseEntity(RESPONSE_BODY_FOR_MODERATION_REQUEST, HttpStatus.ACCEPTED);
         }
         if (requestStatus != RequestStatus.SUCCESS) {
-            throw new RuntimeException("Error updating license");
+            throw new RuntimeException("License update failed with status: " + requestStatus);
         }
         HalResource<License> halResource = createHalLicense(licenseUpdate);
         return new ResponseEntity<>(halResource, HttpStatus.OK);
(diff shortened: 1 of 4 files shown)


Aman-Cool:
Hey @GMishx, small fix here; `PATCH /api/licenses/{id}` was silently returning `200 OK` whenever the backend rejected the write (permission denied or business rule violation). The `FAILURE` status from the backend just never got checked, controller fell straight through to `200`. 
Added the throw in the service layer to match what `deleteLicenseById` already does in the same class, and a `requestStatus != SUCCESS` catch-all in the controller so any currently-unhandled statuses can't silently slip through in the future either. 
This is in the same vein as #4096 (fixing the REST layer being dishonest about failures) and #4070 (where PATCH was also silently doing the wrong thing due to Thrift defaults). 
Would love your review when you get a chance!

Aman-Cool:
Hey @shivumrut, just wanted to flag this so it doesn't catch you off guard; both our PRs touch `LicenseController.java` and `Sw360LicenseService.java` but in completely different methods. 
Yours (#4096, extending #4082) is in `uploadLicenses` and `addLicenseType`, mine is in `updateLicense`, so there's no logic conflict at all. 
If yours merges first I'll just rebase. The bug here is sort of the flip side of what you're fixing; you're catching exceptions that were being rewrapped and hiding 403/404, I'm catching a return value that was being ignored and hiding a rejected write entirely. 
Really like the direction of #4096 by the way, same spirit of making the REST layer honest.

GMishx (approved):
LGTM

PR #4103, opened 2026-04-18
fix(components): prevent duplicate components by name only, case insensitive

### LINKED ISSUE #4100: Duplicate components created
### Description

When user are trying to upload the components via excel velocity tool, Sometimes the component creation is duplicated in the app. 

#### How to reproduce

Create a component with same name and Type.

Image

Image

ADITYA-CODE-SOURCE:
## Problem
When creating a component via REST API or Excel upload, duplicate components could be created with the same name (case-insensitive mismatch).

Example: Component "ABC" exists, trying to add "abc" would succeed because import used case-sensitive check.

## Solution
- Removed duplicate check from `ComponentDatabaseHandler.updateComponents()` because it breaks merge/update operations
- Changed `ComponentImportUtils.updateComponents()` to use case-insensitive `TreeSet` instead of `HashSet`

## Changes
- `libraries/importers/src/main/java/org/eclipse/sw360/importer/ComponentImportUtils.java`
  - Line 378: `HashSet` → `TreeSet<>(String.CASE_INSENSITIVE_ORDER)`

## Behavior
| Scenario | Result |
|----------|--------|
| Component "ABC" (OSS) exists | Existing |
| Create "ABC" (OSS) | Blocked (duplicate) |
| Create "ABC" (COTS) | Blocked (duplicate) |
| Create "abC" (INTERNAL) | Blocked (duplicate) |

All components with the same name (case-insensitive) are blocked, regardless of type.

## Verification
- REST API: Already has case-insensitive check in `addComponent()`
- Excel Import: Now uses `TreeSet<>(CASE_INSENSITIVE_ORDER)` for case-insensitive check
- Merge/Update operations: Not affected (duplicate check not added to updateComponents)

---
fix #4100


ADITYA-CODE-SOURCE:
Hello sir  @GMishx @KoukiHama @ag4ums @arunazhakesan,

This PR fixes issue #4100 - prevents duplicate components by checking both name AND type.

Could you please review?

GMishx:
Thank you for the fix @ADITYA-CODE-SOURCE but we **do not want** a type based check. There should be only one component for a given name, case insensitive, irrespective of its type.

| Scenario | Before Fix | After Fix |
|----------|-----------|-----------|
| Component "ABC" (OSS) exists | - | - |
| Create "ABC" (OSS) | Created (dup) | Blocked |
| Create "ABC" (COTS) | Created (bug) | Blocked |
| Create "abC" (INTERNAL) | - | Blocked |

ADITYA-CODE-SOURCE:
> Thank you for the fix @ADITYA-CODE-SOURCE but we **do not want** a type based check. There should be only one component for a given name, case insensitive, irrespective of its type.
> 
> Scenario	Before Fix	After Fix
> Component "ABC" (OSS) exists	-	-
> Create "ABC" (OSS)	Created (dup)	Blocked
> Create "ABC" (COTS)	Created (bug)	Blocked
> Create "abC" (INTERNAL)	-	Blocked

@GMishx Thank you for the clarification  Understood — name only  case insensitive, no type check  I will update the PR description and code accordingly

ADITYA-CODE-SOURCE:
Hello sir @GMishx,
Thank you for the feedback. I've updated the PR to align with your requirement.
**Changes made:**
1. **Code**: Reverted to the existing `isDuplicate(component.getName(), true)` method which checks name only, case-insensitive
2. **Removed**: The type-based methods (`isDuplicateByNameAndType` and `getComponentIdsByNameAndType`) as they're no longer needed
3. **Description**: Updated to clearly state that type does not affect duplicate detection

The fix now correctly enforces: **one component per name, case-insensitive, regardless of type**.
Could you please take another look when you have time?

Thanks for your guidance!

GMishx:
Can you explain the changes @ADITYA-CODE-SOURCE ? The PR currently looks like it just adds log messages and not solve the actual issue.

ADITYA-CODE-SOURCE:
> Can you explain the changes @ADITYA-CODE-SOURCE ? The PR currently looks like it just adds log messages and not solve the actual issue.

 @GMishx,
I apologize for the confusion in this PR. You're right to question the changes.
The history in this PR has mixed commits from different fixes:
- Commit `ae444f4` - Contains the original component duplicate fix (added type-based checking)
- Commits `8af30fa`, `b6ac01a`, `9cbb63c` - From other PRs (SBOM fix, integration tests)
- Commit `01b1a28` - Reverts to name-only check using existing `isDuplicate(component.getName(), true)`
The actual fix for issue #4100 is in commit `01b1a28` which:
- Reverts to using `isDuplicate(component.getName(), true)` 
- Removes the type-based methods I originally added
However, you're right that the PR description and title are outdated and confusing.

I apologize for the confusion and appreciate your patience.

ADITYA-CODE-SOURCE:
Hello sir  @GMishx, I apologize for the confusion. I've cleaned up the PR completely.

Final fix: Added duplicate check in updateComponents() using isDuplicate(c.getName(), true) - name only (case-insensitive), no type check.

 Thank you . Please review when you have time.

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -925,7 +925,16 @@ private void prepareComponent(Component component) throws SW360Exception {
     }
 
     public RequestSummary updateComponents(Set<Component> components, User user) throws SW360Exception {
-        return RepositoryUtils.doBulk(prepareComponents(components), user, componentRepository);
+        Set<Component> filteredComponents = components.stream()
+                .filter(c -> !isDuplicate(c.getName(), true))
```
GMishx: This function is used by:
1. Merge Releases operation
2. Merge Components operation
3. Merge Vendors operation
4. Import of Releases
5. Import of Components

Checking duplicate here will always be `true` as the component being updated exists in the database.

GMishx (changes requested):

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
ADITYA-CODE-SOURCE: > This function is used by:
> 
> 1. Merge Releases operation
> 2. Merge Components operation
> 3. Merge Vendors operation
> 4. Import of Releases
> 5. Import of Components
> 
> Checking duplicate here will always be `true` as the component being updated exists in the database.

hello sir @GMishx
 I've updated the fix. Removed duplicate check from updateComponents() (would break merge/update) 
Added case-insensitive check in ComponentImportUtils using TreeSet<>(CASE_INSENSITIVE_ORDER) instead 
Could you please review?

GMishx (approved):
Should be good to go.

PR #4126, opened 2026-05-04
fix(keycloak): gson dependency issue.

rudra-superrr:
Issue: The SW360 Keycloak providers (event-listener, user-storage-provider) fail to initialize with NoClassDefFoundError: Could not initialize class org.eclipse.sw360.keycloak.common.Sw360UserService.

Root cause: The sw360-keycloak-common shaded JAR was missing the Gson dependency (com.google.code.gson:gson). The IBM Cloudant SDK uses Gson internally to parse HTTP responses from CouchDB. Without it, the SDK throws "Error processing the HTTP response" during the Sw360UserService static initializer, and the JVM permanently marks the class as unloadable.

Fix: Added com.google.code.gson:gson to the shade plugin's <artifactSet><includes> and added a corresponding <relocation> entry to relocate com.google.gson → org.eclipse.sw360.shaded.gson to avoid classpath conflicts with Keycloak's runtime.

### DIFF
--- a/keycloak/sw360-keycloak-common/pom.xml
+++ b/keycloak/sw360-keycloak-common/pom.xml
@@ -105,6 +105,7 @@
                                     <include>org.eclipse.sw360:nouveau-handler</include>
 
                                     <include>com.ibm.cloud:*</include>
+                                    <include>com.google.code.gson:gson</include>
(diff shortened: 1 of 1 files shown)


GMishx (approved):
Tested, working as expected.

PR #4127, opened 2026-05-04
feat(test): jacoco reports in CI

GMishx:

This PR does:
1. Remove all `System.out` and `System.err` and replace them with logger.
2. Fix the surefire configuration:
    1. Single `argLine` was split in two causing overwrite and fix that.
    2. Provide path to mockito jar and add it as a top level test dependency.
    3. Prevent jGiven from polluting the output channel and giving warning `[WARNING] Corrupted channel by directly writing to native stream in forked JVM 1.`
3. Generate JaCoCo reports in CI and save them for 7 days as artifacts.
4. Ignore Thrift generated classes from Coverage reports.


amritkv (approved):
LGTM !

PR #4202, opened 2026-05-20
fix(report): Encode Content-Disposition filename and reduce debug log verbosity

bibhuti230185:

## Summary

Fixes two issues when generating license-info reports for projects with non-ASCII characters (™, ü, ñ, etc.) in their name:

1. **Content-Disposition header rejected by Tomcat** — The raw Unicode filename caused an `UnmappableCharacterException`, Tomcat silently removed the header, and the browser fell back to a generic filename (`LicenseInfo-2026-05-16.zip`) losing the intended `.docx` extension.
2. **Debug log pollution** — `Sw360ProjectService:266` logged the full attachment-usage key string (hundreds of KB) at DEBUG level on every request.

### Suggest Reviewer
@GMishx 

## Changes

| File | Change |
|------|--------|
| `SW360ReportController.java` | Added `setContentDisposition()` helper using RFC 5987 dual-filename encoding. Replaced all 3 header-setting sites. |
| `Sw360ProjectService.java` | Changed `log.debug("Resulting string: ...")` → `log.trace(...)` |
| `SW360ReportControllerContentDispositionTest.java` | Unit test for encoding logic (ASCII, ™, spaces, umlauts) |
| `ReportTest.java` | Integration test verifying Content-Disposition header via `TestRestTemplate` |

### RFC 5987 Encoding

```
Content-Disposition: attachment; filename="LicenseInfo-Gridscale X_ Protection-1.2.1.docx"; filename*=UTF-8''LicenseInfo-Gridscale%20X%E2%84%A2%20Protection-1.2.1.docx
```

- `filename` — ASCII-safe fallback (non-ASCII replaced with `_`) that Tomcat can serialize
- `filename*` — Full Unicode name that modern browsers prefer for display/save

## How To Test?

1. Create a project with name containing special characters (e.g. `Gridscale X™ Protection Data Manager`)
2. Generate a license-info report (DOCX format) via `/api/reports?module=licenseInfo&projectId=...&generatorClassName=DocxGenerator&variant=REPORT`
3. Verify:
   - Downloaded file has correct Unicode name with `.docx` extension
   - No `WARNING` in Tomcat catalina log about invalid `Content-Disposition`
   - DEBUG logs do not contain the full attachment key dump

https://github.com/user-attachments/assets/722fbaf0-accc-4bc7-b09d-72c140bd0fae

## Checklist

Must:
- [x] Signed-off-by present in commit
- [x] Unit tests added
- [x] Integration test added
- [x] Pre-commit hooks pass (formatting, conventional commit)

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
@@ -263,7 +263,7 @@ public class Sw360ProjectService implements AwareOfRestServices<Project> {
                 selectedData.add(stringResult);
             }
         }
-        log.debug("Resulting string: {}", selectedData);
+        log.trace("Resulting string: {}", selectedData);
         return selectedData;
     }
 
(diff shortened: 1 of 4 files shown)


GMishx (approved):
LGTM

PR #4221, opened 2026-05-27
fix(project): guard empty release IDs in project count

bibhuti230185:
## Summary
Fixes project update flow when release-id input is empty/null by guarding `countProjectsByReleaseIds(...)` in REST service before Thrift backend call.

### Changes:
- Guard added in `rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java`
  - return `0` when `releaseIds` is null/empty
- Unit tests added in `rest/resource-server/src/test/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectServiceTest.java`
  - empty/null short-circuit
  - non-empty Thrift path
  - Thrift-failure fallback to `0`
  - delegation checks for release-id search methods

### Suggest Reviewer
@GMishx 

### How To Test?

Run unit tests:
```bash
mvn test -pl rest/resource-server -Dtest=Sw360ProjectServiceTest
```

Manual verification:
1. Create a project without direct releases.
2. Update that project.
3. Confirm no `SW360Exception(why:Invalid empty input!)` occurs.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
@@ -1462,6 +1462,9 @@ public class Sw360ProjectService implements AwareOfRestServices<Project> {
      * @throws TException
      */
     public int countProjectsByReleaseIds(Set<String> releaseIds) {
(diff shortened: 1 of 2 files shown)


GMishx (approved):
Good to go.

PR #4236, opened 2026-05-29
fix(project): save attachmentusage returns 400.

rudra-superrr:
Summary
Saving attachment usages on the Attachment Usage page fails with a 400 Bad Request when any release has been unlinked from the project (or a sub-project has been removed from the hierarchy) after the page was loaded. The entire save operation is blocked, and all user changes are lost.

Steps to Reproduce
- Open a project's Attachment Usage tab (with sub-projects and linked releases)
- In another tab, unlink a release or sub-project from the project
- Go back to the original tab and click "Save Usages"
- Result: 400 error — all changes lost

Expected Behavior
- Save should succeed for all valid entries
- Invalid/stale entries should be skipped with a warning shown to the user
- Stale attachment usage records in the DB should be cleaned up automatically so the warning doesn't recur

Root Cause
The [validate] method threw [DataIntegrityViolationException] on the first invalid release, and [validateProjectHierarchyChain]/[validateReleaseOwnership] threw [BadRequestClientException] for stale hierarchy entries — aborting the entire request.

Fix
- Changed validation to warn-and-skip: invalid entries are removed from the save payload and warnings are returned in the response
- Added self-healing: stale [AttachmentUsage] records in CouchDB (referencing releases no longer linked) are automatically deleted on the first save attempt
- Response includes a [warnings] array so the frontend can display which entries were skipped

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -2287,11 +2287,28 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
                 Set<String> totalReleaseIds = projectService.getReleaseIds(id, user, true);
                 changedUsages.addAll(selectedUsages);
                 changedUsages.addAll(deselectedUsages);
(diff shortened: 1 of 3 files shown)


GMishx (approved):
Changes looks good.

PR #4238, opened 2026-06-01
Fix/atachments

deo002:

> Handle null error when there is no attachment linked to a release when  fetching from projects/{id}/licenseDbObligations endpoint.
> Create a new license if it is not found in the database but detected in the clx report. While not the main issue, it is a nice to have.

### Issue
Request:
```
curl 'http://localhost:8080/resource/api/projects/1308e0f092bc1ff58b96af832c6234ba/licenseDbObligations?page=0&page_entries=10&sort=' \
  -H 'Authorization: Basic xyz'
```
Response:
```
{
  "timestamp" : "2026-06-01T10:55:13.832336484Z",
  "status" : 400,
  "error" : "Bad Request",
  "message" : "Cannot invoke \"java.util.Set.iterator()\" because the return value of \"org.eclipse.sw360.datahandler.thrift.components.Release.getAttachments()\" is null"
}
```

### How To Test?
Attach clx file to a project's release which has sibling releases with zero attachments.
Then hit the /licenseDbObligations endpoint. It should not return an error.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -888,7 +888,7 @@ public class RestControllerHelper<T> {
         }
     }
 
-    private void createMissingLicense(String licenseId) throws Exception {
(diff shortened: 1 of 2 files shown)


GMishx (approved):
LGTM

PR #4240, opened 2026-06-01
fix(rest): handle null project clearing state

GMishx:
## What

This change fixes `PATCH /api/projects/{id}` for projects that have a null `clearingState` (legacy/partially-populated records), and updates OpenAPI documentation for the endpoint responses.

## Changes

- `rest/resource-server/.../ProjectController.java`
  - Add OpenAPI response docs for:
    - `200` Project successfully updated
    - `202` Accepted - update request sent to moderation
  - In `patchProject`:
    - Move moderation comment/write-check logic before update
    - Default `sw360Project.clearingState` to `OPEN` when null
    - Call `PermissionUtils.checkEditablePermission(...)` using `ProjectClearingState`

- `libraries/datahandler/.../PermissionUtils.java`
  - Change `checkEditablePermission` signature from `String` to `ProjectClearingState`
  - Add null fallback to `ProjectClearingState.OPEN`

- `rest/resource-server/.../integration/ProjectTest.java`
  - Add regression test:
    - `should_update_project_when_clearing_state_is_null`

## Why

`patchProject` previously dereferenced `getClearingState().name()` and could fail for projects with missing clearing state data. This change makes the flow null-safe and documents both observed success paths (`200` and `202`).

## How to validate

1. Run project tests for the resource-server module.
2. Verify the new integration test passes.
3. Check generated OpenAPI output for `PATCH /api/projects/{id}` includes both `200` and `202`.

### DIFF
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/PermissionUtils.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/PermissionUtils.java
@@ -178,8 +178,11 @@ public class PermissionUtils {
         }
     }
 
-    public static boolean checkEditablePermission(String name, User user, Map<String, Object> reqBodyMap, Project sw360Project) {
-        if (!name.equals(ProjectClearingState.CLOSED.name()) || PermissionUtils.isAdmin(user)) {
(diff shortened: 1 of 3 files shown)


rudra-superrr (approved):
LGTM

PR #4258, opened 2026-06-06
feat(ci): update keycloak tag if image updates

GMishx:
Update the keycloak image tag version if base image version is updated by dependabot


GMishx:
Close in favor of https://github.com/eclipse-sw360/sw360/pull/4259

PR #4311, opened 2026-06-26
fix(Project): Fixed vendor selection upon edit

### LINKED ISSUE #4310: Current Vendor not Selected in Vendor Modal in edit project and component page
### Description

The current vendor does not come pre-selected because vendor id is not sent from backend. 

#### How to reproduce

Check edit page of components/projects.

### Versions

* Docker version OR
* Last commit id on main

#### SW360 logs

Farooq-Fateh-Aftab:

>

This PR fixes a bug where `vendorId` was missing from the `GET /projects` and `GET /projects/{id}` API responses, causing the vendor selector to render empty when a user opened the **Edit Project** dialog in the UI.

### Root Cause

The backend `fillVendor()` method resolves the stored `vendorId` into a full `Vendor` object on the `Project` entity. In the REST layer, after that `Vendor` object is embedded as a HAL resource (`sw360:vendors`), `setVendor(null)` is called to keep the flat JSON clean. Because `fillVendor()` had already unset `vendorId`, the field was absent from every project response – so the frontend had no ID to pre-populate the vendor selector.

> * Which issue is this pull request belonging to and how is it solving it? (*https://github.com/eclipse-sw360/sw360/issues/4310*)

### Changes

| File | Change |
|------|--------|
| `ProjectController.java` (×2) | Before calling `setVendor(null)`, restore `vendorId` from `vendor.getId()` so it remains a direct field in the JSON response |
| `RestControllerHelper.java` | In `convertToEmbeddedProject()`, derive `vendorId` from the hydrated `Vendor` object (or fall back to the raw field) when building embedded project representations |

Closes https://github.com/eclipse-sw360/sw360/issues/4310

### Checklist
Must:

### Frontend Releated PR
https://github.com/eclipse-sw360/sw360-frontend/pull/1816

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/core/RestControllerHelper.java
@@ -965,6 +965,13 @@ public class RestControllerHelper<T> {
         if (project.isSetLinkedProjects()) {
             embeddedProject.setLinkedProjects(project.getLinkedProjects());
         }
+        // fillVendor() unsets vendorId and replaces it with the full Vendor object,
+        // so derive the ID back from the vendor object when present.
+        if (project.getVendor() != null && project.getVendor().getId() != null) {
+            embeddedProject.setVendorId(project.getVendor().getId());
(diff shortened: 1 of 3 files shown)


GMishx (approved):
LGTM

PR #4323, opened 2026-07-02
feat(search): Add search support for moderation requests

deo002:

Added search support for moderatiion requests based on key value pairs. It will be used in the Advanced Search component in frontend.

By default, only the moderation requests in which the user is a requesting user or a moderator are shown. If there are other moderators or requesting user in the search parameters then requests of all the users in the search parameters along with the requests of the current user are shown. 

### How To Test?
Try making api calls to the GET moderationrequests endpoint via curl/swagger with luceneSearch as true/false and various search parameters.
 
### Checklist
Must:


GMishx (approved):
Good to go.

PR #4339, opened 2026-07-06
fix(package): Added indexing and pagination for retrieval

saad-mohammed1:
Added a dedicated index to optimize package search and retrieval.
Refactored package retrieval logic in PackageRepository to use the optimized query.
Added pagination support in SW360PackageService.
Updated PackageController to expose paginated package retrieval through the REST API.

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/packages/PackageController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/packages/PackageController.java
@@ -39,6 +39,7 @@ import org.eclipse.sw360.datahandler.couchdb.lucene.NouveauLuceneAwareDatabaseCo
 import org.eclipse.sw360.datahandler.resourcelists.PaginationParameterException;
 import org.eclipse.sw360.datahandler.resourcelists.PaginationResult;
 import org.eclipse.sw360.datahandler.resourcelists.ResourceClassNotFoundException;
+import org.eclipse.sw360.datahandler.thrift.PaginationData;
 import org.eclipse.sw360.datahandler.thrift.RequestStatus;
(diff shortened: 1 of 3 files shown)


Suhas2109:
Hi @saad-mohammed1 ,
Test was successful!

Suhas2109:
Hi @akshitjoshii ,
Could you please do the code review?

akshitjoshii (dismissed):
LGTM. follows the same pagination pattern as in vendor and project pages.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/packages/PackageController.java
```diff
@@ -281,13 +283,30 @@ public ResponseEntity<?> getPackagesForUser(
             }
             sw360Packages.addAll(packageService.refineSearch(restrictions, sw360User));
         } else {
-            sw360Packages = packageService.getPackagesForUser();
+            boolean requestContainsPaging = request.getParameterMap().containsKey("page")
+                    || request.getParameterMap().containsKey("page_entries");
+            boolean useDbPagination = requestContainsPaging && restrictions.isEmpty() && !orphanPackage;
```
GMishx: Please use `pageable` parameter instead of querying raw values from the `request` parameter.

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/packages/PackageController.java
saad-mohammed1: Thanks for the suggestion @GMishx . Updated the implementation to use the Pageable parameter instead of querying the raw request parameters.

GMishx (approved):
LGTM

PR #4359, opened 2026-07-14
fix(#4066): invaild pURL validation error

### LINKED ISSUE #4066: Fix error message in API response when `pURL` is invalid
### Description

Need to refactor the error message when the `pURL` structure is invalid. The error message received from the backend is `Dependent document Id/ids not valid.` which is confusing to the user.

Actually, this issue was raised due to invalid `pURLs`. So, it would be great to show the error be like `Invalid pURL` or something which can be transparent to the user instead of non-transparent error `Dependent document Id/ids not valid`

The server error log captured at top level error as `Invalid pURL for package: <pkg_name>`, `The packageURL type contains invalid character` and next level error is `Dependent document Id/ids not valid`

#### How to reproduce

- Try to `Add Package`/ `Edit Package`
- Enter a `pURL` like : `pkg:github@test/version@1.2.3` or something similar where purl starts with `pkg:<package_manager>
- Hit `Add Package` / `Update Package`

Image

Shivani2118:

> * this issue was raised due to invalid pURLs. So, it would be great to show the error be like Invalid pURL or something which can be transparent to the user instead of non-transparent error Dependent document Id/ids not valid.(*Refer to issue #4066*)

Closes #4066

Enter a pURL like : pkg:github@test/version@1.2.3 or something similar where purl starts with `pkg:<package_manager>
Hit Add Package / Update Packages?

### Checklist
Must:

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java
@@ -190,15 +190,24 @@ public class PackageDatabaseHandler extends AttachmentAwareDatabaseHandler {
             return new AddDocumentRequestSummary().setRequestStatus(AddDocumentRequestStatus.INVALID_INPUT).setMessage("Invalid Pacakge Type!");
         }
 
+        String purlError = validatePurl(pkg.getPurl());
+        if (purlError != null) {
+            log.error(String.format("Invalid PURL for package: '%s' - %s", SW360Utils.printName(pkg), purlError));
+            return new AddDocumentRequestSummary().setRequestStatus(AddDocumentRequestStatus.INVALID_INPUT)
+                    .setMessage(purlError);
+        }
+
         try {
             PackageURL purl = new PackageURL(pkg.getPurl());
             pkg.setPackageManager(PackageManager.valueOf(purl.getType().toUpperCase()));
         } catch (MalformedPackageURLException e) {
             log.error(String.format("Invalid PURL for package: '%s'", SW360Utils.printName(pkg)), e);
-            return new AddDocumentRequestSummary().setRequestStatus(AddDocumentRequestStatus.INVALID_INPUT).setMessage("Invalid Pacakge URL!");
+            return new AddDocumentRequestSummary().setRequestStatus(AddDocumentRequestStatus.INVALID_INPUT)
+                    .setMessage(e.getMessage() + " in '" + pkg.getPurl() + "'. Expected format: pkg:type/namespace/name@version");
         } catch (IllegalArgumentException e) {
             log.error(String.format("Invalid Package Manager for package: '%s'", SW360Utils.printName(pkg)), e);
-            return new AddDocumentRequestSummary().setRequestStatus(AddDocumentRequestStatus.INVALID_INPUT).setMessage("Invalid Pacakge Manager!");
+            return new AddDocumentRequestSummary().setRequestStatus(AddDocumentRequestStatus.INVALID_INPUT)
+                    .setMessage("Unsupported package manager type in pURL '" + pkg.getPurl() + "'. " + e.getMessage());
         }
 
         // Prepare package for database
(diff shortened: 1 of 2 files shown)


backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java
```diff
@@ -288,7 +288,7 @@ public RequestStatus updatePackage(Package updatedPkg, User user) throws SW360Ex
             updatedPkg.setPackageManager(PackageManager.valueOf(purl.getType().toUpperCase()));
         } catch (MalformedPackageURLException e) {
             log.error(String.format("Invalid PURL for package: %s", packageId), e);
-            return RequestStatus.INVALID_INPUT;
+            throw fail(400, "Invalid pURL: %s", e.getMessage());
         } catch (IllegalArgumentException e) {
             log.error(String.format("Invalid Package Manager for package: %s", packageId), e);
             return RequestStatus.INVALID_INPUT;
```
GMishx: What about this response? How this would be differentiated and informed to user?

GMishx:
Also, please use conventional commit format. `fix(eclipse-sw360#4066): invaild pURL validation error` is not a valid commit format followed in SW360.

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java
Shivani2118: The response includes the specific validation message returned by PURL parser, so the user will be informed of the exact reason for the invalid pURL.(eg.,"Invalid pURL: Invalid purl: does not contain both a type and name" OR "Invalid pURL: The PackageURL type contains invalid characters" )

amritkv:
Hey @Shivani2118 !
The issue was to fix the error message from backend side itself.
Instead of some generic message like : `Dependent document Id/ids not valid`, if you check the backend log attached in the issue #4066, there is an error line saying : `The packageURL type contains invalid characters`.
We want something similar which is more meaningful to the user.

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java
GMishx: No, the user will get `Invalid Purl or linked release id.` from https://github.com/eclipse-sw360/sw360/blob/619090976189d9eb310cdd79c1af071847717685/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/packages/SW360PackageService.java#L74

But from the log message, we see the exception is thrown when the package manager is invalid (Thrown by [PackageManager.valueOf](https://github.com/eclipse-sw360/sw360/blob/619090976189d9eb310cdd79c1af071847717685/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java#L288)). Still causing confusion about the ACTUAL issue.

backend/common/src/main/java/org/eclipse/sw360/datahandler/db/PackageDatabaseHandler.java
```diff
@@ -452,4 +467,77 @@ private void removeLeadingTrailingWhitespace(Package pkg) {
     public Map<PaginationData, List<Package>> getPackagesWithPagination(PaginationData pageData) {
           return packageRepository.getPackagesWithPagination(pageData);
     }
+
+    /**
+     * Validates pURL structure and returns error message with position, or null if valid.
+     */
+    private String validatePurl(String purl) {
```
GMishx: This is already done in `new PackageURL(pkg.getPurl())` on line https://github.com/eclipse-sw360/sw360/pull/4359/changes#diff-3cdd27e00398d3fa799dc4871459c42231e1c447b780f5fdd2aef0cb89c2a78fL194

All you had to do is catch the exception and reply with appropriate message.

Please remove this redundant, self-implementation function.

GMishx (changes requested):

GMishx (approved):
Changes looks good.

GMishx:
@amritkv , can you please test this PR if you are getting proper messages now?

amritkv:
Hey @GMishx & @Shivani2118 !
Tested this PR and it is working as expected.

Note : @Shivani2118 , we've to make necessary changes to the frontend PR https://github.com/eclipse-sw360/sw360-frontend/pull/1856

PR #4370, opened 2026-07-17
Block license writes with HTTP 410 when LicenseDB integration is enabled

Sandipmandal25:
## Summary

Makes SW360 license write endpoints read-only when `licensedb.enabled=true`. Since LicenseDB is the source of truth, direct license write operations now return **HTTP 410 Gone**.

## Changes

- Added `throwIfLicenseDBEnabled()` in `LicenseController`
- Guarded 7 license write endpoints with HTTP 410 responses
- Left SW360-native metadata/overlay endpoints (`updateWhitelist`, obligation links, license types) unchanged
- Added `@ApiResponse(410)` to all gated endpoints
- Added test covering the read-only behavior when LicenseDB is enabled

**Suggested reviewers:** @GMishx  @amritkv @Farooq-Fateh-Aftab

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
@@ -82,6 +82,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static org.eclipse.sw360.rest.resourceserver.core.RestControllerHelper.throwIfLicenseDBEnabled;
 import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.linkTo;
 
 @BasePathAwareController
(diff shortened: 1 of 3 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
```diff
@@ -470,6 +478,13 @@ public RepositoryLinksResource process(RepositoryLinksResource resource) {
         return resource;
     }
 
+    private void throwIfLicenseDBEnabled() {
```
GMishx: Please move this into rest control helper along with throw if security user.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
```diff
@@ -470,6 +478,13 @@ public RepositoryLinksResource process(RepositoryLinksResource resource) {
...
+        if (SW360Utils.readConfig(SW360ConfigKeys.LICENSEDB_ENABLED, false)) {
+            throw new HttpClientErrorException(HttpStatus.GONE,
```
GMishx: `GONE` is not a good status here. I'd recommend using `405` with header `Allow: GET` so the client knows, I cannot modify the data, only get it. Or a `403`.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
```diff
@@ -486,10 +501,12 @@ private HalResource<License> createHalLicense(License sw360License) {
     )
     @PreAuthorize("hasAuthority('WRITE')")
     @ApiResponses(value = {
-            @ApiResponse(responseCode = "200", description = "All licenses deleted successfully.")
+            @ApiResponse(responseCode = "200", description = "All licenses deleted successfully."),
+            @ApiResponse(responseCode = "410", description = "License management is disabled — LicenseDB integration is active.")
```
GMishx: Please avoid unnecessary Unicode characters like `—` in source files.

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/LicenseController.java
Sandipmandal25: moved!

GMishx (approved):
Good to go.

PR #4292, opened 2026-06-22
Add sync engine to import licenses and obligations from LicenseDB

Sandipmandal25:
## Summary

Adds the LicenseDB sync engine as the second part of the SW360 and LicenseDB integration.

This builds on the HTTP client added in #4268 and adds the complete flow for fetching licenses and obligations from LicenseDB and storing them in SW360.

## Changes

- added `importAllLicenseDBLicenses()` in `LicenseDatabaseHandler`
- reads the LicenseDB URL, credentials, and enabled flag from `LICENSEDB_REST`
- fetches active obligations from `/api/v1/obligations/export`
- fetches active licenses from `/api/v1/licenses/export`
- imports obligations first because licenses reference them by UUID
- matches existing data using `externalIds["licensedb-id"]`
- falls back to shortname matching during the first sync
- preserves SW360 specific fields like `checked`, `FSFLibre`, `whitelist`, and manually added obligations
- stores the last successful sync time in the `licensedb-sync-state` CouchDB document
- added `LicenseDBDataMapper` to map LicenseDB DTOs to SW360 Thrift objects
- maps obligation types like `RISK`, `RESTRICTION`, `RIGHT`, and `OBLIGATION`

## Endpoint

- added ADMIN only endpoint `POST /api/licenses/import/LicenseDB`

## Suggested reviewers

* @GMishx 
* @deo002 
* @Farooq-Fateh-Aftab

PR #4372, opened 2026-07-21
Add admin endpoints for LicenseDB sync and health monitoring

Sandipmandal25:
## Summary

This PR adds admin controls for the LicenseDB sync lifecycle.

It builds on the sync engine from #4292 and adds a small admin API to check LicenseDB connectivity, view the current sync status, trigger a full sync, and run an incremental sync based on changes from LicenseDB.

## What changed

* Added admin endpoints for LicenseDB health, sync status, full sync, and incremental sync
* Added a service layer to keep the controller thin
* Added a LicenseDB health check to verify connectivity
* Added incremental sync support using the LicenseDB audits API
* Reused the existing full-sync logic where possible
* Added report based sync status and timestamp lookup without fixed state documents
* Added Thrift methods for sync status and health checks with admin access control

## Endpoints

* `GET /api/licenseDB/health`
* `GET /api/licenseDB/syncStatus`
* `POST /api/licenseDB/sync`
* `POST /api/licenseDB/sync/incremental`

All endpoints require `ADMIN` authority.

## Suggested reviewers

@GMishx
@deo002
@Farooq-Fateh-Aftab

### DIFF
--- a/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
+++ b/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
@@ -100,6 +100,7 @@ public class LicenseDatabaseHandler {
     private final ObligationNodeRepository obligationNodeRepository;
     private final LicenseTypeRepository licenseTypeRepository;
     private final LicenseObligationListRepository obligationListRepository;
+    private final LicenseDBSyncReportRepository syncReportRepository;
     private final LicenseModerator moderator;
     private final CustomPropertiesRepository customPropertiesRepository;
     private final DatabaseRepositoryCloudantClient[] repositories;
@@ -108,7 +109,6 @@ public class LicenseDatabaseHandler {
     private static boolean IMPORT_STATUS = false;
     private static long IMPORT_TIME = 0;
     private static final long TIME_OUT = 1800000; // 30 minutes: 30 * 60 * 1000;
-    private static final String LICENSEDB_SYNC_STATE_DOC_ID = "licensedb-sync-state";
     private String obligationText;
     private final Logger log = LogManager.getLogger(LicenseDatabaseHandler.class);
 
@@ -126,6 +126,7 @@ public class LicenseDatabaseHandler {
         licenseTypeRepository = new LicenseTypeRepository(db);
         customPropertiesRepository = new CustomPropertiesRepository(db);
         obligationListRepository = new LicenseObligationListRepository(db);
+        syncReportRepository = new LicenseDBSyncReportRepository(db);
 
         repositories = new DatabaseRepositoryCloudantClient[]{
                 licenseRepository,
@@ -1164,13 +1165,16 @@ public class LicenseDatabaseHandler {
 
         IMPORT_STATUS = true;
         IMPORT_TIME = currentTime;
+        String startTime = Instant.now().toString();
 
         try {
             String enabled = SW360Utils.getConfigByKey(SW360ConfigKeys.LICENSEDB_ENABLED);
             if (!"true".equals(enabled)) {
                 log.info("LicenseDB integration is disabled. Skipping sync.");
-                return requestSummary.setRequestStatus(RequestStatus.FAILURE)
+                requestSummary.setRequestStatus(RequestStatus.FAILURE)
                         .setMessage("LicenseDB integration is disabled.");
+                createSyncReport("FULL", startTime, requestSummary);
+                return requestSummary;
             }
             String baseUrl = SW360Utils.getConfigByKey(SW360ConfigKeys.LICENSEDB_BASE_URL);
             String username = SW360Utils.getConfigByKey(SW360ConfigKeys.LICENSEDB_USERNAME);
@@ -1178,32 +1182,30 @@ public class LicenseDatabaseHandler {
 
             if (isNullOrEmpty(baseUrl) || isNullOrEmpty(username) || isNullOrEmpty(password)) {
                 log.error("LicenseDB configuration is incomplete (missing baseUrl, username, or password).");
-                return requestSummary.setRequestStatus(RequestStatus.FAILURE)
+                requestSummary.setRequestStatus(RequestStatus.FAILURE)
                         .setMessage("LicenseDB configuration is incomplete.");
+                createSyncReport("FULL", startTime, requestSummary);
+                return requestSummary;
             }
 
             LicenseDBConnector connector = createLicenseDBConnector(baseUrl, username, password);
 
             // Fetch obligations first because license records reference them by UUID
-            List<LicenseDBObligationDTO> obligationDTOs;
-            List<LicenseDBLicenseDTO> licenseDTOs;
-            try {
-                obligationDTOs = connector.fetchAllObligations();
-                licenseDTOs = connector.fetchAllLicenses();
-            } catch (IOException e) {
-                String msg = "LicenseDB sync failed (connection error): " + e.getMessage();
-                log.error(msg, e);
-                return requestSummary.setRequestStatus(RequestStatus.FAILURE).setMessage(msg);
-            }
+            List<LicenseDBObligationDTO> obligationDTOs = connector.fetchAllObligations();
+            List<LicenseDBLicenseDTO> licenseDTOs = connector.fetchAllLicenses();
             log.info("LicenseDB sync: fetched {} active obligations and {} active licenses",
                     obligationDTOs.size(), licenseDTOs.size());
 
-            return processLicenseDBSync(obligationDTOs, licenseDTOs, requestSummary);
+            RequestSummary result = processLicenseDBSync(obligationDTOs, licenseDTOs, requestSummary);
+            createSyncReport("FULL", startTime, result);
+            return result;
 
         } catch (SW360Exception e) {
             String msg = "LicenseDB sync failed: " + e.getMessage();
             log.error(msg, e);
-            return requestSummary.setRequestStatus(RequestStatus.FAILURE).setMessage(msg);
+            requestSummary.setRequestStatus(RequestStatus.FAILURE).setMessage(msg);
+            createSyncReport("FULL", startTime, requestSummary);
+            return requestSummary;
         } finally {
             IMPORT_STATUS = false;
         }
@@ -1360,7 +1362,6 @@ public class LicenseDatabaseHandler {
                 "{\"licensesCreated\":%d,\"licensesUpdated\":%d,\"obligationsCreated\":%d,\"obligationsUpdated\":%d}",
                 licensesCreated, licensesUpdated, obligationsCreated, obligationsUpdated);
         log.info("LicenseDB sync complete: {}", message);
-        updateLastSyncTimestamp();
         return requestSummary.setTotalElements(total).setTotalAffectedElements(affected)
                 .setMessage(message).setRequestStatus(RequestStatus.SUCCESS);
     }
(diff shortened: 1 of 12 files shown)


backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
```diff
@@ -1428,6 +1428,130 @@ private void syncLicenseObligationList(License license, Set<String> oldObligatio
         }
     }
 
+    private Instant readLastSyncTimestamp() {
+        try {
+            Document doc = null;
+            try {
+                doc = db.getDocument(LICENSEDB_SYNC_STATE_DOC_ID);
```
GMishx: I have issues only with this. Currently, the sync state is overwritten after each sync. I cannot track when and how many times the sync ran, what was the status for each run, how many success, how many failure. I cannot keep track of when the sync started, when it ended to understand the performance of the sync.

GMishx: I'd recommend checking the reporting done in `SVMSyncHandler` for better implementation.
https://github.com/eclipse-sw360/sw360/blob/9b5207e5badddb2e47fc89cc7799d06d30cbbb24/backend/vmcomponents/src/main/java/org/eclipse/sw360/vmcomponents/handler/SVMSyncHandler.java#L86

You can see, this processor uses a class called `VMProcessReporting`, containing all the helpful properties, stored multiple times in the DB with a given `type` so it can be analyzed over a period of time.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/tools/LicenseDBConnector.java
```diff
@@ -63,23 +68,84 @@ public List<LicenseDBObligationDTO> fetchAllObligations() throws IOException {
                 .collect(Collectors.toList());
     }
 
+    public boolean pingHealth() {
+        try {
+            doGetAsNode(baseUrl + "/api/v1/licenses?limit=1&page=1");
```
GMishx: There is a dedicated `/health` endpoint for this very specific use-case. Why not using that?

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/tools/LicenseDBConnector.java
```diff
@@ -63,23 +68,84 @@ public List<LicenseDBObligationDTO> fetchAllObligations() throws IOException {
...
+        }
+        return changed;
+    }
+
     private <T> List<T> doGet(String url, TypeReference<List<T>> typeRef) throws IOException {
+        return executeGet(url, response -> {
+            int code = response.getCode();
+            if (code != 200) {
+                log.error("LicenseDB request failed with HTTP {} for: {}", code, url);
+                throw new IOException("LicenseDB request returned HTTP: " + code);
```
GMishx: Please use `SW360Exception`. You can indicate the `code` there, so catcher can easily decide what to do based on error code.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/tools/LicenseDBConnector.java
```diff
@@ -63,23 +68,84 @@ public List<LicenseDBObligationDTO> fetchAllObligations() throws IOException {
...
+            }
+        });
+    }
+
+    private JsonNode doGetAsNode(String url) throws IOException {
+        return executeGet(url, response -> {
+            int code = response.getCode();
+            if (code != 200) {
+                log.error("LicenseDB request failed with HTTP {} for: {}", code, url);
+                throw new IOException("LicenseDB request returned HTTP: " + code);
```
GMishx: Here as well.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/admin/licensedb/Sw360LicenseDBAdminService.java
```diff
@@ -0,0 +1,59 @@
...
+        throwIfNotAdmin(user);
+        return ThriftClients.makeLicenseClient().importIncrementalLicenseDBLicenses(user);
+    }
+
+    public Map<String, Object> getHealth(User user) throws TException {
+        throwIfNotAdmin(user);
+        LicenseService.Iface client = ThriftClients.makeLicenseClient();
+        Map<String, String> syncStatus = client.getLicenseDBSyncStatus(user);
+        boolean connected = client.pingLicenseDBHealth(user);
+        Map<String, Object> health = new LinkedHashMap<>();
```
GMishx: Please create the response as a POJO (in the same namespace if you want), so the properties can be documented easily and parsed by Swagger too. You can make use of Jackson to omit `null` values and reuse the same class in `getSyncStatus()` too by setting `connected` as `null` (or a dummy).

GMishx (changes requested):
Some suggestions.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
Sandipmandal25: resolved each sync now creates a new report document instead of overwriting

GMishx: I am not sure @Sandipmandal25 , you are still reading the status from a fixed ID here `LICENSEDB_SYNC_STATE_DOC_ID`

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/tools/LicenseDBConnector.java
```diff
@@ -49,37 +56,111 @@ public LicenseDBConnector(String baseUrl, LicenseDBTokenManager tokenManager) {
...
+            HttpGet request = new HttpGet(url);
+            request.setHeader("Authorization", "Bearer " + token);
+            request.setHeader("Accept", "application/json");
+            try (CloseableHttpClient httpClient = HttpClients.custom()
+                    .setDefaultRequestConfig(REQUEST_CONFIG).build()) {
+                return httpClient.execute(request, response -> {
+                    int code = response.getCode();
+                    if (code != 200) {
+                        log.error("LicenseDB request failed with HTTP {} for: {}", code, url);
+                        throw new IOException(String.valueOf(code));
```
GMishx: Still thrown as `IOException`, just changed the `throws` of the function to `SW360Exception`. This fails your compilation and does not satisfy the change requested earlier.

GMishx (changes requested):
Still open points from previous review are there. Please recheck.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
Sandipmandal25: I ve changed it so each sync now creates a separate report document with its own id like vmprocessreporting. For reading the status was planning to store a pointer to the latest report in the fixed doc and fetch the actual report from there. Would that work or or would you prefer a view based query to find the latest report directly?

GMishx: How can you be sure about the `LICENSEDB_SYNC_STATE_DOC_ID`? I see some statement which is updating the document ID with this value. This is very bad practice to update the ID of a DB document.

The latest sync must always be fetched from the DB, filtered by the report type and success status and sorted by updated/created date. Not by a fixed ID. This is bad design.

GMishx (changes requested):
Just one pain point.

Sandipmandal25:
> Just one pain point.

will discuss this tomorrow in meet then i will continue.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
```diff
@@ -1428,34 +1436,208 @@ private void syncLicenseObligationList(License license, Set<String> oldObligatio
...
+            IndexField endDateField = new IndexField.Builder().add("endDate", "asc").build();
+            IndexDefinition indexDef = new IndexDefinition.Builder()
+                    .addFields(typeField)
+                    .addFields(statusField)
+                    .addFields(endDateField)
+                    .build();
+            db.createIndex(indexDef, SYNC_REPORT_INDEX_DDOC, "byTypeStatusEndDate", "json");
+        } catch (Exception e) {
+            log.warn("Could not create sync report index: {}", e.getMessage());
         }
```
GMishx: Index creation should be part of Repository, `LicenseRepository` in this example. We have structured format for that.

https://github.com/eclipse-sw360/sw360/blob/f65a7939e5a5b1b00e8c63bd95ec994d70614c19/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseRepository.java#L30

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
```diff
@@ -1428,34 +1436,208 @@ private void syncLicenseObligationList(License license, Set<String> oldObligatio
...
+            PostFindOptions query = db.getQueryBuilder()
+                    .selector(selector)
+                    .useIndex(Collections.singletonList(SYNC_REPORT_INDEX_DDOC))
+                    .addSort(Collections.singletonMap("endDate", "desc"))
+                    .limit(1L)
+                    .build();
+
+            FindResult result = db.getInstance().getClient().postFind(query).execute().getResult();
+            if (result != null && result.getDocs() != null && !result.getDocs().isEmpty()) {
+                return result.getDocs().get(0);
```
GMishx: Same here, should be in Repository too. Not here.

Check existing filters like following
https://github.com/eclipse-sw360/sw360/blob/f65a7939e5a5b1b00e8c63bd95ec994d70614c19/backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseRepository.java#L58-L60

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
```diff
@@ -1428,34 +1436,208 @@ private void syncLicenseObligationList(License license, Set<String> oldObligatio
...
+                } catch (Exception e) {
+                    log.warn("Could not parse endDate from sync report: {}", e.getMessage());
+                }
+            }
+        }
+        return Instant.EPOCH;
+    }
+
+    public Map<String, String> getLicenseDBSyncStatus(User user) {
+        Map<String, String> status = new HashMap<>();
```
GMishx, line 1445: Hope you'd not hate me to ask this. But it will be good to create a class (POJO is good too) to hold this data.

GMishx: You'd need it to create a new Repository anyways.

GMishx (changes requested):

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
GMishx: Now it is resolved. With some more change required.

backend/licenses-core/src/main/java/org/eclipse/sw360/licenses/db/LicenseDatabaseHandler.java
Sandipmandal25: created a pojo and a dedicated repository for it

GMishx: I see you've made the repository and used the new POJO `LicenseDBSyncReport` everywhere. Any reason this function `getLicenseDBSyncStatus` was not updated? Maybe missed to push the changes. Please check again @Sandipmandal25

Sandipmandal25: i need to update this to use a proper pojo and repository for sync reports, but i m a bit confused about the return type. should i also change the thrift contract from `map<string, string>` to a typed struct or should i keep the map at the thrift boundary and use the pojo internally?

GMishx: Oh, I missed there's thrift involved here. Let's not change it then. It will be double rework.

Sandipmandal25: got it thanks. keeping the thrift contract as map and use the pojo internally then convert to map only at the thrift boundary.

GMishx (approved):
Good to go.

PR #4374, opened 2026-07-21
fix(Attachments): Fixed `Checked By` for attachment update

### LINKED ISSUE #4373: Attachments tab showing wrong `checked By` user
### Description

Currently in SW360, it is not possible to change the "checkedBy" status of an existing, already accepted or rejected attachment in a correct way.
For example :
1. When you as userB change an attachment which was *accepted by userA* to *rejected*, the checkedBy value is not updated, so SW360 stores that it is *rejected by userA* while actually it should be *rejected by userB*.
2. When you try to change the status from "accepted" to "Not Checked" in the UI, you get:

```
Error: Cannot deserialize value of type `org.eclipse.sw360.datahandler.thrift.attachments.CheckStatus` from String "NOT_CHECKED": not one of the values accepted for Enum class: [NOTCHECKED, REJECTED, ACCEPTED] at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: java.util.LinkedHashSet[1]->org.eclipse.sw360.datahandler.thrift.attachments.Attachment["checkStatus"])
```

### How to reproduce

Let's take an example attachments of `Project` page :

1. Go to a project detail page, then `Attachments` tab
2. Let's Say the current `Approval Status` is `Accepted` and `Checked By` is `admin1@sw360.org`
3. Now login with different account, let's say : `admin2@sw360.org`
4. Click `Edit Project` for the same project
5. Go to `Attachments` tab
6. Change `Approval Status` to `Rejected`
7. `Update Project`
8. Go to `Attachments` tab of `Project` detail page
9. Look for the `Checked By` column and it will show `Rejected` in red color but the `Checked By` is still showing as `admin1@sw360.org`. Ideally this should be `admin2@sw360.org`

amritkv:
## Description
When updating a project (or component/release), the attachment's **`checkedBy`** (and `checkedTeam`, `checkedOn`) were taken **straight from the frontend request**.

So if **userB** changed an attachment to **Rejected** that **userA** had **Accepted**, SW360 still stored it as *checked by userA* — the wrong person got recorded.

## Root cause
`Sw360AttachmentService.fillCheckedAttachmentData` kept the client-supplied value:

```java
attachment.setCheckedBy(attachment.getCheckedBy() != null ? attachment.getCheckedBy() : user.getEmail());
```

Because the frontend always sends `checkedBy`, the backend never replaced it with the real logged-in user.

## Fix
The backend now **owns** the check-audit fields and ignores what the client sends for them.

- Accepts from the request **only**: `attachmentContentId`, `checkStatus`, `checkedComment`.
- Sets `checkedBy` / `checkedTeam` / `checkedOn` from the **authenticated user** (token) + server date.
- **Transition-aware** (avoids a reverse bug):
  - Check status **changed** → stamp the **current user** as checker.
  - Check status **unchanged** → keep the **original checker** (an unrelated edit must not rewrite it).
  - Status **NOTCHECKED** → clear the checked fields.

This mirrors how `createdBy` already works on creation.

## Changes
| File | Change |
|------|--------|
| `Sw360AttachmentService.java` | Hardened `fillCheckedAttachmentData` (never trusts client); added transition-aware `setCheckedAttachmentDataFromRequest` |
| `ProjectController.java` | Wired into `patchProject` **and** `patchProjectWithNetwork` |
| `ComponentController.java` | Wired in; removed dead call + unused import |
| `ReleaseController.java` | Wired in |
| `Sw360AttachmentServiceTest.java` | 7 new unit tests |

## Behavior (before → after)
| Scenario | Before | After |
|----------|--------|-------|
| userB rejects what userA accepted | stored as userA ❌ | stored as userB ✅ |
| Client spoofs `checkedBy` in payload | trusted ❌ | ignored, real user used ✅ |
| Unrelated edit resends full attachment list | could rewrite checker | original checker preserved ✅ |
| Status set to NOTCHECKED | inconsistent | checked fields cleared ✅ |

### Checklist
- [x] The test cases in this PR are generated using Claude Opus 4.8 model

Linked frontend PR : https://github.com/eclipse-sw360/sw360-frontend/pull/1879

Closes : https://github.com/eclipse-sw360/sw360/issues/4373

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java
@@ -573,14 +573,67 @@ public class Sw360AttachmentService {
             attachment.setCheckStatus(CheckStatus.NOTCHECKED);
         }
         if (!CheckStatus.NOTCHECKED.equals(attachment.getCheckStatus())) {
-            attachment.setCheckedBy(attachment.getCheckedBy() != null ? attachment.getCheckedBy() : user.getEmail());
-            attachment.setCheckedTeam(attachment.getCheckedTeam() != null ? attachment.getCheckedTeam() : user.getDepartment());
-            attachment.setCheckedOn(attachment.getCheckedOn() != null ? attachment.getCheckedOn() : SW360Utils.getCreatedOn());
+            // checkedBy/checkedTeam/checkedOn are backend-owned audit fields.
+            // Never trust the values sent by the client - always derive the checker
+            // identity from the authenticated user and use the server-side date.
+            attachment.setCheckedBy(user.getEmail());
+            attachment.setCheckedTeam(user.getDepartment());
+            attachment.setCheckedOn(SW360Utils.getCreatedOn());
         } else {
             attachment.unsetCheckedBy();
             attachment.unsetCheckedTeam();
             attachment.setCheckedComment("");
             attachment.unsetCheckedOn();
+        }
+    }
+    /**
+     * Authoritatively resolves the attachment check audit fields
+     * ({@code checkedBy}, {@code checkedTeam}, {@code checkedOn}) using the authenticated
+     * user, ignoring whatever the client sent for those fields. Only
+     * {@code attachmentContentId}, {@code checkStatus} and {@code checkedComment} from the
+     * request are honoured to describe the check state.
+     *
+     *   1. check status unchanged vs. the stored attachment -> the original
+     *       checker identity and date are preserved, so an unrelated update (e.g. a
+     *       full attachment-list resubmit) does not rewrite who checked it.
+     *   2. check status changed (accepted/rejected, or switched between them) or
+     *       the attachment is new -> the current user (from the token) and the server
+     *       date are stamped.
+     *
+     * @param incomingAttachments attachments coming from the request body
+     * @param storedAttachments   attachments currently persisted for the entity
+     * @param user                the authenticated user derived from the token
+     */
+    public void setCheckedAttachmentDataFromRequest(Set<Attachment> incomingAttachments,
+            Set<Attachment> storedAttachments, User user) {
+        if (CommonUtils.isNullOrEmptyCollection(incomingAttachments)) {
+            return;
+        }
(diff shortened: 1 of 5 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java
```diff
@@ -573,16 +573,69 @@ public void fillCheckedAttachmentData(Attachment attachment, User user) throws S
...
+     *       the attachment is new -> the current user (from the token) and the server
+     *       date are stamped.
+     *
+     * @param incomingAttachments attachments coming from the request body
+     * @param storedAttachments   attachments currently persisted for the entity
+     * @param user                the authenticated user derived from the token
+     */
+    public void setCheckedAttachmentDataFromRequest(Set<Attachment> incomingAttachments,
+            Set<Attachment> storedAttachments, User user) {
+        if (incomingAttachments == null || incomingAttachments.isEmpty()) {
```
GMishx: Please use common utils


```suggestion
        if (CommonUtils.isNullOrEmptyCollection(incomingAttachments)) {
```

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/attachment/Sw360AttachmentService.java
GMishx, line 570: This function can be removed now since `setCheckedAttachmentDataFromRequest()` is doing the same job.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -2886,6 +2888,13 @@ public ResponseEntity<?> patchProjectWithNetwork(
         updateProject.unsetReleaseIdToUsage();
         sw360Project.unsetReleaseIdToUsage();
 
+        if (updateProject.getAttachments() != null && !updateProject.getAttachments().isEmpty()) {
```
GMishx: Please use common utils


```suggestion
        if (CommonUtils.isNullOrEmptyCollection(incomingAttachments)) {
```

GMishx (changes requested):
Minor cleanup

GMishx (approved):
LGTM

PR #4377, opened 2026-07-22
fix(projects): prevent NPE in releaseIdToUsage

amritkv:
### Issue: 

Fixed null pointer expception while updating `relaseIdToUsage` of a new project.

### Suggest Reviewer
@GMishx


GMishx (approved):
LGTM

PR #4383, opened 2026-07-23
fix(project): allow patch project with MR

amritkv:
#### Description

Allowed the patch call to project with MR


GMishx (approved):
Others will still get MR created.

PR #4394, opened 2026-07-24
fix(component): handle NPE for release and component subscription.

rudra-superrr:
Description: Handling NPE for release and component subscription.


GMishx (approved):
LGTM

PR #4395, opened 2026-07-24
fix(component): restrict the self component in used by components.

rudra-superrr:
Description: restrict the self component in used by components table.


GMishx (approved):
LGTM

PR #4408, opened 2026-07-29
Feat/licensedb clipparser enrichment

Sandipmandal25:
## Summary

Adds CLIParser CouchDB enrichment and makes the LicenseDB sync trigger asynchronous.

When LicenseDB integration is enabled, CLIParser now prefers license text from CouchDB over CLIXML content. The sync endpoint returns `202 Accepted` immediately and runs the sync in the background.

## Changes

* Added `LicenseService.Iface` to the CLIParser constructor.
* Added SPDX ID match, shortname fallback, and CLIXML preservation.
* Added Guava LoadingCache for on-demand license lookup.
* Enrichment runs only when `licensedb.enabled=true`.
* Made `POST /api/licenses/import/LicenseDB` asynchronous.
* Added four tests for matching, fallback, preservation, and disabled state.

## Suggested Reviewers

@GMishx
@deo002
@Farooq-Fateh-Aftab

### DIFF
--- a/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
+++ b/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
@@ -9,6 +9,9 @@
  */
 package org.eclipse.sw360.licenseinfo.parsers;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Sets;
 import org.apache.commons.text.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -16,13 +19,18 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.thrift.TException;
 import org.eclipse.sw360.datahandler.common.CommonUtils;
+import org.eclipse.sw360.datahandler.common.SW360ConfigKeys;
+import org.eclipse.sw360.datahandler.common.SW360Utils;
 import org.eclipse.sw360.datahandler.couchdb.AttachmentConnector;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
 import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
 import org.eclipse.sw360.datahandler.thrift.attachments.AttachmentContent;
+import org.eclipse.sw360.datahandler.thrift.licenseinfo.LicenseInfo;
 import org.eclipse.sw360.datahandler.thrift.licenseinfo.LicenseInfoParsingResult;
 import org.eclipse.sw360.datahandler.thrift.licenseinfo.LicenseNameWithText;
 import org.eclipse.sw360.datahandler.thrift.licenseinfo.ObligationAtProject;
+import org.eclipse.sw360.datahandler.thrift.licenses.License;
+import org.eclipse.sw360.datahandler.thrift.licenses.LicenseService;
 import org.eclipse.sw360.datahandler.thrift.users.User;
 import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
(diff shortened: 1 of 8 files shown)


backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/CLIParser.java
```diff
@@ -61,9 +67,20 @@ public class CLIParser extends AbstractCLIParser {
     private static final String CLI_ROOT_ELEMENT_NAME = "ComponentLicenseInformation";
     private static final String CLI_ROOT_XPATH = "/ComponentLicenseInformation";
     private static final String CLI_ROOT_ELEMENT_NAMESPACE = null;
+    private static final String SPDX_LICENSE_IDENTIFIER_KEY = "SPDX-License-Identifier";
+
+    private LicenseService.Iface licenseClient;
+    private Map<String, License> spdxIdToLicense;
+    private Map<String, License> shortnameToLicense;
```
GMishx: Please consider using [Guava cache](https://github.com/google/guava/wiki/CachesExplained) here as with the current implementation, all the licenses from the DB will be loaded in the maps even if they are used or not and then the object will be thrown. This bootstrapping will accumulate over time and make the overall CLI processing slow.

If we can atleast use the mechanism provided by Guava to load objects in cache on demand, we can eliminate loading the unused licenses (90% of them). I am not sure if we can persist the cache in memory since the `CLIParser` is a use and throw object, but this will still help speedup the process.

Please check and let me know if it makes sense.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/license/Sw360LicenseService.java
```diff
@@ -331,6 +335,24 @@ public RequestSummary importLicenseDBInformation(User sw360User) throws TExcepti
...
     }
 
+    public RequestSummary importLicenseDBInformationAsync(User sw360User) throws TException {
+        LicenseService.Iface sw360LicenseClient = getThriftLicenseClient();
+        if (PermissionUtils.isUserAtLeast(UserGroup.ADMIN, sw360User)) {
+            CompletableFuture.runAsync(() -> {
+                try {
+                    sw360LicenseClient.importAllLicenseDBLicenses(sw360User);
+                } catch (TException e) {
+                    log.error(e);
```
GMishx: Please put a helpful message for the developer looking at the exception to understand from where it came.

GMishx (changes requested):
You'd be extending this to the other inheritors of `AbstractCLIParser` as well?

backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/CLIParser.java
Sandipmandal25: switched to guava loading cache with on demand lookups and ttl expiration.

Sandipmandal25:
> You'd be extending this to the other inheritors of `AbstractCLIParser` as well?

 moved the enrichment to abstractcliparser so combinedcliparser gets it too.

backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
```diff
@@ -74,8 +84,36 @@ public abstract class AbstractCLIParser extends LicenseInfoParser {
     private static final String OBLIGATION_TEXT_ELEMENT_NAME = "Text";
     private static final String OBLIGATION_LICENSE_ELEMENT_NAME = "Licenses";
 
+    private static final int CACHE_TIMEOUT_MINUTES = 15;
+    private static final int CACHE_MAX_ITEMS = 100;
```
GMishx: Let's be little more generous here :-)

```suggestion
    private static final int CACHE_TIMEOUT_MINUTES = 30;
    private static final int CACHE_MAX_ITEMS = 1000;
```

GMishx (changes requested):
One minor tweak. Rest looks good.

Can I merge this branch once you update the values or should I wait for #4372?

backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
Sandipmandal25: updated!

GMishx (approved):
LGTM

PR #4410, opened 2026-07-30
feat(backend): require X-SW360-Internal-Token for backend WAR access

Shivamrut:

## Summary

Closes a security gap where backend Spring Boot WARs on Tomcat were reachable without authentication. End users and anyone who can hit Tomcat could call backends directly (e.g. `GET /components/api/components`) with no login. Only `/resource/...` (resource-server) enforced JWT/Basic; the backends themselves did not.

When enabled, every backend WAR requires a shared secret header. External clients keep using resource-server only; resource-server and backend `*Clients` attach the header automatically. Direct calls without a valid token get `401`.

## Problem

Anyone who could reach Tomcat could call backends directly, for example:

```http
GET http://localhost:8080/components/api/components
```

No login. Resource-server (`/resource/...`) had auth; backends did not.

That meant network exposure of Tomcat (or a misconfigured proxy) was enough to read/mutate backend data without going through the public API security model.

## Fix

When enabled, backends require a secret header:

```http
X-SW360-Internal-Token: <your-secret>
```

Intended traffic path:

```text
You / browser  →  only /resource/...  (JWT / Basic)
                      │
                      ▼
              resource-server adds X-SW360-Internal-Token
                      │
                      ▼
              backend WARs (/components, /users, /health, …)
```

Direct `curl` to `/components/...` (or any backend WAR) **without** the header → **401**.

Business responses are unchanged when the caller is authorized; this only gates access.

## What changed

- **`BackendInternalAuth`** (`libraries/datahandler`) — shared config + constant-time token check
  - Header: `X-SW360-Internal-Token`
  - Properties: `backend.internal.auth.enabled`, `backend.internal.token`
  - Env overrides: `SW360_BACKEND_INTERNAL_AUTH_ENABLED`, `SW360_BACKEND_INTERNAL_TOKEN`
  - Fail-closed if enabled with an empty token
- **`BackendInternalAuthFilter` + auto-config** (`backend/common`) — registered via Spring Boot `AutoConfiguration.imports` for all backends that depend on `backend-common` (including `/health`)
- **`BackendRestClients.shared()`** — attaches the header when auth is enabled and a token is set (resource-server + backend↔backend via `*Clients`)
- Sample `sw360.properties`, Docker/`README_DOCKER.md`, and backend instructions updated
- Unit coverage for the filter

**Default:** `backend.internal.auth.enabled=false` (local/dev convenience). Production should enable it and set a per-deployment secret (not committed to git).

## How to turn it on (local / screenshot verification)

1. Edit `/etc/sw360/sw360.properties` (or set env vars):

```properties
backend.internal.auth.enabled=true
backend.internal.token=test-secret-123
```

Equivalent env:

```bash
export SW360_BACKEND_INTERNAL_AUTH_ENABLED=true
export SW360_BACKEND_INTERNAL_TOKEN=test-secret-123
```

2. Rebuild and redeploy so WARs include the filter (old WARs ignore the props):

```bash
mvn package -P deploy -DskipTests -Dbase.deploy.dir=/opt/apache-tomcat-11.0.4
# then restart Tomcat
```

3. Confirm catalina logs contain:

```text
Backend internal auth is enabled (header X-SW360-Internal-Token).
```

4. Verify with:

```bash
# Reject without token → 401
curl -i http://127.0.0.1:8080/health/api/health

# Reject wrong token → 401
curl -i -H 'X-SW360-Internal-Token: wrong' http://127.0.0.1:8080/health/api/health

# Accept good token → 200
curl -i -H 'X-SW360-Internal-Token: test-secret-123' http://127.0.0.1:8080/health/api/health

# Same pattern on another WAR
curl -i http://127.0.0.1:8080/users/api/users
curl -i -H 'X-SW360-Internal-Token: test-secret-123' http://127.0.0.1:8080/users/api/users
```

Expected: **401** without/wrong token; **200** with the configured secret.

> **Screenshots:** verification screenshots (401 without token / 200 with token) will be attached in a follow-up PR comment.

## How to test (reviewers)

- With auth **disabled** (default): backends behave as before (no header required).
- With auth **enabled** + token set: exercise the curls above; also hit `/resource/...` as a normal user and confirm resource-server → backend calls still succeed (clients attach the header).
- Optional: wrong token must never succeed; empty token with `enabled=true` must fail closed.

### Checklist

Must:
- [x] Related changes documented in PR description
- [ ] Screenshots attached in a PR comment

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/common/security/BackendInternalAuthAutoConfiguration.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/common/security/BackendInternalAuthAutoConfiguration.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Shivamrut<gshivamrut@gmail.com>, 2026. Part of the SW360 Portal Project.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.common.security;
+
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.Ordered;
+
+/**
+ * Registers {@link BackendInternalAuthFilter} for every backend Spring Boot WAR that
+ * depends on backend-common (via Spring Boot auto-configuration).
+ */
+@AutoConfiguration
+public class BackendInternalAuthAutoConfiguration {
+
+    @Bean
+    public FilterRegistrationBean<BackendInternalAuthFilter> backendInternalAuthFilter() {
+        FilterRegistrationBean<BackendInternalAuthFilter> registration = new FilterRegistrationBean<>();
+        registration.setFilter(new BackendInternalAuthFilter());
+        registration.addUrlPatterns("/*");
+        registration.setName("backendInternalAuthFilter");
+        registration.setOrder(Ordered.HIGHEST_PRECEDENCE);
+        return registration;
+    }
+}
--- a/backend/common/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
+++ b/backend/common/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
@@ -1,0 +1,1 @@
+org.eclipse.sw360.common.security.BackendInternalAuthAutoConfiguration
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/rest/BackendInternalAuth.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/rest/BackendInternalAuth.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright Shivamrut<gshivamrut@gmail.com>, 2026. Part of the SW360 Portal Project.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.datahandler.rest;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.util.Properties;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.eclipse.sw360.datahandler.common.CommonUtils;
+
+/**
+ * Shared secret for SW360 backend WAR HTTP access.
+ *
+ * <p>When enabled, every backend request must carry {@link #HEADER_NAME} matching the
+ * configured token. Resource-server and backend {@code *Clients} attach it via
+ * {@link BackendRestClients}. The real token must be set per deployment (properties or
+ * env); do not commit production secrets.
+ */
+public final class BackendInternalAuth {
+
+    private static final Logger log = LogManager.getLogger(BackendInternalAuth.class);
+
+    public static final String HEADER_NAME = "X-SW360-Internal-Token";
+
+    public static final String PROP_ENABLED = "backend.internal.auth.enabled";
+    public static final String PROP_TOKEN = "backend.internal.token";
+
+    public static final String ENV_ENABLED = "SW360_BACKEND_INTERNAL_AUTH_ENABLED";
+    public static final String ENV_TOKEN = "SW360_BACKEND_INTERNAL_TOKEN";
+
+    private static final String PROPERTIES_FILE_PATH = "/sw360.properties";
+
+    private static final boolean ENABLED;
+    private static final String TOKEN;
+
+    static {
+        Properties props = CommonUtils.loadProperties(BackendInternalAuth.class, PROPERTIES_FILE_PATH);
+        ENABLED = parseEnabled(firstNonBlank(System.getenv(ENV_ENABLED), props.getProperty(PROP_ENABLED, "false")));
+        TOKEN = nullToEmpty(firstNonBlank(System.getenv(ENV_TOKEN), props.getProperty(PROP_TOKEN, "")));
+
+        if (ENABLED && TOKEN.isEmpty()) {
+            log.error("{} is true but no token is configured ({} or env {}). "
+                    + "All backend HTTP requests will be rejected until a token is set.",
+                    PROP_ENABLED, PROP_TOKEN, ENV_TOKEN);
+        } else if (ENABLED) {
+            log.info("Backend internal auth is enabled (header {}).", HEADER_NAME);
+        } else {
+            log.info("Backend internal auth is disabled ({}=false).", PROP_ENABLED);
+        }
+    }
+
+    private BackendInternalAuth() {}
+
+    public static boolean isEnabled() {
+        return ENABLED;
+    }
+
+    /**
+     * Configured token for outbound clients. Empty when unset. Never log this value.
+     */
+    public static String getToken() {
+        return TOKEN;
+    }
+
+    /**
+     * Whether {@link BackendRestClients} should send {@link #HEADER_NAME}.
+     */
+    public static boolean shouldAttachHeader() {
+        return ENABLED && !TOKEN.isEmpty();
+    }
+
+    /**
+     * Constant-time compare of the request header to the configured token.
+     */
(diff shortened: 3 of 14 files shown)


Shivamrut:
Backend Auth working SS:
Screenshot from 2026-08-02 22-09-01

libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/rest/BackendInternalAuth.java
```diff
@@ -0,0 +1,113 @@
...
+     * Constant-time compare of the request header to the configured token.
+     */
+    public static boolean matches(String providedHeaderValue) {
+        if (!ENABLED) {
+            return true;
+        }
+        if (TOKEN.isEmpty() || providedHeaderValue == null) {
+            return false;
+        }
+        byte[] expected = TOKEN.getBytes(StandardCharsets.UTF_8);
```
Shivamrut, line 91: String.equals() (and == length checks) can return early on the first mismatch, so timing can leak how much of the guess matched. That’s a classic timing side-channel against shared tokens/API keys.

MessageDigest.isEqual compares in constant time (same work for equal-length inputs), so wrong tokens don’t get a “warmer/colder” signal from response time.

backend/common/src/main/java/org/eclipse/sw360/common/security/BackendInternalAuthAutoConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.common.security;
+
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.Ordered;
+
```
Shivamrut, line 16: If in future a new WAR without backend-common is added,
It would need to:
1. depend on backend-common (preferred - get the filter for free), or
2. manually register BackendInternalAuthFilter / the same auto-config in that WAR.

backend/common/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
```diff
@@ -0,0 +1 @@
+org.eclipse.sw360.common.security.BackendInternalAuthAutoConfiguration
```
Shivamrut, line 1: any WAR that depends on backend-common picks up this class, which registers BackendInternalAuthFilter.

Shivamrut (commented):
Explanation for some changes

Shivamrut:
Hi @GMishx Please review

GMishx:
Hey @Shivamrut , I have some reservation with the proposed approach.

### 1. Shared secret in header is insufficient
The static secret in the header is super fast and super easy to implement. But is also super easy to break as it is sniffable. Local connections are generally not TLSd and thus headers are un-encrypted (only TLS 1.3 supports that). The rotation of the secret is also pretty difficult.

### 2. Proposed approach
SW360 already ships with a `authorization-server`. It already has most of the bells and whistles (RSA key pair management, rotation, etc). Let's utilize it and take it up a notch with [RFC 8693](https://dev.to/kanywst/rfc-8693-deep-dive-token-exchange-310i). I am envisioning something like:
```
                          ┌────────────────────────────┐
                          │      authorization-server    │
                          │  (Spring Authorization       │
                          │   Server + RFC 8693)         │
                          │                              │
                          │  - issues client tokens      │
                          │  - performs token exchange   │
                          │  - exposes /jwks endpoint    │
                          │  - JKS keystore shared via   │
                          │    K8s Secret / volume mount │
                          └──────────────┬─────────────┘
                                         │ ① validate / exchange
                                         │
              ┌──────────────────────────▼────────────────────────┐
              │                   resource-server                     │
              │  accepts: API Token | Basic Auth | OAuth2 | OIDC      │
              │  for all inbound methods → performs token exchange    │
              │  → attaches short-lived internal JWT to backend calls │
              └──────┬───────────────────┬───────────────┬───────┘
                      │ ② internal JWT     │                │
                      ▼                    ▼                 ▼
              ┌───────────────┐  ┌──────────────┐  ┌─────────────┐
              │    licenses    │  │  components   │   │  projects    │
              │    service     │  │   service     │   │   service    │
              │                │  │               │   │              │
              │ validates JWT  │  │ validates JWT │   │ validates JWT│
              │ against /jwks  │  │ against /jwks │   │ against /jwks│
              └───────────────┘  └──────────────┘  └─────────────┘
```

When we move past the current stage and go into something like micro-service and auto scaling, `authorization-server` is equipped enough to use the certificates as shared secrets and with Spring, the backend services can also manage the scaling easily.

### 3. Sample end-to-end workflow
```mermaid
sequenceDiagram
    actor Client
    participant RS as resource-server
    participant AS as authorization-server
    participant LS as licenses-service
    participant LI as licenseinfo-service

    Client->>RS: GET /api/licenses<br/>Authorization: Token "api-key"

    note over RS: Validates API token<br/>Resolves user + roles internally

    RS->>AS: POST /oauth2/token<br/>grant_type=urn:ietf:params:oauth:grant-type:token-exchange<br/>subject_token='signed user assertion'<br/>audience=licenses<br/>scope=service.internal
    AS-->>RS: Internal JWT<br/>(aud=licenses, sub=user@example.com,<br/>scope=service.internal, exp=now+60s)

    note over RS: Caches token until exp-5s

    RS->>LS: GET /api/licenses<br/>Authorization: Bearer 'internal-JWT'

    note over LS: Validates JWT against AS /oauth2/jwks<br/>Checks aud=licenses, scope=service.internal<br/>Extracts sub + sw360_roles

    LS->>LI: GET /api/licenseinfo<br/>Authorization: Bearer 'same internal-JWT'

    note over LI: Same validation —<br/>aud already includes licenseinfo<br/>or audience=sw360-backend (shared)

    LI-->>LS: 200 OK
    LS-->>RS: 200 OK
    RS-->>Client: 200 OK

    note over AS: HA note: all AS instances share<br/>the same JKS keystore (K8s Secret /<br/>volume mount). kid is deterministic<br/>(RFC 7638 thumbprint). JWT validation<br/>is stateless — scales freely.
```

This can be the workflow irrespective of the authentication mechanism used by the client to contact `resource-server`. Only thing will be to provide the appropriate scope and the same token can be repurposed by the backend service to call another backend service.

You can also include the user information, role, group, etc. into the JWT token, so less back-and-forth the the DB.

### 4. High-level implementation changes
`authorization-server` (minimal changes, mostly config)
- Enable RFC 8693 in `SecurityConfig`: register `TokenExchangeAuthenticationConverter` and `TokenExchangeAuthenticationProvider` on the token endpoint.
- Register `resource-server` as an OAuth2 client with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange` and `scope=service.internal`.
- Extend `Sw360TokenCustomizerConfig`: when `context.getAuthorizationGrantType()` is `TOKEN_EXCHANGE`, set `aud` to the requested audience parameter and add `sw360_roles`, `sw360_email`, etc. from the subject token's claims instead of hard-coding `aud: ["sw360-REST-API"]`.
- The `KeyManager` / JKS keystore infrastructure is already HA-ready - same keystore path is shared via volume mount or K8s Secret across all instances. ✅

`resource-server`
- Add an `InternalTokenService`: a Spring `@Component` backed by a `WebClient` that calls `authorization-server`'s token endpoint.
    - For any inbound call: call token exchange with a self-signed assertion carrying the resolved user claims as `subject_token` (using `urn:ietf:params:oauth:token-type:jwt`).
- Cache the result per (`user_sub`, `target_audience`) with a TTL of `token.exp - 5s`.
- Add a `WebClient` or `RestTemplate` `ExchangeFilterFunction` that injects the internal JWT into every outgoing backend call.

`backend services` (e.g. `licenses`)
- Add `spring-boot-starter-oauth2-resource-server` to `pom.xml` (it currently has only `spring-boot-starter-web` - no auth at all).
- Add to `application.yml`:
  ```yaml
  spring:
    security:
      oauth2:
        resourceserver:
          jwt:
            issuer-uri: ${AUTH_SERVER_URL:http://localhost:8090/authorization}
  ```
- Add a `SecurityConfig` that requires all `/api/` to carry a valid JWT with `scope=service.internal` and `aud` containing the service name. This can be part of the library so each service need not implement one of its own.
- Extract `sub` and `sw360_roles` from the JWT for any per-request authorization logic.

> The `authorization-server` dependency is `spring-security-oauth2-authorization-server` which includes the RFC 8693 `TokenExchangeAuthenticationProvider` - no new library dependency is needed. The JKS keystore scaling story is already solved by the existing `KeyManager` design. The bulk of the work is in `resource-server`'s `InternalTokenService` and wiring security into each backend service.

---

Lets discuss this in the GSoC weekly call.

**Note:** The content of heading 4 is mostly AI generated, but validated.

PR #4421, opened 2026-07-31
feat(projects): create new quick filter field

GMishx:

Allow Projects to be searched by `name, description, tag or projectResponsible` fields and a search text. This will be helpful in searches like Link Project pop-up.

This makes it possible to use `/projects?searchText=some+value+in+name+or+tag`

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectSearchHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectSearchHandler.java
@@ -109,6 +109,13 @@ public class ProjectSearchHandler extends BaseNouveauSearchHandler<Project> {
 
     private final NouveauLuceneAwareDatabaseConnector connector;
 
+    private static final List<Project._Fields> QUICK_FILTER_FIELDS = List.of(
+            Project._Fields.NAME,
+            Project._Fields.DESCRIPTION,
+            Project._Fields.TAG,
+            Project._Fields.PROJECT_RESPONSIBLE
+    );
+
     public ProjectSearchHandler(Cloudant client, String dbName) throws IOException {
         super(Project.class, "projects", PROJECT_INDEX_DEFINITION);
         DatabaseConnectorCloudant db = new DatabaseConnectorCloudant(client, dbName);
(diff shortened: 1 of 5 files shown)


backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectSearchHandler.java
```diff
@@ -109,6 +109,13 @@ public class ProjectSearchHandler extends BaseNouveauSearchHandler<Project> {
 
     private final NouveauLuceneAwareDatabaseConnector connector;
 
+    private final List<Project._Fields> QUICK_FILTER_FIELDS = List.of(
```
rudra-superrr: it should be static final,  otherwise a new List.of(...) is allocated every time a ProjectSearchHandler is constructed.

GMishx: Forgot to do that. Pushed the fix.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -303,7 +306,9 @@ public ResponseEntity<CollectionModel<EntityModel<Project>>> getProjectsForUser(
             filterMap.put(Project._Fields.NAME.getFieldName(), values);
         }
 
-        if (luceneSearch && !filterMap.isEmpty()) {
+        if (CommonUtils.isNotNullEmptyOrWhitespace(searchText)) {
```
rudra-superrr, line 313: Controller silently ignores name, tag, etc. when searchText is present.

Problem: No runtime warning or error when a caller passes both name=Spring&searchText=foo. The name value is put into filterMap first, then the searchText branch runs and completely ignores it. The caller has no way to know their name filter was dropped.

Fix: Move filterMap population inside the else branch so it is not even computed when searchText is present.

backend/projects/src/main/java/org/eclipse/sw360/projects/ProjectHandler.java
```diff
@@ -104,6 +104,14 @@ public Map<PaginationData, List<Project>> refineSearchPageable(
         return searchHandler.search(subQueryRestrictions, user, paginationData);
     }
 
+    @Override
+    public Map<PaginationData, List<Project>> searchFilteredProjects(
+            String searchText, User user, PaginationData pageData) throws TException {
+        assertUser(user);
```
rudra-superrr, line 111: Empty/null searchText reaches Nouveau with no guard in handler.
Fix: assertNotEmpty(searchText);

rudra-superrr (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
GMishx: That is by design and documented in the endpoint, if this field is present, other filteres will be ignored. But I can throw a 400 instead.

backend/projects/src/main/java/org/eclipse/sw360/projects/ProjectHandler.java
GMishx: That is handled in the nouveau base search handler.

rudra-superrr:
Working as expected.

PR #4423, opened 2026-07-31
fix(User): fix fullname user search

rudra-superrr:
Description: Searching for a user's full name (e.g. "Gaurav Mishra") returned no results because full name was not indexed and not included in the OR query across user fields.


GMishx (approved):
Tested, working as expected.

PR #4429, opened 2026-08-03
feat(obligation): warn when multiple CLI files are accepted for a release 

### LINKED ISSUE #4428: Add warning message for multiple CLI files for the component in SW360
### Description

Currently, SW360 does not provide any indication when a component has multiple CLI files. This can lead to missing obligations because the tool fails to aggregate obligations from multiple accepted CLI file.

#### How to reproduce

- Create a Component and add a Release (e.g., OpenSSL 3.0.1)

- Upload two CLI files (type: COMPONENT_LICENSE_INFO_XML) as attachments to the release

- Set Check Status = ACCEPTED on both CLI files

- Create a Project and link the release to it

- Navigate to Project → Obligations tab

Shivani2118:

> Summary
Added a warnings field in the license obligations response to notify users when multiple CLI files are accepted for a release.(#4428 )

Closes #4428 

### How To Test?

- Create a Component and add a Release (e.g., OpenSSL 3.0.1)
- Upload two CLI files (type: COMPONENT_LICENSE_INFO_XML) as attachments to the release
- Set Check Status = ACCEPTED on both CLI files
- Create a Project and link the release to it
- Navigate to Project → Obligations tab

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -3572,6 +3572,10 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
             Map<String, Object> responseBody = new HashMap<>();
             responseBody.put("processedLicenses", processedLicenses);
             responseBody.put("obligationStatusMap", obligationStatusMap);
+            List<String> multipleCLIWarnings = projectService.getReleasesWithMultipleCLIWarnings(releases);
+            if (!multipleCLIWarnings.isEmpty()) {
+                responseBody.put("warnings", multipleCLIWarnings);
+            }
             return new ResponseEntity<>(responseBody, HttpStatus.OK);
         } else {
             obligationStatusMap = projectService.setLicenseInfoWithObligations(obligationStatusMap,
@@ -3589,6 +3593,10 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
             }
 
             Map<String, Object> responseBody = createPaginationMetadata(pageable, obligationStatusMap);
+            List<String> multipleCLIWarnings = projectService.getReleasesWithMultipleCLIWarnings(releases);
+            if (!multipleCLIWarnings.isEmpty()) {
+                responseBody.put("warnings", multipleCLIWarnings);
+            }
             HalResource<Map<String, Object>> halObligation = new HalResource<>(responseBody);
             return new ResponseEntity<>(halObligation, HttpStatus.OK);
         }
@@ -3669,6 +3677,10 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
 
 
         Map<String, Object> responseBody = createPaginationMetadata(pageable, oblData);
+        List<String> multipleCLIWarnings = projectService.getReleasesWithMultipleCLIWarnings(releases);
+        if (!multipleCLIWarnings.isEmpty()) {
+            responseBody.put("warnings", multipleCLIWarnings);
+        }
         HalResource<Map<String, Object>> halObligation = new HalResource<>(responseBody);
         return new ResponseEntity<>(halObligation, HttpStatus.OK);
     }
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
@@ -3264,7 +3264,7 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
             if (!releases.isEmpty()) {
                 final Map<String, String> releaseIdToAcceptedCLI = Maps.newHashMap();
                 obligationStatusMap = CommonUtils.nullToEmptyMap(projectService.setLicenseInfoWithObligations(
-                        Maps.newHashMap(), releaseIdToAcceptedCLI, releases, sw360User));
+                        Maps.newHashMap(), releaseIdToAcceptedCLI, releases, sw360User, id, new ArrayList<>()));
             }
         }
 
@@ -3558,7 +3558,7 @@ public class ProjectController implements RepresentationModelProcessor<Repositor
     }
 
     @Operation(
-            description = "Get license obligation data of project tab.",
+            description = "Get license obligation data of project tab. Look out for an optional **warnings** field (list of strings) in the response: it appears when a release has multiple CLI files, indicating the first CLI was selected automatically.",
             tags = {"Projects"}
     )
     @ApiResponses(value = {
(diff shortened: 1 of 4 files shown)


rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
GMishx, line 1: Please document the new response key `warnings` using the OpenAPI annotations. Otherwise the clients will never know about it.

GMishx (changes requested):
Documentation needs to be updated.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
```diff
@@ -808,8 +808,12 @@ public Map<String, ObligationStatusInfo> setLicenseInfoWithObligations(
             }
             final String releaseId = release.getId();
 
-            if (approvedCliAttachments.size() == 1) {
-                final Attachment filteredAttachment = approvedCliAttachments.get(0);
+            if (approvedCliAttachments.size() > 1) {
```
amritkv: Here this `approvedCliAttachments` var might have some non-accepted CLI-XMLs as in the previous logic, if there are no approved CLI-XML is found, in the next if loop it is checking for just CLI-XMLs.

amritkv (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
amritkv: Closing this comment as this an expected behaviour

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
```diff
@@ -817,14 +808,20 @@ public Map<String, ObligationStatusInfo> setLicenseInfoWithObligations(
...
-            if (approvedCliAttachments.size() == 1) {
-                final Attachment filteredAttachment = approvedCliAttachments.get(0);
-                final String attachmentContentId = filteredAttachment.getAttachmentContentId();
-
-                if (releaseIdToAcceptedCLI.containsKey(releaseId)
-                        && releaseIdToAcceptedCLI.get(releaseId).equals(attachmentContentId)) {
-                    releaseIdToAcceptedCLI.remove(releaseId);
+            // If attachment usage selected a specific CLI for this release, use only that one
+            List<Attachment> cliAttachmentsToProcess = approvedCliAttachments;
+            if (releaseIdToAcceptedCLI.containsKey(releaseId)) {
```
GMishx: `releaseIdToAcceptedCLI` still contains map of `ReleaseID => ObligationStatus`. This maps accepted CLIs.

Instead, `List<AttachmentUsage> allUsagesByProject = projectService.getUsedAttachments(usedBy, null);` `AttachmentUsage` should be used in case there are more than 1 accepted CLI.

1. Fetch all accepted CLI from the Release. If there is only one, you are done.
2. If there are no accepted CLI, fetch all CLI, irrespective of the status. If there is only one, you are done.
3. If you reach here, that means there is more than 1 CLI. Then we should consult the `AttachmentUsage`.
    - In the usage, find the current Release, then check if it has a CLI usage marked or not.
    - If there is only 1 CLI, use that and you are done.
4. If you reach here, the worst case, there is no AttachmentUsage or there is still more than 1 CLI in AttachmentUsage. In such case, select the first CLI and put on the warning.

GMishx: This part is done correctly in `getReleasesWithMultipleCLIWarnings` but not utilized here where it is needed.

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/ProjectController.java
```diff
@@ -4929,4 +4855,28 @@ public List<String> getAllProjectGroups() {
...
+    private static class ProjectLicenseObligationsReleaseViewResponse {
+        @Schema(description = "Processed license entries")
+        public List<LicenseInfoParsingResult> processedLicenses;
+
+        @Schema(description = "Obligation status information keyed by obligation title")
+        public Map<String, ObligationStatusInfo> obligationStatusMap;
+
+        @Schema(description = "Warning messages when more than one CLI is selected in Attachment Usage for a release")
+        public List<String> warnings;
+    }
```
GMishx: Please do not create these classes. They will become hard to maintain in the future. Please put a `@Schema(example=)` in the `@Operation` itself, or update the comment to something like **Look out for warnings**

GMishx (changes requested):

rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/project/Sw360ProjectService.java
```diff
@@ -873,6 +894,44 @@ public Map<String, ObligationStatusInfo> setLicenseInfoWithObligations(
         return obligationStatusMap;
     }
 
+    public List<String> getReleasesWithMultipleCLIWarnings(String projectId, List<Release> releases) {
+        Map<String, AttachmentUsage> licenseInfoAttachmentUsage = getLicenseInfoAttachmentUsage(projectId);
+        if (licenseInfoAttachmentUsage.isEmpty()) {
+            return Collections.emptyList();
+        }
```
GMishx: There can be scenarios where you just created a Release, attached two CLI files and maybe even approved both of them. Then use this Release in a newly created Project, which will not have attachment usage. In that case, this will always return empty list without checking the actual attachments (which will be done in `setLicenseInfoWithObligations`).

Attachment Usage should be treated as optional, not required.

GMishx (changes requested):
Hey @Shivani2118 , the PR is now much closer to the implementation we wanted.

There is one performance optimization suggestion. The function `getLicenseInfoAttachmentUsage` is complex and is being called twice at `setLicenseInfoWithObligations` and `getReleasesWithMultipleCLIWarnings` function and the response is also converted to same map in both functions.

What if, we modify the function `setLicenseInfoWithObligations` to accept one more parameter which is a `List<String> warnings` and set the warning in the same function itself? Then the caller just check if the list has some entry, then put that in the `responseBody`. Doing so, we can completely remove `getReleasesWithMultipleCLIWarnings` function.

Let me know what you think about this.

Shivani2118:
> Hey @Shivani2118 , the PR is now much closer to the implementation we wanted.
> 
> There is one performance optimization suggestion. The function `getLicenseInfoAttachmentUsage` is complex and is being called twice at `setLicenseInfoWithObligations` and `getReleasesWithMultipleCLIWarnings` function and the response is also converted to same map in both functions.
> 
> What if, we modify the function `setLicenseInfoWithObligations` to accept one more parameter which is a `List<String> warnings` and set the warning in the same function itself? Then the caller just check if the list has some entry, then put that in the `responseBody`. Doing so, we can completely remove `getReleasesWithMultipleCLIWarnings` function.
> 
> Let me know what you think about this.

Hey @GMishx , this approach makes sense to me. just to confirm, you suggest passing `List<String> warnings` to  `setLicenseInfoWithObligations`, setting it in the response body when it has entries,and remove   `getReleasesWithMultipleCLIWarnings` .

GMishx (approved):
Changes looks good.

GMishx:
Issues found while testing:
1. If a Release has 2 "Accepted" CLI and 1 "Rejected" CLI and the Attachment Usage has the Rejected file selected, the endpoint still throws warning which is confusing. This happens because the `// Step 3:` (which should be Step 2) in `Sw360ProjectService.java` filters `cliCandidates` based on `usedContentIds` which makes the `fromUsage` empty and the `cliCandidates` never gets updated.
2. The Release View, which uses the `/projects/{id}/licenseObligations?view=true` endpoint still shows both approved CLI even if the attachment usage has only 1 selected.
    image

GMishx (approved):
LGTM

PR #4336, opened 2026-07-03
Feat/archival on thrift removal

taanvi2205:
## Summary

This PR introduces a new backend archival module and service-api DTOs for creating archive bundles of SW360 entities. The current implementation supports archival of Releases end-to-end, including entity JSON export, changelog/metadata structure, attachment streaming, manifest generation, and archival record tracking.

It also extends the archival flow for Projects by collecting the Project and its linked Releases into the same bundle. Linked Releases that are still referenced by other live Projects are marked as `keepAlive` in the manifest and tracked with the `KEPT_ALIVE` archival status instead of being treated as removable archived entities.

### Archival Backend Module

- Added a new `backend/archival` module.
- Added Spring Boot application setup for the archival service.
- Added archival REST controller and request handling flow.
- Added archival database persistence layer:
  - `ArchivalDatabaseHandler`
  - `ArchivalRecordDocument`
  - `ArchivalRepository`
- Added archive result handling and service constants.
- Added security/application configuration for the module.

### Archive Bundle Generation

- Added archive bundle builder that streams archive data to an output stream.
- Added manifest generation for archived entities.
- Added checksum generation for archived entity content.
- Added attachment metadata and byte-count tracking.
- Added support for writing entity documents and attachments into the archive bundle.
- Added archive format documentation.

### Release Archival

- Wired `Sw360EntityProvider` for `RELEASE` archival.
- Reads Release data from SW360 using `ComponentDatabaseHandler`.
- Resolves the requesting user through `UserRepository`.
- Serializes Thrift Release entities to clean JSON using `ThriftJson`.
- Streams release attachments from CouchDB using `Sw360AttachmentSource`.
- Tracks archival records as `IN_PROGRESS`, `ARCHIVED`, or `FAILED`.

### Project Archival

- Added a Project-specific archival flow.
- Collects the Project document itself.
- Collects all Releases linked to the Project.
- Detects whether linked Releases are still referenced by other live Projects.
- Marks shared linked Releases as `keepAlive` in the archive manifest.
- Adds `KEPT_ALIVE` archival status for Releases that are bundled but should remain live.
- Deletes the archived Project using SW360’s existing `deleteProject` flow.
- Does not cascade-delete Releases from the live database.

### Tests

- Added tests for archival handler behavior.
- Added tests for archive bundle generation.

## Testing

- Added unit coverage for the new archival module pieces.

PR #4435, opened 2026-08-05
Feat/archival resource server

taanvi2205:
## Summary

Routes archival through the resource server instead of the browser calling the standalone archival service directly, matching the other migrated SW360 services. This removes CORS from the archival module and lets the resource server handle auth and identity forwarding.

## Changes

libraries/datahandler - ArchivalServiceRestClient (+ interface & factory): calls the archival backend, forwards the user via X-User-* headers, streams the TAR.GZ through.

rest/resource-server - ArchivalController at /resource/api/archival/* + Sw360ArchivalService.

backend/archival - removes the now-unneeded CORS config.

## Note

The base branch was squashed before this PR was created. As a result, the Lombok `@Data` conversion of the archival service-api DTOs is already part of the base branch and does not appear as a change here.

## Testing

Compiles successfully. End-to-end local testing could not be completed due to environment limitations. The published Docker image lacks the new controller, and using a locally built `resource.war` causes classloader conflicts due to library version mismatches.


taanvi2205:
Hi @GMishx! Base is `feat/thrift/replace-spring` because `feat/archival-on-thrift-removal` only exists in my fork and can't be selected as a base here, so the commits from #4336 are showing up in this diff as well.

GMishx:
Hey @taanvi2205 , https://github.com/eclipse-sw360/sw360/pull/4336 is merged. Please rebase your branch.

PR #4439, opened 2026-08-06
feat(LicenseInfo): enrich parsed licenses with LicenseDB obligations

Sandipmandal25:
## Summary

Enriches CLI license obligations from SW360/CouchDB when LicenseDB integration is enabled.

## Changes

- Reuses the existing CLIParser CouchDB license enrichment flow.
- Fetches obligations for the matched SW360 license.
- Applies only LicenseDB-managed obligations identified by `licensedb-ob-id`.
- Keeps non LicenseDB/manual obligations untouched.

## Suggested Reviewers

@GMishx  
@deo002  
@Farooq-Fateh-Aftab

### DIFF
--- a/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
+++ b/backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
@@ -31,6 +31,7 @@ import org.eclipse.sw360.datahandler.thrift.licenseinfo.LicenseNameWithText;
 import org.eclipse.sw360.datahandler.thrift.licenseinfo.ObligationAtProject;
 import org.eclipse.sw360.datahandler.thrift.licenses.License;
 import org.eclipse.sw360.datahandler.thrift.licenses.LicenseService;
+import org.eclipse.sw360.datahandler.thrift.licenses.Obligation;
 import org.eclipse.sw360.datahandler.thrift.users.User;
 import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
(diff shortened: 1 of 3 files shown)


Sandipmandal25:
small note:  the Licensedb obligation id key is written directly here. i kept it this way to avoid adding extra dependency from another module but i can move it to a shared constant if needed.

Sandipmandal25:
@GMishx i m not 100% sure if this is exactly the flow you had in mind,but my understanding was to refresh licensedb managed obligations from couchdb during cli enrichment similar to the license text flow. happy to adjust if you suggest a better approach.

backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
```diff
@@ -311,9 +312,40 @@ protected void enrichFromCouchDB(LicenseInfo licenseInfo) {
...
+    private void enrichObligationsFromCouchDB(LicenseNameWithText licenseNameWithText, License couchDbLicense) {
+        try {
+            List<Obligation> obligations = licenseClient.getObligationsByLicenseId(couchDbLicense.getId());
+            if (obligations == null || obligations.isEmpty()) {
+                return;
+            }
+            Set<ObligationAtProject> enrichedObligations = new HashSet<>();
+            for (Obligation obligation : obligations) {
+                if (obligation.getExternalIds() == null
+                        || !obligation.getExternalIds().containsKey("licensedb-ob-id")) {
```
GMishx: Please use constants like `LicenseDBDataMapper.EXTERNAL_ID_LICENSEDB_OB`

licenses-core can be made part of this module if it is not breaking any other module. If there are breakages, move it as a constant of this class.

GMishx (changes requested):

GMishx:
> @GMishx i m not 100% sure if this is exactly the flow you had in mind,but my understanding was to refresh licensedb managed obligations from couchdb during cli enrichment similar to the license text flow. happy to adjust if you suggest a better approach.

Yes, this is exactly what we discussed on Tuesday :+1:

backend/licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/AbstractCLIParser.java
Sandipmandal25: licenses-core isnt a dependency of the licenseinfo module so so defined the constant locally

PR #4452, opened 2026-08-08
fix(projects): handle private sub-projects

GMishx:

If a Project contains a sub-project which is not visible to current user following endpoints throw 500 exceptions:
* `/projects/{id}/tabCounts`
* `/projects/{id}/licenseClearing`
* `/projects/{id}/linkedProjects`

This PR handles the cases and NPE.

### How To Test?
1. Login as User1 and create 2 Projects, PublicProject and a PrivateProject.
    - PublicProject has visibility set to EVERYONE
    - PrivateProject has visibility set to PRIVATE
2. Make link the PrivateProject as a sub-project of PublicProject.
3. Login as User2 which should not have visibility for PrivateProject.
4. Open the PublicProject with the new sw360-frontend Project.
5. The page should load with warning that there are linked Projects not visible to current user. The page uses all the mentioned endpoints.


rudra-superrr (approved):
LGTM

PR #4458, opened 2026-08-10
fix(project): fix releases list in obligation page.

rudra-superrr:
Issue: If a user goes to the Obligations tab of a Project, each and every Obligation shows each and every Linked Release irrespective if the Obligation is coming from the Release or not. There should only be the Release which has that Obligation.

Description: fix releases list in obligation page. Now obligations are shown as per releases.


GMishx (approved):
LGTM

Issue #2065, opened 2023-07-26
Provide Import support for Legacy SBOMS


afsahsyeda:
[//]: # (Copyright Siemens AG, 2021. Part of the SW360 Portal Project)
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

Currently, if a component in the CDX SBOM does not have a purl, it does not get created
![image](https://github.com/eclipse-sw360/sw360/assets/115608771/56af24cd-5c5b-4ea6-b513-f62438152cf9)

This holds particularly true for SBOMs that are not managed through packages or those that are considered legacy (C/C++ projects). When dealing with such SBOMs, it is necessary to address this issue by generating the component and release rather than the package.

This is related to #1999

honeyverma1:
SW360 currently reports it as an invalid package and imports neither a package nor a release. What should the intended behavior be ?? After research, I have come to discover that the fallback behaviour is to generate a component and release (as also mentioned above in the issue).
@GMishx can you please greenlight me to work on it, i would love to work on this issue.

akshitjoshii:
taken care in https://github.com/eclipse-sw360/sw360/pull/3002

PR #4465, opened 2026-08-11
fix(backend): implemented a fallback function for CycloneDX SBoMs wit…

honeyverma1:
This PR aims to solve the issue #2065 

## Summary
Whenever a SBoM is imported in the projects section with no existing purl, the application only created the project and displayed a message regarding it as an invalid package.
Below is the screenshot of the error -
issue2065 before

The fix proposed here is to create a fallback function which does not invalidate the package but creates/reuse it as a component+release and connect it to the project made.
Below is the screenshot of the same page after the fix -
Screenshot 2026-08-11 174246

## Changes
1. I created a fallback function that is called when there is no package url found in the parsed SBoM.
2. The fallback function returns a boolean variable.
3. If the function returns true, the fallback was successful.
4. Otherwise the usual path will be followed as it was before.

### Suggest Reviewer
@GMishx @rudra-superrr @arunazhakesan 

### Test
I used a very basic JSON file as my CycloneDX SBOM. I will also upload it here for reference
[test-no-purl.json](https://github.com/user-attachments/files/30936240/test-no-purl.json)

PS - I did not write any javadoc for the function as I am not yet fully aware of the practises that are followed but would love to learn more day-by-day.

Thank you for reviewing.


honeyverma1:
hey @akshitjoshii saw you assigned for the pr ?? i would appreciate your thoughts or opinions regarding this pr, thank you.

akshitjoshii:
Hi @honeyverma1, this issue is already resolved.
#3002

akshitjoshii:
components without PURL are also imported successfully without the creation of pkg which is the desired behaviour.

GMishx:
Thank you for your contributions @honeyverma1 . Closing this PR as issue is already resolved in https://github.com/eclipse-sw360/sw360/pull/3002 . Thanks @akshitjoshii

PR #4475, opened 2026-08-13
fix(release): single-character version searches

rudra-superrr:
Description: Changed n-gram minimum from 2 to 1 so single-character version searches (for example 2) are indexed and matched, including segmented versions like 2.4.1_conflict.


GMishx (approved):
LGTM

PR #4490, opened 2026-08-17
fix(rest): route empty lucene calls for components

GMishx:

This PR fixes the route for calls `/components?luceneSearch=true`. Here the filterMap is empty does not need nouveau but can be handled by default route as well.

Also fix the wrong tag of Merge Requests.

### Suggest Reviewer
@rudra-superrr 

### How To Test?
1. Make call to `/components?luceneSearch=true` and `/components?luceneSearch=false` and compare the results should be the same.
2. Make call to `/components?luceneSearch=true` with some parameters and the search should work as expected.
3. Load the Swagger docs for the API, there should be only one Moderation Requests controller.

### Checklist
Must:


rudra-superrr (approved):
LGTM

PR #4502, opened 2026-08-20
fix(rest): return 404 when patching a non-existent vendor

### LINKED ISSUE #4501: PATCH /vendors/{id} doesn't return 404 for a non-existent vendor
### Description
`PATCH /vendors/{id}` should return a `404` when the vendor doesn't exist, but instead it fails with an unhandled error.

In `Sw360VendorService.vendorUpdate()`, `existingVendor` is fetched by id and only patched if non-null — but it's then passed to `sw360VendorClient.updateVendor()` unconditionally, even when it's `null`:

```java
Vendor existingVendor = sw360VendorClient.getByID(id);
if (existingVendor != null) {
    // ...patch fields...
}
RequestStatus requestStatus = sw360VendorClient.updateVendor(existingVendor, sw360User);
```

So instead of a clean `404`, a request for a vendor id that doesn't exist ends up failing lower down when `null` hits the Thrift call. The `@ApiResponses` on `updateVendor()` in `VendorController` also don't document `404` as a possible outcome, so the API contract doesn't match reality either.

Other update endpoints in this codebase (e.g. `Sw360LicenseService`) already throw `ResourceNotFoundException` in this exact situation, so this looks like a gap rather than intentional behavior.

#### How to reproduce
- `PATCH /vendors/{id}` with an `id` that doesn't exist, and any valid vendor body
- Instead of a `404`, the request fails lower down when the null vendor reaches the Thrift client

Image

kavya685:

`PATCH /vendors/{id}` doesn't return a 404 when the vendor id doesn't exist. In `Sw360VendorService.vendorUpdate()`, `existingVendor` is fetched by id and only patched if it's not null, but it still gets passed to `updateVendor()` even when it is null. Added a check that throws `ResourceNotFoundException` in that case, and updated the Swagger docs to mention 404 as a possible response.

Fixes #4501
No new dependencies.

Added `Sw360VendorServiceTest.java` - tests that it throws 404 when vendor doesn't exist, that it still only updates non-null fields when it does exist, and that thrift exceptions still get wrapped like before.

To test manually: PATCH /vendors/{id} with a bad id, should get 404 now instead of an error.

### Checklist
Must:

### DIFF
--- a/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/Sw360VendorService.java
+++ b/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/vendor/Sw360VendorService.java
@@ -30,6 +30,7 @@ import org.eclipse.sw360.datahandler.thrift.vendors.VendorSortColumn;
 import org.eclipse.sw360.rest.resourceserver.core.BadRequestClientException;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.data.rest.webmvc.ResourceNotFoundException;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.security.access.AccessDeniedException;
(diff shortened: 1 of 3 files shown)


kavya685:
This is my first contribution here, happy to make any changes if needed. Let me know if anything should be adjusted!

GMishx (approved):
Changes looks good.

PR #4503, opened 2026-08-20
fix(Release): invalid source code url

rudra-superrr:
**Description**
Currently, while creating a Release, the backend validates the Source Code Download URL. If the URL is invalid or unreachable, the Release creation fails.
This validation should be removed from the backend so that an invalid Source Code Download URL does not prevent Release creation.
Frontend-side validation can be introduced separately in a follow-up change.

**Expected Behavior**
Release creation should not fail because the Source Code Download URL is invalid or unreachable.
Remove the backend validation/check for the Source Code Download URL during Release creation.
The Source Code Download URL should still be stored as provided.
Frontend validation will be handled separately at a later stage.


GMishx (approved):
LGTM

PR #4516, opened 2026-08-21
fix(backend): harden attachment filesystem path handling

bibhuti230185:
## Summary

This PR updates attachment file path handling in `backend/common/src/main/java/org/eclipse/sw360/datahandler/db/DatabaseHandlerUtil.java` 

## Changes

- add a resolved-path containment check before writing attachment content
- use the resolved path in the file existence check
- keep write behavior unchanged for valid paths
- align the related log message text

## Files changed

- `backend/common/src/main/java/org/eclipse/sw360/datahandler/db/DatabaseHandlerUtil.java`

## Suggested Reviewer
@GMishx


GMishx (approved):
LGTM

PR #4520, opened 2026-08-23
fix(backend): add self-merge guard to mergeVendors, mergeComponents, mergeReleases

### LINKED ISSUE #4519: Merging a vendor, component, or release into itself silently deletes it
### Description

`mergeVendors`, `mergeComponents`, and `mergeReleases` in `VendorDatabaseHandler.java` and `ComponentDatabaseHandler.java` do not guard against `mergeTargetId` and `mergeSourceId`  when they have same ID. When both IDs refer to the same document, the method still runs its normal merge steps (update target, repoint references, delete source) and unconditionally deletes the source at the end, since target and source are the same document, this silently deletes the vendor/component/release the caller intended to keep, while still returning `RequestStatus.SUCCESS`.

#### How to reproduce

1. Create a vendor (or component, or release) and note its ID, e.g. `V1`.
2. Call the merge endpoint/method with the same ID as both target and source:
mergeVendors(mergeTargetId="V1", mergeSourceId="V1", mergeSelection, user)

3. Observe the response: `RequestStatus.SUCCESS`.
4. Attempt to fetch `V1` afterward, it no longer exists.

The same reproduction steps apply to `mergeComponents` and `mergeReleases`

kavya685:

`mergeVendors` (`VendorDatabaseHandler.java`), `mergeComponents`, and `mergeReleases` (both in `ComponentDatabaseHandler.java`) did not guard against `mergeTargetId` and `mergeSourceId` being the same ID. Since each method's final step unconditionally deletes the source document, passing the same ID for both silently deletes the vendor/component/release the caller intended to keep, while still returning `RequestStatus.SUCCESS`.

This PR adds an early guard as the first statement in each of the three methods, rejecting the request with `RequestStatus.FAILURE` before any database fetch, permission check, or side effect occurs.

No new dependencies were added or updated.

Issue: Fixes #4519

Run the added unit/integration tests:
- `VendorDatabaseHandlerTest#testMergeVendors_selfMerge_returnsFailure`
- `ComponentDatabaseHandlerTest#testMergeComponents_selfMerge_returnsFailure`
- `ComponentDatabaseHandlerTest#testMergeReleases_selfMerge_returnsFailure`

### Checklist
Must:

### DIFF
--- a/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -957,6 +957,11 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
     public RequestStatus mergeComponents(String mergeTargetId, String mergeSourceId, Component mergeSelection,
                                          User sessionUser) throws TException {
+        if (mergeTargetId.equals(mergeSourceId)) {
+            log.error("Cannot merge component [" + mergeSourceId + "] into itself.");
+            return RequestStatus.FAILURE;
+        }
+
         Component mergeTarget = getComponent(mergeTargetId, sessionUser);
         Component mergeSource = getComponent(mergeSourceId, sessionUser);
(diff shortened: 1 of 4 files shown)


GMishx (approved):
LGTM

PR #4522, opened 2026-08-25
fix(project): handled update of fields set by backend

amritkv:
Description :

Clicking on Duplicate button for a Project opens "Create Project" view. But the field "modifiedBy" is also filled with old info which is wrong. This field should be empty and let backend fix it.

Frontend PR merged here : https://github.com/eclipse-sw360/sw360-frontend/pull/1933

### Suggest Reviewer
@GMishx


GMishx (changes requested):
This is a very complex solution for a very simple problem. Please update `org.eclipse.sw360.rest.resourceserver.project.ProjectController#normalizeStateForDuplicatedProject` instead to simply drop these fields (rather than throwing exceptions and breaking the backwards compatibility).

PR #4526, opened 2026-08-25
chore(api): perf for components and releases

GMishx:
Performance improvements for /components and /releases
1. Use Nouveau by default if no filtering is provided and `luceneSearch` is true.
2. Use `parallelStream` to perform operations where possible.


rudra-superrr (commented):
Testing was successful.
LGTM

rudra-superrr (approved):

PR #4553, opened 2026-09-02
feat(projects): sort groups before returning

GMishx:

Sort the list of groups returned by the endpoint `/groups` in the Project service since backend sends a Set which is non-sortable entity in Java. While at it, move the Set to sorted list in Project Service not Controller and make performance optimization.


rudra-superrr (approved):
LGTM

PR #4557, opened 2026-09-02
feat(api): exit early if auth fails

GMishx:

Exit early in the auth chain if the auth fails. Don't send the request down the chain.


rudra-superrr (approved):
LGTM

PR #4564, opened 2026-09-03
fix(clients): prevent sequence() from dropping duplicate/null results

### LINKED ISSUE #4562: FutureUtils.sequence() silently drops duplicate results and throws NPE on null results
### Description

`FutureUtils.sequence()` in the `clients` module silently drops results when two or more futures complete with equal values, because it collects results into a `ConcurrentHashMap<T, Boolean>` keyed by the result itself instead of a list. Since a Map can't hold duplicate keys, if two independent futures happen to return equal values, only one makes it into the returned collection - no error, no warning, just a smaller result set than expected.

It also throws an unrelated `NullPointerException` if any future completes successfully with a `null` result, since `ConcurrentHashMap` doesn't allow null keys.

#### How to reproduce

Call `FutureUtils.sequence()` with a collection of futures where at least two resolve to equal values, e.g.:

```java
List<CompletableFuture<Integer>> futures = Arrays.asList(
        CompletableFuture.completedFuture(1),
        CompletableFuture.completedFuture(1),
        CompletableFuture.completedFuture(2));

Collection<Integer> results = FutureUtils.block(FutureUtils.sequence(futures, ex -> true));
```

Expected: `results` has 3 elements (`[1, 1, 2]`), one per future.
Actual: `results` has 2 elements (`[1, 2]`) - one of the `1`s is silently lost.

Similarly, a future completing with `null` throws NPE instead of being included in the results.

kavya685:

`FutureUtils.sequence()` collects results into a `ConcurrentHashMap<T, Boolean>` keyed by the result value itself, instead of a list. Two problems from that:
- If two different futures happen to complete with equal values, the map only keeps one entry - the other silently disappears from the returned collection. No error, no log, just fewer results than futures you passed in.
- If any future completes with `null`, it throws an NPE, since `ConcurrentHashMap` doesn't allow null keys.

Fixed by swapping the map for a `Collections.synchronizedList(new ArrayList<>())` - still thread-safe for concurrent completions, but now every successful future contributes its own entry regardless of duplicates or nulls.

Fixes https://github.com/eclipse-sw360/sw360/issues/4562

No new dependencies added.

### Suggest Reviewer
@GMishx 

Added two regression tests to `FutureUtilsTest`

### Checklist
Must:

### DIFF
--- a/clients/client/src/main/java/org/eclipse/sw360/clients/utils/FutureUtils.java
+++ b/clients/client/src/main/java/org/eclipse/sw360/clients/utils/FutureUtils.java
@@ -15,12 +15,13 @@ import org.eclipse.sw360.http.utils.HttpConstants;
 import org.eclipse.sw360.http.utils.HttpUtils;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


GMishx (changes requested):
The function `org.eclipse.sw360.clients.utils.FutureUtils#sequence` is called by no one and can be removed instead.

kavya685:
> The function `org.eclipse.sw360.clients.utils.FutureUtils#sequence` is called by no one and can be removed instead.

Done. Also removed those 3 tests related to FutureUtils.sequence()

GMishx (approved):
Good to go.

PR #4565, opened 2026-09-03
fix(fossology): FOSSology Clearing Process Stuck at Step 2

rudra-superrr:
Problem Statement
When the feature flag disable.clearing.fossology.report.download=true is enabled, the FOSSology clearing process gets stuck at "Scanning source in progress" status with a PROCESSING response from the backend. The process should complete after the scan (step 2) is finished with:

- Process status: DONE
- Message: "Scanning source done"
- Progress bar: 100%
- Process steps: 2 (upload/01 and scan/02, no report/03)

Instead, the frontend remains stuck polling indefinitely because the backend continues returning PROCESSING status even after the scan completes.

Root Causes Identified

1. Hardcoded 3-Step Validation - The completion check expected exactly 3 steps (upload, scan, report), but when report generation is disabled, only 2 steps exist, causing the completion logic to always return false.

2. Process Status Never Set to DONE - When report download is disabled, the handler wasn't explicitly setting processStatus = DONE after scan completion, leaving it in IN_WORK status.

3. Lock-Based Blocking - The REST endpoint checked if a lock was held before verifying actual process completion, so even if the process was done, it returned PROCESSING while the lock was being held.

4. No Final Persistence - The orchestration service didn't make a final call to trigger the handler's completion logic when report generation was skipped.

How the Fix Solves the Problem
Data Flow (After Fix):

1. Scan Completes (FOSSologyHandler.java)

- Scan job returns success status (≥ 0)
- Handler checks if report download is disabled
- If disabled → Sets processStatus = DONE and persists to database

2. Orchestration Finalizes (Sw360ReleaseService.java)

- Service detects scan complete
- Makes final fossologyProcess() call if report disabled
- Handler runs again, confirms DONE status, persists it

3. Endpoint Validates (ReleaseController.java)

- Calls isFOSSologyProcessCompleted(fossologyProcess)
- Method checks: status=DONE, step count=2, all steps complete
- Returns status = SUCCESS ✅

### DIFF
--- a/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
+++ b/backend/fossology/src/main/java/org/eclipse/sw360/fossology/FossologyHandler.java
@@ -400,6 +400,11 @@ public class FossologyHandler implements FossologyService.Iface {
 
                         // Persist any IN_WORK or DONE updates from handleReportStepV2
                         updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
+                    } else {
+                        // Report download is disabled: mark process as DONE after scan completes
+                        log.info("Report generation disabled, marking FOSSology process as DONE after scan completion");
+                        fossologyProcess.setProcessStatus(ExternalToolProcessStatus.DONE);
+                        updateFossologyProcessInRelease(fossologyProcess, release, user, componentClient);
                     }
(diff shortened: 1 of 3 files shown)


GMishx (approved):
Changes looks good.

PR #4567, opened 2026-09-03
fix(clients): prevent sequence() from dropping duplicate/null results

### LINKED ISSUE #4562: FutureUtils.sequence() silently drops duplicate results and throws NPE on null results

Nitish08-08:
Fixes #4562.

## What changes did you make?
Replaced the use of `ConcurrentHashMap` in `FutureUtils.sequence()` with `Collections.synchronizedList(new ArrayList<>())`. 

Previously, `ConcurrentHashMap` would drop results if multiple futures completed with equal values (since map keys must be unique) and threw a `NullPointerException` if any future completed with `null`. Using a thread-safe list resolves both problems while correctly collecting results asynchronously.

Additionally, I added two new automated test cases to `FutureUtilsTest.java`:
- `testSequenceWithDuplicateValues`: Verifies that duplicate results are preserved.
- `testSequenceWithNullValue`: Ensures null results are collected correctly without throwing NPE.

## Pre-Code Checklist
- [x] Verified architectural layer (clients)
- [x] Ensured no new test class was strictly required (updated existing `FutureUtilsTest.java`)
- [x] Ensured EPL-2.0 file headers are intact

## PR Checks 
*(Ensure these pass once the CI runs)*
- [x] Code builds and formats cleanly (`mvn spotless:check`)
- [x] Conventional commit format applied
- [x] ECA (Eclipse Contributor Agreement) is signed

### DIFF
--- a/clients/client/src/main/java/org/eclipse/sw360/clients/utils/FutureUtils.java
+++ b/clients/client/src/main/java/org/eclipse/sw360/clients/utils/FutureUtils.java
@@ -15,12 +15,12 @@ import org.eclipse.sw360.http.utils.HttpConstants;
 import org.eclipse.sw360.http.utils.HttpUtils;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


GMishx:
Thank you for your contribution. However, this is being solved by https://github.com/eclipse-sw360/sw360/pull/4564 already.

PR #4580, opened 2026-09-04
fix(cd): add the missing props for docker env

GMishx:

Add the missing configs to docker's `sw360.properties.template` which are read by ENV. Most of them are for thrift and 2 for SVM.


rudra-superrr (approved):
LGTM

PR #4582, opened 2026-09-05
fix(releases): allow security user to fetch batch

GMishx:

SECURITY_USER role is allowed to read releases but not create anything. But the endpoint `/releases/batch-summary` uses `POST` to fetch information, thus making it an edge-case which was not allowed.

This PR updates the `EndpointsFilter` to handle such exceptions. At the same time, improve logging on `/releases` endpoint for failures caused by missing components.


rudra-superrr (approved):
LGTM

PR #4584, opened 2026-09-05
fix(project): minor fixes for linkProjects API

amritkv:
### Description

Minor fixes for `/projetcs/[id]/linkProjects` API which includes:
 - Added API response `409 CONFLICT` for the projects which are already linked or has cyclic dependency
 - Restructured the example param for `array of project IDs`
 - Refactored the check for already linked projects

### Suggest Reviewer
@GMishx


GMishx (approved):
LGTM

PR #4588, opened 2026-09-07
fix(ci): report true on Build and Test for docker

GMishx:
If the PR contains changes only on Dockerfile (like Dependabot), the build should be skipped and the Build and Test should still return true so the PR can be merged.

This should work with PRs like https://github.com/eclipse-sw360/sw360/pull/4568


rudra-superrr (approved):
LGTM

PR #4591, opened 2026-09-09
fix(rest): Exempt security user on allowed apis

amritkv:
Exempted `SECURITY_USER` role to access below APIs:
 - `POST` and `DELETE` calls to token generation and token revoke APIs at Preference page
 - `POST` call to `/releases/batch-summary` API

### Suggest Reviewer
@GMishx


GMishx (approved):
Good to go.

PR #4600, opened 2026-09-10
fix(ModerationRequest): Reject duplicate attachment requests with a clear comment

### LINKED ISSUE #4599: Release MR with duplicate attachment does not fail
### Description
If a Moderation Request was created on a, say Release, which is adding an attachment. At the same time, if another user adds the same attachment with name and SHA (type source). Then, if the Moderation Request is accepted, the backend sends success response but the Release is not updated due to the restriction there can be only one source with same name and checksum.

#### How to reproduce

1. Edit a Release with normal user and add a Source attachment. This results in a MR.
2. Edit the same Release with privileged user and add the same Source attachment. The Release should be updated.
3. The MR shows success message but is never closed.
4. With a Moderator, open the MR created in step 1 and accept it.

### Expected Result

1. The MR should be accepted and if the Release update failed, it should be closed.
2. The status, why the MR failed should be stored appended the message instead.

### Versions

* Docker version OR
* Last commit id on main

#### SW360 logs

Farooq-Fateh-Aftab:

>
When a Moderation Request (MR) for a **Release** adds an attachment (name + SHA1) that is concurrently added directly to the same Release by another user, accepting the MR previously reported a misleading result: `Sw360ModerationRequestService.acceptRequest()`only checked for `RequestStatus.SUCCESS` and, for any other status (including `RequestStatus.DUPLICATE_ATTACHMENT` returned by`ComponentDatabaseHandler.updateRelease`), it simply returned `ModerationState.REJECTED` to the caller **without ever calling `ModerationService.refuseRequest(...)`**. As a result:

- The Release was **not** updated (correctly, since it would violate the no-duplicate-attachment constraint), but
- The moderation request was **left open** (`PENDING`/`INPROGRESS`) in the database, even though the API response looked like a completed/successful action.

> * Which issue is this pull request belonging to and how is it solving it? (* https://github.com/eclipse-sw360/sw360/issues/4599*)

Closes: https://github.com/eclipse-sw360/sw360/issues/4599

### How To Test?
1. Create a Release with an attachment `A` (type `SOURCE`, filename `foo.jar`, SHA1 `X`).
2. Submit a moderation request for the same Release that adds an attachment with the  same filename (`foo.jar`) and SHA1 (`X`) as an addition.
3. As a different user with write access, directly add the same attachment (`foo.jar` / `X`)  to the Release, so it is now already present when the MR is later reviewed.
4. As a moderator, accept the moderation request via  `PATCH /moderationrequest/{id}` with `{"action": "ACCEPT"}`.
5. Verify:
   - The response reports the MR status as `REJECTED` (HTTP 202).
   - The moderation request is now closed (`REJECTED`) when fetched via
     `GET /moderationrequest/{id}` — it no longer appears under open/pending requests.
   - The moderation request's decision comment is
     `"Duplicate attachment already present. Hence, rejected."`
   - The Release itself is unchanged (no duplicate attachment was added).

### Checklist
Must:

### DIFF
--- a/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Constants.java
+++ b/libraries/datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Constants.java
@@ -65,6 +65,8 @@ public class SW360Constants {
     public static final String PACKAGE_IDS = "packageIds";
     public static final String PROJECT_SEARCH_EMPTY_TOKEN = "__EMPTY__";
     public static final String PROJECT_FILTER_KEY_ATTACHMENT_CREATED_BY = "attachmentCreatedBy";
+    public static final String DUPLICATE_ATTACHMENT_REJECTION_COMMENT =
+            "Duplicate attachment already present. Hence, rejected.";
 
     // Proper values of the "type" member to deserialize to CouchDB
     public static final String TYPE_OBLIGATION = "obligation";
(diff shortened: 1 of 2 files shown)


GMishx (approved):
LGTM

PR #4627, opened 2026-09-21
fix(component): Fixed component visibility

amritkv:
Fixed component visibility issue.

### Suggest Reviewer
@GMishx


GMishx (approved):
LGTM

